PR #8, opened 2017-01-04
Watch for circular references during HollowObjectMapper#initializeTypeState, fail fast

nblair:
While working with hollow for the first time, I was trying to generate types for a fairly complex domain model. Running `HollowObjectMapper#initializeTypeState(Class)` on some of my types would generate `StackOverflowError`s that were kind of tricky to diagnose. 

It turns out that there is a circular reference within the data model I'm working with, not immediately obvious to see (which isn't supported per http://hollow.how/data-modeling/#circular-references).

I thought it might be helpful for others in similar circumstances to have hollow fail fast and point directly to the class that couldn't be mapped.

Sample Exception message:

```
java.lang.IllegalStateException: circular reference detected on field private final java.util.List com.netflix.hollow.core.write.objectmapper.DirectCircularReference.children; this type of relationship is not supported
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 **.ipr
 **.iml
 **.iws
+.idea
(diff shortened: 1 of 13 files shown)


nblair:
Looking into the test failure, I'm going to try another approach to this.

dkoszewnik:
This is a good idea, thanks.  Can we avoid the thread local variable?  Let us know if you need help understanding or addressing the test failures.

PR #12, opened 2017-01-21
Shard large types

dkoszewnik:
What was the problem?
---------------------

Currently, during a delta transition each individual record type builds the entire next state of the type in memory while the current state is retained to answer requests.  This means that the memory pool Hollow reserves needs to be large enough to compose an extra copy of the largest type.  This is a miniaturized version of the [double snapshot](http://hollow.how/advanced-topics/#double-snapshots) problem.

What is the solution?
---------------------

With this change, large types can be transparently sharded into smaller individual chunks, each of which can update independently.  The effect of this change is that Hollow only needs to draw on the memory pool enough to update the largest chunk across all types.


ghost (approved):
Paired with Drew on code review. Looks good.

PR #16, opened 2017-02-03
High Level Producer API

ghost:
* [x] introduce a `WriteState` which models a *data state* or *blob* and provides convenience methods for populating it 
    * [x] has a `version` that will correspond to the published blob
    * [x] `add(Object)` method that behaves as `HollowObjectMapper.add(Object)` does
    * [ ] a schema and record oriented input API that requires less boilerplate than using the low level input API (e.g. `HollowWriteStateEngine` , `HollowSchema`, `HollowWriteRecord`, et al)
        * [x] split off schema portion of API; schema definition shouldn't be done every cycle
* [x] provide a `HollowProducer` cycle & stage based API
    * [x] coordinates `Publisher`, `Announcer`, and `VersionMinter` interfaces
    * [x] hook to be notified of cycles
    * [x] hooks to be notified of stages within a cycle
* [x] introduce a `HollowTask` SAM that accepts a `WriteState`; users implement a `HollowTask` with their data population code; `HollowProducer` runs their task during the appropriate cycle stage 
* [x] check-for/add missing copyright headers
* [x] FIX: integrity check stage doesn't work if `HollowProducer.Blob` implementations delete their temp files like they're supposed to
* [ ] make sure the S3 infrastructure reference code still works

### General

* [x] replace `stdout`/`stderr` printing with listener notifications as applicable
    * [x] remove remaining output unless absolutely needed
* [ ] nomenclature: use "Data Model" consistently when referring to a collection of schemas

### HollowProducer

* [x] add integrity check stage
    * [x] round trip this cycle to a read state (S2)
    * [x] load previous cycle to a read state (S1)
    * [x] check delta integrity
        * [x] `S1.apply(forwardDelta).checksum == S2.checksum`
        * [x] `S2.apply(reverseDelta).checksum == S1.checksum`
    * [x] notify Producer listeners
* [x] add validation listener stage
* [x] move `VersionMinter` into Hollow; no need for now to expose a dial to customize it
* [x] move announcement and blob retrievers to `restore(...)` instead of constructor
* [x] have `restore()` take a `long version` instead of passing an announcement retriever
* [x] move several types as inner classes on `HollowProducer` and drop the `Hollow` prefix on their class names

### StateTransition
* [x] remove from public API; change public API methods back to returning a `long version`
* [x] remove no-arg constructor, must always construct at least as a "restored from snapshot" state
* [x] remove `StateTransition` entirely once we're holding onto previous and current read states (for integrity check)

### FilesystemPublisher and friends

* [x] pull an`AbstractHollowPublisher` up into hollow for all reference filesystem staged publishers to extend (FilesystemPublisher, S3publisher)
* [x] move `FilesystemBlob` into Hollow

## Code Review 2

* [x] remove `HollowConsumer` and `*.hollow.consumer` from the public producer API to prevent the nascent consumer API from holding up releasing the producer API
  * [x] change `restore(...)` to `restore(long,HollowReadStateEngine)` or `restore(HollowClient)` or similar
  * [x] add "This code is Alpha and is *definitely* subject to change" to consumer API


ghost:
@dkoszewnik, here's my first take on a high level record-oriented API as well as the beginning of `WriteState` class:

```java
WriteState writeState = new WriteState(writeEngine, blobVersion);

Records stringRecords = writeState.objectRecordsOfType("String");
Records movieRecords = writeState.objectRecordsOfType("Movie");

stringRecords
  .defineString("value")
  .commitSchema();
movieRecords
  .defineLong("id")
  .defineReference("title", "String")
  .defineInt("releaseYear")
  .commitSchema();

int titleOrdinal = stringRecords
  .setString("value", "The Matrix")
  .add();
int movieOrdinal = movieRecords
  .setLong("id", 1)
  .setReference("title", titleOrdinal)
  .setInt("releaseYear", 1999)
  .add();
```

Working in conjunction with the envisioned `Producer` API:

```java
producer.runCycle(new HollowTask() {
  public void doIt(WriteState writeState) {
    Records movieRecords = writeState.objectRecordsOfType("Movie");
    ...

    // define schemas once per cycle
    movieRecords
      .defineLong("id")
      .defineReference("title", "String")
      .defineInt("releaseYear")
      .commitSchema();
    ...

    // populate write state with all the data
    for(/* each movie in source data */) {
      int titleOrdinal = // add a String record for movie title
      movieRecords
        .setLong("id", movieId)
        .setReference("title", titleOrdinal)
        .setInt("releaseYear", releaseYear)
       .add();
    }
  }
});
```

hollow/src/test/java/com/netflix/hollow/api/producer/ObjectRecordsTest.java
```diff
@@ -0,0 +1,112 @@
...
+    
+    @Before
+    public void setup() {
+        writeEngine = new HollowWriteStateEngine();
+        writeState = new WriteState(writeEngine, 13L);
+    }
+
+    @Test
+    public void definingSchemas() {
+        writeState.objectRecordsOfType("String")
```
ghost: The name `fooRecordsOfType` needs work. It's alright for `objectRecordsOfType`, but will end up sounding like actions for the collection types: `setRecordsOfType`, `listRecordsOfType`, and `mapRecordsOfType`.

Would this be better:

* `objectsOfType(typeName)`
* `setsOfType(typeName)`
* `listsOfType(typeName)`
* `mapsOfType(typeName)`

ghost: ^ @dkoszewnik

ghost: I think the `ofType` part is confusing/misleading for the collection types as well. Maybe short plural nouns is enough:

* `writeState.objects(type)`
* `writeState.sets(type, elementType, keys...)`
* `writeState.lists(type, elementType)`
* `writeState.maps(type, keyType, valueType, keys...)`

ghost:
I'm going to move the record oriented API stuff to a separate PR so that we can iterate on that without holding up the rest of the Producer API changes.

ghost (commented):
Per discussion with @dkoszewnik, schema definition should be done once (per state restoration) and not done every cycle. Will split off a separate, fluent schema API.

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ *  Beta API subject to change.
+ * 
+ * @author Tim Taylor {@literal<timt@netflix.com>}
+ */
+public class HollowConsumer {
+
+    // TODO: timt: is this needed, or do we just use a HollowReadStateEngine in place of this? Created for now
+    //   to have symmetry with HollowProducer.WriteState
+    public static interface ReadState {
```
ghost: Producer-side code for restore and integrity check that loads read states is still looking for a good home in the Consumer-side of the API. Jury is still out on whether `ReadState` justifies its own weight as compared to passing a `HollowReadStateEngine` and `long version` where needed.

It does provide a lighter weight way to couple a read state engine with the version ID used to arrive at it (lighter weight than passing a `HollowClient`). However, a read state engine could be mutated under the covers making a `ReadState` tell a lie.

ghost: `HollowConsumer.ReadState` has nice symmetry with `HollowProducer.WriteState`. But `WriteState` was desired as part of the `runCycle` public API; it's unlikely `ReadState` will surface anywhere in the producer public API, so from a producer viewpoint `ReadState` is internals.

 Too soon to know if the upcoming consumer API will have need of a read state engine abstraction or if we'll just have `HollowReadStateEngine` in the consumer public API.

Issue #21, opened 2017-02-27
Transient variables are not ignored


brettwooldridge:
Can support for ignoring transient variables be added?  Maybe a new annotation (``@HollowTransient``) is needed.

We have a Kotlin class that looks something like this:
```kotlin
data class Monitor(
   var instanceId : Int = -1,
   var setId : Int = -1,
   var hostId : Int = -1,
   var monitorName : String = "",
   var monitorType : String = "",
   var json : Map<String, Int> = emptyMap(),
   @Transient
   var isAccessLimited : Boolean = false)
```
Currently, ``isAccessLimited`` is generated into the model:
```
String {
	string value;
}
Integer {
	int value;
}
MapOfStringToInteger Map<String,Integer> @HashKey(value);
Monitor {
	int instanceId;
	int setId;
	int hostId;
	String monitorName;
	String monitorType;
	MapOfStringToInteger properties;
	boolean isAccessLimited;
}
```

dkoszewnik:
Hi Brett,

Thanks very much for the report.  Transient variables should be ignored by the `HollowObjectMapper` with the latest version __v2.2.0__, are you observing this issue with the latest release?

Thanks again,
Drew.

brettwooldridge:
Yes, we're using 2.2.0.  Keep in mind that we are using Kotlin.  I will take a javap dump of the class when I get into the office.

If there is no objection, I intend to submit a pull request for the annotation.  It seems there is still a valid use case.  For example, our data classes are also annotated for JPA persistence and JSON serialization.  Using framework annotations provides the necessary axes of control.

dkoszewnik:
Sounds good, no objections to adding a `@HollowTransient` annotation.  Looking forward to the pull request, thanks again.

dkoszewnik:
Brett, thanks very much for your pull request.  I have released this in [v2.2.1](https://github.com/Netflix/hollow/releases/tag/v2.2.1).

brettwooldridge:
@dkoszewnik Awesomeness. 👍

PR #23, opened 2017-02-28
Add new @HollowTransient annotation in support of issue #21.

brettwooldridge:


hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/HollowTransient.java
```diff
@@ -0,0 +1,30 @@
...
+package com.netflix.hollow.core.write.objectmapper;
+
+import java.lang.annotation.ElementType;
+
+import java.lang.annotation.Target;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE, ElementType.FIELD})
```
dkoszewnik, line 27: Should the target be only `ElementType.FIELD`?

PR #29, opened 2017-03-08
Update listener

kineshsatiya:
- Add methods to listener to have callback before populating write state and on completion of the populating task.
- Modify status when publishing of blobs is complete to expose blob type for more details like size and type.
- Publisher to return size of the blob published in bytes.

@dkoszewnik @toolbear @davesu please review the changes. Let me know if you know of a better way to get size of the blob published. I've made a change in Publisher interface which requires publish to return size of the blob in bytes. 

Feedback from Drew
- Keep the pattern for listener interface, that is, callback on starting a cycle stage & completion of stage. 
- Refactor the Blob interface to a concrete implementation that uses File system to write Snapshot & Deltas.

Updates
- Added OnBlobArtifactPublishComplete as a callback method when a blob is successfully published.
- Changed Blob interface into a concrete implementation using file system path.
- Changed interface Publisher to an abstract class with default implementations to get blob.


hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -182,36 +200,49 @@ public void removeListener(HollowProducerListener listener) {
...
+
+    private void publishBlob(WriteState writeState, Artifacts artifacts, Blob.Type blobType) throws IOException {
+        HollowBlobWriter writer = new HollowBlobWriter(writeEngine);
+        PublishStatus.Builder builder = (new PublishStatus.Builder());
+        try {
+            switch (blobType) {
+                case SNAPSHOT:
+                    artifacts.snapshot = publisher.openSnapshot(writeState.getVersion());
+                    artifacts.snapshot.write(writer);
+                    break;
```
toolbear: publish, too

ghost: Please to be actually publishing

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -400,18 +452,92 @@ private void announce(HollowConsumer.ReadState readState) {
...
+
+        private Blob(String namespace, long fromVersion, long toVersion, String dir, Blob.Type type) {
+            this.namespace = namespace;
+            this.fromVersion = fromVersion;
+            this.toVersion = toVersion;
+            this.type = type;
+            this.dir = dir;
+
+            switch (type) {
+                case SNAPSHOT:
```
ghost, line 484: If you want, you can move methods onto the enum itself and replace switch with polymorphism.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducerListener.java
```diff
@@ -115,6 +131,16 @@
...
 
     /**
+     * Called once a blob has been published successfully or failed to published. Use {@link PublishStatus#getBlob()} to get more details on blob type and size.
+     * This method is called for every {@link com.netflix.hollow.api.producer.HollowProducer.Blob.Type} that was published.
+     *
+     * @param publishStatus Status of publishing. {@link PublishStatus#getStatus()} returns {@code SUCCESS} or {@code FAIL}.
+     * @param elapsed       time taken to publish the blob
+     * @param unit          unit of elapsed.
+     */
+    public void onBlobArtifactPublishComplete(PublishStatus publishStatus, long elapsed, TimeUnit unit);
```
ghost: Let's reserve `*Complete` for notifications that have a companion `*Start`.

Can maybe omit `Blob` from the name.

ghost (approved):
Good to go once artifacts are actually published.

PR #30, opened 2017-03-08
Make jetty-server an optional dependency

ghost:
* upgraded `jetty-server` version; removed it from the transitive dependencies we expose
* meaningful exception message when attempting to use `HollowDiffUIServer` or `HollowHistoryUIServer` without `jetty-server` provided

Calling code will now have to provide `jetty-server` explicitly or have it provided by their container/runtime.

When a caller doesn't provide `jetty-server` and attempts to use `HollowDiffUIServer` they'll see an exception like:

```
Caused by: com.netflix.hollow.ui.jetty.OptionalDependencyException: please add jetty-server (org.eclipse.jetty:jetty-server) to your dependencies
	at com.netflix.hollow.ui.jetty.AbstractOptionalDependencyHelper.newFactory(AbstractOptionalDependencyHelper.java:33)
	at com.netflix.hollow.diff.ui.jetty.OptionalDependencyHelper.uiServerFactory(OptionalDependencyHelper.java:25)
	at com.netflix.hollow.diff.ui.jetty.HollowDiffUIServer.<clinit>(HollowDiffUIServer.java:24)
```

### DIFF
--- a/hollow-diff-ui/build.gradle
+++ b/hollow-diff-ui/build.gradle
@@ -9,7 +9,7 @@ sourceSets {
  
 dependencies {
     compile project(':hollow')
-    compile 'org.eclipse.jetty:jetty-server:9.3.0.M1'
+    compile 'org.eclipse.jetty:jetty-server:9.4.2.v20170220'
     compile 'org.apache.velocity:velocity:1.7'
     compile 'commons-codec:commons-codec:1.8'
     compile 'commons-io:commons-io:2.3'
--- a/hollow-diff-ui/build.gradle
+++ b/hollow-diff-ui/build.gradle
@@ -9,11 +9,12 @@ sourceSets {
  
 dependencies {
     compile project(':hollow')
-    compile 'org.eclipse.jetty:jetty-server:9.4.2.v20170220'
     compile 'org.apache.velocity:velocity:1.7'
     compile 'commons-codec:commons-codec:1.8'
     compile 'commons-io:commons-io:2.3'
     compile 'com.google.code.gson:gson:2.8.0'
+
+    compileOnly 'org.eclipse.jetty:jetty-server:9.4.2.v20170220'
 
     testCompile 'junit:junit:4.11'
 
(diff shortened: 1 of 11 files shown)


ghost:
@dkoszewnik is our BDFL and has final say on whether he likes this change. Looking for feedback from @twicksell or @jhspaybar on how I'm avoiding `NoClassDefFound` so that I can provide a meaningful exception message about adding a `jetty-server` dependency.

hollow-diff-ui/build.gradle
```diff
@@ -9,12 +9,13 @@ sourceSets {
  
 dependencies {
     compile project(':hollow')
-    compile 'org.eclipse.jetty:jetty-server:9.3.0.M1'
     compile 'org.apache.velocity:velocity:1.7'
     compile 'commons-codec:commons-codec:1.8'
     compile 'commons-io:commons-io:2.3'
     compile 'com.google.code.gson:gson:2.8.0'
 
+    compileOnly 'org.eclipse.jetty:jetty-server:9.+'
```
twicksell: Since jetty likes to publish milestones and rc's to maven central, its probably worth noting that 9.+ like this will still get you milestones and rc's. You're not shipping them to your consumers, but it might cause you some headaches as you go.

dkoszewnik (approved):
Looks good.  Let's add the requisite dependency to the hollow-reference-implementation prior to release.

hollow-diff-ui/build.gradle
ghost: Right. Even though I know better, I keep thinking `+` notation works like similar semver notations.

ghost:
I changed to an explicit jetty-server version and updated hollow-reference-implementation to [have a jetty-server dependency](https://github.com/Netflix/hollow-reference-implementation/commit/41bb3ec2d0ce259630d19840db7f96ddd5c9feb6).

ima gonna merge and make a release. Strictly speaking this is a breaking change, but it's tooling and fixed by adding a dependency, and only breaks the build if you relied on the server instead of using your own runtime's jetty, etc.; I'll go with a minor version bump.

Issue #31, opened 2017-03-08
Build tools plugin for generating consumer api


IgorPerikov:
**UPDATE**
**Build plugins are developing as a standalone projects**:
- [gradle plugin](https://github.com/nebula-plugins/nebula-hollow-plugin)
- [maven plugin](https://github.com/IgorPerikov/hollow-maven-plugin)

Hi! When I was generating consumer api, the first thought I had - there must be a gradle/maven plugin for doing that. I think that executing or embedding java code to generate part of source code is not very convenient. So if the team will find this useful too, I would like to participate and create them. 

Plugin will be configurable with:

- apiClassName
- packageName
- packagesToScan

I see plugin's core part something like that:
```
HollowWriteStateEngine writeEngine = new HollowWriteStateEngine();
HollowObjectMapper mapper = new HollowObjectMapper(writeEngine);
Collection<Class<?>> datamodelClasses = retrieveClasses();
for (Class<?> clazz : datamodelClasses) {
    mapper.initializeTypeState(clazz);
}
String directory = buildDirectoryPath(packageName);
new HollowAPIGenerator(apiClassName, packageName, writeEngine).generateFiles(directory);
```

The main design question is how to indicate which classes should be added (how retrieveClasses method will collect them), I see 3 possible ways to do that:

- specify packagesToScan in buildscript and task will add all classes from these packages
- same as previous, but add only classes that is annotated with some new annotation - `@HollowEntity` ?
- annotations only

I don't like the 3rd one at all, 1st one seems the easiest, but may be not very comfortable for end-users, I like the 2nd one, but it needs new annotation, which should be placed in `com.netflix.hollow:hollow` artifact. I guess there might be some restrictions about this option.

What do you think about it?

dkoszewnik:
Hi Igor, great idea, I believe that we and others would find this useful.

I think we can start with your first suggestion -- specify `packagesToScan` in the buildscript, and the task will add all classes from those packages (recursively?).  It seems to be a reasonable assumption that the POJOs can be grouped under their own package(s) -- if it turns out someone wants to ignore a class from these packages, perhaps we can use the new `@HollowTransient` annotation at the class level to indicate that it should _not_ be serialized, rather than requiring an annotation for inclusion.

Thanks for participating!

IgorPerikov:
plugins are created as standalone projects here:
[gradle](https://github.com/nebula-plugins/nebula-hollow-plugin)
[maven](https://github.com/IgorPerikov/hollow-maven-plugin)

PR #36, opened 2017-03-12
gradle plugin for consumer api generation

IgorPerikov:
According to https://github.com/Netflix/hollow/issues/31

I need some help - push commits or give explanation, whatever fits you best

1) I am not sure what is correct way for publishing plugins and stuff under Netflix TM, should it be published to some special place or directly to https://plugins.gradle.org/ ? I haven't published it yet anywhere, so only local publishing is available at the moment if you wanna see it in action

2) It's also a question for me - how to version plugin - separately from hollow-core (if so - then what is it - 1.0.0 or 0.1.0?) or inherit from root

3) And my english is not very good, so README might be kinda awkward and some names over-detailed


ghost:
@IgorPerikov 

(1) I'm not certain, either. Let me get back to you. My first thought is that something as special purpose as a Hollow gradle plugin doesn't belong in plugins.gradle.org. Stay tuned.

(2) Maybe we should go with having the plugin exist as a standalone project? This probably depends on the outcome of (1). If we move to a separate project I recommend starting with a `0.y.z` that we bump to `1.0.0` once stable. Stay tuned on this as well.

(3) No worries. Create a `README.md` that you think will help people use the plugin. We'll assist you with the editing.

ghost:
Hi @IgorPerikov.

I got some advice from our gradle plugin experts. Here's the plan:

1. Build it using the "Nebula plugin" plugin. [http://nebula-plugins.github.io/](nebula-plugins.github.io/)
2. Use a stand-alone project with its own repository
  * initially we can work from a repository you create *or* if you prefer I'll create a `Netflix/hollow-plugin`
  * when we're ready to release (or have an RC) we'll move it under the `nebula-plugins` org as `nebula-plugins/nebula-hollow-plugin`
3. Prior to release we'll test out the plugin by doing local maven publishes
4. After release the Nebula system will automatically publish the plugin artifacts to Bintray and the Gradle Plugin Portal for us

Confirming that we should start with a `0.0.1` or `0.1.0` version. We'll switch to `1.0.0` when we move it under the Nebula umbrella.

Does that help, @IgorPerikov?

ghost:
~This is a relatively minimal nebula plugin project you can use as a starting point:~

~REDACTED URL~

IgorPerikov:
Hi @toolbear! Yes, sure, it clears up all the questions. 

Just one moment
> ~REDACTED URL~

sadly it's not available for me, is it under internal access only?

ghost:
@IgorPerikov yes, my mistake, that plugin isn't available yet. Also, I'm going to redact that repo URL.

One of our gradle plugin wizards looked at this PR and says your plugin looks idiomatic, so perhaps following  http://nebula-plugins.github.io/ will be sufficient for you to incorporate the `nebula-plugin-plugin`? If not, I can find a suitable, public plugin repo for you to use as reference.

IgorPerikov:
@toolbear thanks, yes, I've looked through some nebula plugins and, guess, it would be enough

IgorPerikov:
@toolbear, I've just realised, that it would be better if you'll create some empty `Netflix/hollow-plugin` repo as you [mentioned before](https://github.com/Netflix/hollow/pull/36#issuecomment-286594151). Currently all subprojects receives general stuff from parent script that is useless for plugin (like `nebula.netflixoss` plugin).

IgorPerikov:
Hi, @toolbear! I finished plugin as a standalone project, if separate repo is a problem, I can push it to my own repo. Just wondering - would it be a problem further or not

IgorPerikov:
https://github.com/Netflix/hollow/issues/31#issuecomment-289195125

PR #46, opened 2017-04-04
Remove annoying stdout output.

brettwooldridge:


ghost:
Thanks for the contribution, @brettwooldridge.

Would you be interested in trying something a bit more minimal but also pluggable? I'm thinking along the lines of:

```java
public abstract class Logger {
    // or the default is a no-op logger
    public static volatile Logger INSTANCE = PrintStreamLogger.DEFAULT;

    public abstract void info(String message, Object... args);
    public abstract void warn(String message, Object... args);
    public abstract void error(String message, Object... args);
}
```

```java
public class HollowBlobReader {
    …
    public void readSnapshot(InputStream is, HollowFilterConfig filter) throws IOException {
        …
        Logger.INSTANCE.info("SNAPSHOT COMPLETED IN {}ms",endTime - startTime);
        Logger.INSTANCE.info("TYPES: {}", typeNames);
        …
    }
}
```

Here's a `PrintStreamLogger` adapted from one we use internally that honors the Log4j/SLF4J *placeholder* expressions:

```
public final class PrintStreamLogger extends Logger {
    public static final Logger DEFAULT = new PrintStreamLogger(System.out, System.err);
    private static final Pattern PLACEHOLDER = Pattern.compile("{}", Pattern.LITERAL);

    private final PrintStream out;
    private final PrintStream err;

    public PrintStreamLogger(PrintStream out, PrintStream err) {
        this.out = out;
        this.err = err;
    }

    @Override
    public void info(String message, Object... args) {
        log(out, message, args);
    }

    @Override
    public void warn(String message, Object... args) {
        log(err, message, args);
    }

    @Override
    public void error(String message, Object... args) {
        log(err, message, args);
    }

    private void log(PrintStream stream, String message, Object... args) {
        final int argc=args.length;
        final String formattedMessage;
        final Throwable cause;
        if (argc > 0) {
            Matcher matcher = PLACEHOLDER.matcher(message);
            StringBuffer sb = new StringBuffer();
            for (int i=0; i < argc; i++) {
                if (!matcher.find()) break;
                matcher.appendReplacement(sb, args[i] != null ? args[i].toString() : "null");
            }
            formattedMessage = matcher.appendTail(sb).toString();
            cause = Throwable.class.isInstance(args[argc-1]) ? (Throwable)args[argc-1] : null;
        } else {
            cause = null;
            formattedMessage = message;
        }
        stream.println(formattedMessage);
        if(cause != null) cause.printStackTrace(stream);
    }
}
```

We use that logger for local dev to avoid dependencies on our full runtime stack and swap in a real logger in production environments.

brettwooldridge:
@toolbear Hmmmm.

With a single ``Logger`` instance you loose the ability to alter the logging level on a per-package or per-class basis.

The benefit of using ``java.util.logging`` is that it is built-in to the JDK, so no external dependency, there are bridges to other frameworks, such as the aforementioned slf4j (jul-to-slf4j) and log4j, and if course it supports independent loggers per class and thus the ability to alter levels or filter at a more granular level.

Is there some objection to JUL that I'm missing?

EDIT: I forgot one, because jul can be bridged, it allows the underlying logging framework to be switched out without a code change.

brettwooldridge:
Bump.

brettwooldridge:
@toolbear This doesn't seem to be going anywhere, and I'm anxious to get rid of this output, so can you tell me how you "swap in a real logger in production environments" without a code change?

ghost:
@brettwooldridge I appreciate that you'd like to clean up the stdout noise. Sorry that this isn't progressing as quickly as you hoped it might.

Logging frameworks is a big topic and we could easily get off in the weeds. I'll try and steer clear of that while still providing you with context why accepting JUL might not be an easy and obvious choice even if it seems like it is on the surface.

> Is there some objection to JUL that I'm missing?

The short answer is that we'd prefer something lighter weight. JUL does fit the bill of introducing zero external dependencies so I understand why you might feel like we've moved the goal post on you a bit.

The more involved answer:

An alternative we're considering is to omit logging from Hollow altogether. Instead we'd expose listener interfaces as appropriate. Then logging could be added by registering a listener that logged to whatever logging backend you wanted. The beta `HollowProducer` API works this way; for that API it's a good fit, but we haven't convinced ourselves that we should repeat that pattern throughout hollow.

Yet another alternative we've discussed is to replace `System.out` with a static `PrintStream` that can be overridden with say, a `NoOpPrintStream`. This would be the most minimal mechanism to remove the "annoying stdout output", at least to the letter of the PR and the spirit for some, but not if you were hoping to move that output to your logging backend.

Whatever we choose, if there's a way for callers to configure/influence it, which is what we desire, that becomes part of Hollow's public API. What might seem like dragging our heels is us being conservative about committing to public API changes until and unless we're confident that's what we want the API to be for a long while.

> there are bridges to other frameworks...it allows the underlying logging framework to be switched out without a code change.

Consider these cases:

* hollow depends on JUL, your code uses SL4FJ
* hollow depends on SLF4J, your code uses JUL

In both cases you have to at least make sure your app now bridges to your desired logging backend. Introduce a 3rd popular backend and it's easy to see how in large scale, complex ecosystems what seemed like a small burden can grow to be a large one when picking up a transitive dependency means you've inherited the requirement to configure X backend to play nicely.

> With a single Logger instance you loose the ability to alter the logging level on a per-package or per-class basis

Just as one person's garbage is another person's treasure, one persons's trivial log line is another's person's critical piece of information. IMO, tuning per package or class and having several levels of logging doesn't actually the solve the problem, but does make the logging framework more complex than absolutely necessary. Take a single class, logging two things both at, say, INFO level; add two different users of the library where the other user's "important" log line is this user's trivial one and vice versa.

In my experience INFO and ERROR are the only necessary levels; adding WARN is usually a compromise.

^ that was me getting into the weeds. Hopefully it illustrates why a feature that maybe seems like an obvious requirement is less valuable to some.

That said, a single Logger instance loses the name of the class where line was logged. This is useful information when included in log output and is typically used to determine a line number as well, which is also useful.

Instead of a shared logger, it could be a shared logger factory. But that's more complicated and the logging code is gross (you can't just have a static final in your class if you were hoping to swap out the logger globally for a unit test, for instance).

ghost:
So...all that said, I guess we ask for continued patience while we zero in on a choice. We aren't deciding in a vacuum; your contribution and how you'd like to see it work are factored in. Please feel welcome to remain in the discussion in this PR or in chat.

brettwooldridge:
@toolbear Thanks for the response.  I can say I agree with nearly all of it.

I'll take you up on your offer to continue the discussion.  I'll offer three real-world experiences:

#### #1
I am the author of the [NuProcess](https://github.com/brettwooldridge/NuProcess) library, which runs at the heart of the [Facebook Buck Build](https://buckbuild.com) tool.  When I developed this library, it initially used SLF4J.  However, when Facebook began adopting it they quickly became the largest user of the project, outside of my company.  They didn't want Buck to depend on any third-party logging library, so I agreed to convert it to JUL at their request.  It's fine.

#### #2
I am the author of [HikariCP](https://github.com/brettwooldridge/HikariCP), arguably having the largest market share among connection pools for greenfield development.  It is the default pool for the Play framework, is integrated into Spring Boot, and baked into products by [large corporations](https://github.com/brettwooldridge/HikariCP/blob/dev/documents/Wall-of-Fame.md).  It is difficult to count the number of users (developers), but *low* estimates are in the hundreds of thousands.  HikariCP uses SLF4J, and I can honestly say that I've never had a complaint about the requirement or a request to remove it despite being deployed in every stack imaginable.

#### #3
Hollow is being added to my company's rather large enterprise server product.  Our product currently contains over 180 open source libraries.  As you might imagine, some of them use log4j (1.2), some of them use JUL, some of them use log4j2, some of them use Apache Commons Logging, and some of them use SLF4j.  My company's code itself began life using log4j, and it is still in probably +50% of our code.  We have been moving to SLF4j on log4j2, certainly for new code, and converting old code whenever we touch a file.

You might think that would be a nightmare to manage.  In truth, it turned out to be trivial.  We have the following logging related jars:
```
commons-logging-1.2.jar
log4j-1.2-api-2.8.1.jar
log4j-api-2.8.1.jar
log4j-core-2.8.1.jar
slf4j-api-1.7.21.jar
slf4j-log4j12-1.7.21.jar
```
And we have two configuration files, ``commons-logging.properties`` and ``log4j2.yaml``.  The ``commons-logging.properties`` file simply consists of:
```
# Use log4j
org.apache.commons.logging.Log=org.apache.commons.logging.impl.Log4JLogger
```
and hasn't been touched in years.  The ``log4j2.yaml`` file is the primary log configuration, as you would expect.  All of the "bridging" from hither and yon just works.

#### Granularity
I am of the mind that libraries should basically, by default, shut-up and do their job.  But I also understand the necessity to turn up the logging when troubleshooting.

A perfect example is Jetty.  We use Jetty, and by default it basically logs the IP/port it is bound to at startup and then never says a word.  However, if you've ever enabled DEBUG logging for ``org.eclipse.jetty`` you've certainly experienced the veritable firehose of logging that is unleashed.  It is reams; literally megabytes per second under load.  But thankfully, you don't have to do that.  One can simply enable DEBUG logging for ``org.eclipse.jetty.security``, for example, or even zero in on a class; due to the use of a logging framework.  To be honest, I don't actually know what framework it uses without going to look at the code, but I know I can turn it on and off in our master config.

So, configurable logging is *really helpful*, especially for developers of said library (when troubleshooting), but also for consumers of the library (when reporting bugs or troubleshooting).

#### Does it matter?

Coming back to Hollow.  Your suggestion of removing all logging is both insane and completely sane.  It is *sane* because, *at this point*, the number of ``System.out`` statements can probably be counted on two hands.  It is *insane* because as someone who has needed to dive into the internals a bit I feel that Hollow could benefit from quite a bit of debug-level logging (along with the odd warn or error).  In the vein of shut-up and do your job, I honestly can't imaging what I would want Hollow to log at INFO level.

So, I think *no logging* is a perfectly reasonable *short term* fix for Hollow; meaning, simply remove the ``System.out`` calls.  And once the decision on a logging framework has been made, the code can be judiciously re-instrumented for logging.

Considering the three scenarios above, I don't think you could make a mistake going with SLF4J, JUL, *or* log4j2.  If I had my druthers I'd pick SLF4J and not look back.  But if the amount of logging is small, JUL is a decent compromise; a no-op implementation of ``java.util.logging.LogManager`` could even be provided.

That's my $0.02.

ghost:
You make a persuasive case, @brettwooldridge. Sounds like JUL is the pareto optimal solution. 

Let's take your changes, go with JUL, and iterate from there as needed.

Before merging this as is, can you be persuaded to use a subset of JUL levels? If hollow only logged with `info`, `warn`, and `severe` that would align with how our team does logging elsewhere. It would also keep the default behavior the same as it is today with `info` being output by default by JUL.

brettwooldridge:
@toolbear Done and done.

dkoszewnik:
Great discussion, thanks very much.  Released in [v2.4.3](https://github.com/Netflix/hollow/releases/tag/v2.4.3)

PR #47, opened 2017-04-05
Feature/hollow producer restore

davesu:
Enhancement to enable HollowProducer to restore blob even after it has existing non empty state 

- Restore after it was already published states
- Add ability to indicate whether Restore Failure is Fatal (throw RuntimeException)
- Add support for announcement metadata
-

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
@@ -65,11 +65,10 @@ public class HollowProducer {
     private final Publisher publisher;
     private final Validator validator;
     private final Announcer announcer;
-    private final WriteEngineWrapper writeEngineWrapper;
+    private final ProducerDataState producerDataState;
     private final VersionMinter versionMinter;
     private final ListenerSupport listeners;
     private ReadStateHelper readStates;
-    private boolean isRestoreFailureFatal=false;
 
     public HollowProducer(HollowProducer.Publisher publisher,
             HollowProducer.Announcer announcer) {
@@ -84,7 +83,7 @@ public class HollowProducer {
         this.validator = validator;
         this.announcer = announcer;
 
-        writeEngineWrapper = new WriteEngineWrapper();
+        producerDataState = new ProducerDataState();
         versionMinter = new VersionMinterWithCounter();
         listeners = new ListenerSupport();
         readStates = ReadStateHelper.newDeltaChain();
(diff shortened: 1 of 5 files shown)


hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -561,7 +625,7 @@ public void cleanup() {
     }
 
     public static interface Announcer {
-        public void announce(long stateVersion);
+        public void announce(long stateVersion, Map<String, String> metadata);
```
dkoszewnik: Adding metadata tags to announcement feels like unnecessary cognitive load -- we should avoid this. The concept of announcement should remain simple: http://hollow.how/producer-consumer/#announcing-the-state. Specific implementations are free to introduce additional complexity as necessary, but that complexity shouldn't leak into the core producer API.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -358,6 +380,48 @@ private void announce(HollowConsumer.ReadState readState) {
         }
     }
 
+    private class WriteEngineWrapper {
```
dkoszewnik: I'd like to understand more about why the WriteEngineWrapper is necessary -- if we can do without it and if not, whether there is a better name to express its purpose

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -83,27 +84,34 @@ public HollowProducer(
...
-    
+
     public void initializeDataModel(HollowSchema... schemas) {
         long start = currentTimeMillis();
-        HollowWriteStateCreator.populateStateEngineWithTypeWriteStates(writeEngine, Arrays.asList(schemas));
+        writeEngineWrapper.initializeDataModel(schemas);
         listeners.fireProducerInit(currentTimeMillis() - start);
     }
 
+    public void setRestoreFailureFatal(boolean isFatal) {
```
dkoszewnik: I think that the behavior of the system when restore fails should be defined by the caller to restore, rather than have to configure it here.

dkoszewnik (changes requested):

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
davesu: This was implemented for the use case to notify client about schema changes on delta.   Any suggestions on an alternative way to propagate this information?

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
davesu: I agree a better name for this class would be good.  

The purpose was to encapsulate classes that needs to be changed on a Restore.  Originally I was going to use WriteState but that is already used.   

Let's try to come up with something better.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
davesu: Ok, let me try that.

ghost (approved):
Looks good.

It still feels like the code for management state engines and object mappers may find a better home in `WriteStateImpl`, but I think it's appropriate to do that in a separate PR to keep this change more surgical.

Issue #44, opened 2017-03-28
NPE for fields with types: Collection, Iterable etc


jan-zywczak:
I'm not sure what should be correct behaviour in this case, but NPE is there worst one - you have to debug application to find our what was the reason. Maybe just a IllegalArgumentException with information that only List, Set and Map types are supported. Or maybe you can storage all Collections or Iterables the same way as List is serialized.

dkoszewnik:
Thanks very much.  Fixed by #48, released in [v2.4.3](https://github.com/Netflix/hollow/releases/tag/v2.4.3)

Issue #14, opened 2017-01-31
Data Model with inheritence


ctatineni:
Hello,

 1)  if i extend Movie class with some abstract class which has primary key column name id, how do i retrieve the data at the consumer using primary key? , even if i try to retrieve all movies , i do not see _getID  in MovieHollow object.

2) is there way to convert Hollow objects back to data model objects on consumer?

dkoszewnik:
Hi, good suggestion in (1).  We don't currently support POJOs with a class hierarchy, but I believe we can.  I'll experiment with this.

As for (2), there is not an automatic way to go back to POJOs on the consumer.

ctatineni:
Thanks, one more thing. seems like transient variables are not ignored.

dkoszewnik:
Hi, (1) is now addressed in release [v2.2.0](https://github.com/Netflix/hollow/releases/tag/v2.2.0).  Thanks again for this suggestion.

PR #48, opened 2017-04-07
Signal encountering an interface when initializing schema

aburghelea:
In case one of the fields encountered by the HollowObjectTypeMapper while looking for fields in the hierarchy (since 7a722cab82eb069014a9007ec248cd0c67471c56) is an interface, ```Class#getSuperclass()``` would return null, resulting in a ```NullPointerException```. (I guess it's the same as #44). A more suitable exception would be ```IllegalArgumentException```

As I see it, the ```NullPointerException``` is visible since #14. Now I start to wonder if it would be a good idea to ignore this kind of fields and just skip them.


dkoszewnik:
Thanks very much for solving this.

Issue #49, opened 2017-04-14
Feedback on experimental Producer API


brettwooldridge:
@toolbear I thought I'd offer some thoughts on the new Producer API under development.

It looks pretty good in general except for one bit which is a show-stopper for my company's use.  We persist [snapshots and deltas] in a database, not in files.  So, the fact that ``Blob`` is a concrete class prevents the use of the new API.

If ``Blob`` was and interface, and ``Blob.withNamespace()`` replaced with a call to a user configurable factory, that would be perfect.  Of course, it is perfectly reasonable to default to an implemention that is file-backed.

Maybe the existing ``HollowProducer`` constructor could utilize the default implementation, and a new constructor with one additional parameter (``HollowProducer.BlobFactory``) added?  Something along the lines of:
```java
public HollowProducer(
            Publisher publisher,
            Validator validator,
            Announcer announcer) {
    this(publisher, validator, announcer, new FileBackedBlobFactory());
}

public HollowProducer(
            Publisher publisher,
            Validator validator,
            Announcer announcer,
            BlobFactory blobFactory) {
    ...
}
```

I *think*, having not actually been able to utilitize the new API, that this may be the only change required to support our use-case.

dkoszewnik:
@brettwooldridge Thanks for the feedback.  An early iteration of this API actually did have the filesystem usage abstracted away, but I suggested [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) because I couldn't foresee a case where it wouldn't be reasonable to stage blobs on a filesystem -- regardless of their ultimate destination.  

Even though your blobs are ultimately persisted in a database, is your producer running in an environment that doesn't have access to a filesystem to temporarily stage the blobs?

brettwooldridge:
@dkoszewnik There are a couple of issues I'll see if I can articulate.  One is I/O.  Given that our server is often co-resident with the database, a file-based solution effectively doubles I/O; write the snapshot/delta to a file, read the snapshot/delta from the file, insert the blob into the database (writing to the transaction log), commit the transaction (read the transaction log and write to the table).

In our case, we can have several thousand namespaces, one per-tenant in a multi-tenancy environment, and therefore several thousand delta-streams, each with snapshots sized in megabytes.  Update frequency is hard to gauge, and will vary quite a bit by tenant.

[...]

PR #50, opened 2017-04-14
Refactor to allow for non-filesystem based persistence.

brettwooldridge:
I took a stab at issue #49


dkoszewnik:
This use case is now addressed in v2.6.0 with the `BlobStager`.

PR #51, opened 2017-04-17
Fixes #33 Allow the user to specify an ExecutorService to be used by the

### LINKED ISSUE #33: HollowAnnouncementWatcher should implement an interface
Currently, the ``HollowClient`` constructor accepts an instance of ``HollowAnnouncementWatcher``.  Because ``HollowAnnouncementWatcher`` is an abstract class that unconditionally constructs an internal ``Executors.newFixedThreadPool`` instance, this does not scale to deployments where there are hundreds or thousands of ``HollowClient`` instances in a single VM.

While it would be ideal if ``HollowAnnouncementWatcher`` were merely an interface, like ``HollowBlobRetriever`` and other types passed to the ``HollowClient`` constructor, that would be a breaking change for existing code.

Maybe introduce a new interface type, have ``HollowAnnouncementWatcher`` implement that type, and have the ``HollowClient`` accept that type.  That would preserve compatibility for existing clients, while allowing alternative implementations.

EDIT: For context, the use case is multi-tenancy, where each tenant (possibly thousands) has their own data model instances with independent namespaces, snapshots, and delta chains.

brettwooldridge:
HollowAnnouncementWatcher so that many HollowClient instances can
potentially share the same ExecutorService instance.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/client/HollowAnnouncementWatcher.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/client/HollowAnnouncementWatcher.java
@@ -36,6 +36,30 @@ import java.util.concurrent.ThreadFactory;
  */
 public abstract class HollowAnnouncementWatcher {
 
+    private final ExecutorService refreshExecutor;
+
+    /**
+     * Construct a HollowAnnouncementWatcher with a default ExecutorService.
+     */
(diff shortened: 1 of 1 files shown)


dkoszewnik:
Released in [v2.4.4](https://github.com/Netflix/hollow/releases/tag/v2.4.4)

PR #64, opened 2017-06-09
updated CLASS_NAME_SUBSTITUTIONS to cover entire java.lang package

ramakocherlakota:


ramakocherlakota:
The point of this PR is that classes other than String, Integer and Boolean can end up causing trouble for the generated API files.  In the situation we were in, we had a Map<String, Object> field on one of our model objects, which cause Hollow to generate a class called Object, with obvious downstream problems.

PR #66, opened 2017-07-11
HollowPrefixIndex for string search (full and partial)

kineshsatiya:
I've added the initial implementation. Need to improvise more on compactness and thread-safety.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
@@ -59,6 +59,11 @@ public class HollowPrefixIndex implements HollowTypeStateListener {
 
     // initialize field positions and field paths.
     private void initialize() {
+
+        //check arguments
+        if (readStateEngine == null || type == null || fieldPath == null)
+            throw new IllegalArgumentException("Null arguments received");
+
         // split the field path with "." char
         fields = fieldPath.split("\\.");
 
@@ -74,7 +79,7 @@ public class HollowPrefixIndex implements HollowTypeStateListener {
             typeSeen.add(tempType);
             HollowSchema schema = readStateEngine.getSchema(tempType);
 
-            if (schema == null) throw new NullPointerException("Null schema found for type " + tempType);
+            if (schema == null) throw new IllegalArgumentException("Null schema found for type " + tempType);
             if (!schema.getSchemaType().equals(HollowSchema.SchemaType.OBJECT))
                 throw new IllegalArgumentException("Field path should be defined in type Objects only, " +
                         "found field " + fields[i] + " in path defined in schema " + schema.getSchemaType().toString());
--- a/hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
@@ -369,8 +369,12 @@ public class HollowPrefixIndex implements HollowTypeStateListener {
         private long bitsPerOrdinalSet;
 
         private long maxNodes;
-        private FixedLengthElementArray nodes, ordinalSet;
+        private FixedLengthElementArray nodes;
         private long indexTracker;// where to create new node
+
+        private FixedLengthElementArray[] ordinalSets;
+        private long nodeBucketSize = 1 << 10;
+        private int totalBuckets;
 
         /**
          * Create new prefix index. Represents a ternary search tree.
@@ -395,8 +399,13 @@ public class HollowPrefixIndex implements HollowTypeStateListener {
             bitsPerNode = bitsPerKey + (3 * bitsForChildPointer);
 
             nodes = new FixedLengthElementArray(memoryRecycler, bitsPerNode * maxNodes);
-            ordinalSet = new FixedLengthElementArray(memoryRecycler, bitsPerOrdinalSet * maxNodes);
             indexTracker = 0;
+
+            totalBuckets = (int) Math.ceil((double) maxNodes / (double) nodeBucketSize);
+            ordinalSets = new FixedLengthElementArray[totalBuckets];
+            for (int i = 0; i < totalBuckets; i++) {
+                ordinalSets[i] = new FixedLengthElementArray(memoryRecycler, nodeBucketSize * bitsPerOrdinalSet);
+            }
 
             // initialize offsets
             leftChildOffset = bitsPerKey;// after first 16 bits in node is first left child offset.
@@ -407,7 +416,9 @@ public class HollowPrefixIndex implements HollowTypeStateListener {
         // tell memory recycler to use these long array on next long array request from memory ONLY AFTER swap is called on memory recycler
         private void recycleMemory(ArraySegmentRecycler memoryRecycler) {
             nodes.destroy(memoryRecycler);
-            ordinalSet.destroy(memoryRecycler);
+            for (int i = 0; i < totalBuckets; i++) {
+                ordinalSets[i].destroy(memoryRecycler);
+            }
         }
 
         private long getChildOffset(NodeType nodeType) {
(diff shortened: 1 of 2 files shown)


kineshsatiya:
Checklist
- [x] Use FixedLengthElementArray to save list of ordinals matching the prefix
- [x] Use FixedLengthElementArray for representing nodes
- [x] Use field paths to traverse to a referenced string type
- [x] More unit tests
- [x] delta updates
- [x] Copy on write for thread safety

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
```diff
@@ -0,0 +1,438 @@
...
+     * @param type            type in the read state engine. Ordinals for this type will be returned when queried for a prefix.
+     * @param fieldPath       fieldPath should ultimately lead to a string fieldPath, which will be the keys to build the prefix index.
+     */
+    public HollowPrefixIndex(HollowReadStateEngine readStateEngine, String type, String fieldPath) {
+        this.readStateEngine = readStateEngine;
+        this.type = type;
+        this.fieldPath = fieldPath;
+
+        // create memory recycle for using shared memory pools.
+        memoryRecycle = new RecyclingRecycler();
```
dkoszewnik: TL;DR A lot to discuss here -- In this initial iteration, you may want to just use `WastefulRecycler.DEFAULT_INSTANCE`.

If we're using a `RecyclingRecycler`, we probably want to use the one from the `HollowReadStateEngine` -- that way we would be sharing the same memory pool as everything else.  

I just realized the `HollowPrimaryKeyIndex` is using a `WastefulRecycler`.  The reason for this is that if many indexes are created in parallel, using a _shared_ `RecyclingRecycler` will cause thread safety issues.  

With the pattern introduced in the `HollowPrimaryKeyIndex` you _can_ draw on the main memory pool by explicitly passing the recycler from the state engine to the constructor.  But currently you just have to _know_ that that implies you aren't creating multiple indexes in parallel.  

Side note: It looks like the generated APIs `*PrimaryKeyIndex` are not doing this -- but perhaps they should -- the problem with doing that right now is that we can't know that users won't instantiate multiple in parallel.  I think right intermediate-term move is probably to make the `RecyclingRecycler` thread-safe, then use it more liberally.

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
```diff
@@ -0,0 +1,438 @@
...
+
+        // split the field path with "." char
+        fields = fieldPath.split("\\.");
+
+        // arrays to save data for field path
+        fieldPositions = new int[fields.length];
+        fieldTypes = new HollowObjectSchema.FieldType[fields.length];
+
+        // traverse through the field path to save field position and types.
+        Set<String> typeSeen = new HashSet<>();
```
dkoszewnik: We shouldn't have to check for a circular path in the schemas, since it's not possible to get a data model that has one.

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
```diff
@@ -0,0 +1,438 @@
...
+        String tempType = type;
+        HollowObjectTypeReadState readState = (HollowObjectTypeReadState) readStateEngine.getTypeDataAccess(type);
+        for (int i = 0; i < fields.length; i++) {
```
dkoszewnik: Is it possible to reuse the logic from the `PrimaryKey` class?

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
```diff
@@ -0,0 +1,438 @@
...
+ */
+public class HollowPrefixIndex implements HollowTypeStateListener {
+
+    private HollowReadStateEngine readStateEngine;
+    private String type;
+    private String fieldPath;
+
+    private Tst prefixIndex;
+    private volatile Tst prefixIndexVolatile;
+    private RecyclingRecycler memoryRecycle;
```
dkoszewnik: This should reference an `ArraySegmentRecycler` -- it could be either a `RecyclingRecycler` (which is a memory pool), or a `WastefulRecycler` (which simply creates new array segments and drops them on the floor when they are returned).

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
```diff
@@ -0,0 +1,438 @@
...
+     * NOTE: Each delta updates creates a new prefix index and swaps the new with current.
+     */
+    public void listenForDeltaUpdates() {
+        readStateEngine.getTypeState(type).addListener(this);
+    }
+
+    /**
+     * Stop delta updates for this index.
+     */
+    public void stopDeltaUpdates() {
```
dkoszewnik: For consistency with the `HollowPrimaryKeyIndex`, can we use the method name `detachFromDeltaUpdates()`?

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
```diff
@@ -0,0 +1,438 @@
...
+        }
+
+        prefixIndex = tst;
+        prefixIndexVolatile = tst;
+        // safe to return previous long arrays on next request for long array.
+        memoryRecycle.swap();
+    }
+
+    /**
+     * Return the key to index in prefix index. Override this method to support tokens for the key. By default keys are indexed as lower case characters.
```
dkoszewnik: Seems simple enough to access this functionality, and it would be really neat to see it in action and verify that it works.  Does it make sense to add a test for the tokenization, and perhaps a simple tokenization scheme (e.g. split words by whitespace)?

dkoszewnik (changes requested):
Looks great, this is a very exciting new feature.  Please see inline comments.

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
```diff
@@ -0,0 +1,438 @@
...
+        memoryRecycle = new RecyclingRecycler();
+        initialize();
+    }
+
+    // initialize field positions and field paths.
+    private void initialize() {
+
+        //check arguments
+        if (readStateEngine == null || type == null || fieldPath == null)
+            throw new IllegalArgumentException("Null arguments received");
```
davesu: Might be a good idea to be explicit about which argument was null.

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
```diff
@@ -0,0 +1,438 @@
...
+        fields = fieldPath.split("\\.");
+
+        // arrays to save data for field path
+        fieldPositions = new int[fields.length];
+        fieldTypes = new HollowObjectSchema.FieldType[fields.length];
+
+        // traverse through the field path to save field position and types.
+        Set<String> typeSeen = new HashSet<>();
+        String tempType = type;
+        HollowObjectTypeReadState readState = (HollowObjectTypeReadState) readStateEngine.getTypeDataAccess(type);
```
davesu: Error handling: might be good idea to check readState!=null.  In other words, make sure specified type is valid within the Hollow Schema.

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
```diff
@@ -0,0 +1,438 @@
...
+
+    // initialize field positions and field paths.
+    private void initialize() {
+
+        //check arguments
+        if (readStateEngine == null || type == null || fieldPath == null)
+            throw new IllegalArgumentException("Null arguments received");
+
+        // split the field path with "." char
+        fields = fieldPath.split("\\.");
```
davesu: could fields.lenght == 0?

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
```diff
@@ -0,0 +1,434 @@
...
+        return new String[]{key};
+    }
+
+    /**
+     * Query the index to find all the ordinals that match the given prefix.
+     *
+     * @param prefix
+     * @return ordinals of the parent object.
+     */
+    public Set<Integer> query(String prefix) {
```
dkoszewnik: For consistency, can this return a `HollowOrdinalIterator` instead?

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
```diff
@@ -0,0 +1,434 @@
...
+
+        /**
+         * Find all the ordinals that match the given prefix.
+         *
+         * @param prefix
+         * @return
+         */
+        private Set<Integer> query(String prefix) {
+            if (prefix == null) throw new IllegalArgumentException("Cannot query null prefix");
+            Set<Integer> ordinals = new HashSet<>();
```
dkoszewnik: Consider using an `IntList` instead

PR #69, opened 2017-07-21
HollowIncrementalProducer - delta-based producer input

dkoszewnik:
A `HollowIncrementalProducer` interface, which formalizes a high-level API for [Delta-Based Producer Input](http://hollow.how/advanced-topics/#delta-based-producer-input) as described in the documentation.

Each individual mutation is specified by using `.addOrModify(obj)` or one of the two flavors of `.delete()`.  

Only "top-level" `OBJECT` types which have primary keys defined in their schema can be targets for mutation.  Currently, each of these top-level types must not be referenced by any other type in the data model.  

Records of other (non-top-level) types are added when they are referenced by newly added or modified records in top-level types.  Similarly, records of other types will be removed if modifications to top-level types cause the last references to those records to be removed.

All mutations are staged until they are flushed to a new data state with a call to `.runCycle()`.

The `HollowIncrementalProducer` wraps a standard `HollowProducer`.  A `HollowIncrementalProducer` can only be used to _continue_ a delta chain -- delta chains must be begun using the standard `HollowProducer` API.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: java
 jdk:
 - oraclejdk7
+dist: precise
 sudo: false
 install: true
 script: "./buildViaTravis.sh"
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Binaries are available from Maven Central and jcenter.
 
 |GroupID/Org|ArtifactID/Name|Latest Stable Version|
 |-----------|---------------|---------------------|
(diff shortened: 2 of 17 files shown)


brettwooldridge:
Man this would have made my life easier.  At least I can remove a bunch of code now.

dkoszewnik:
Great!  Thanks for the feedback.  I'll release this as a beta API for now while we explore performance, edge cases, and enhancements.

PR #70, opened 2017-07-24
Reverted FileSystem watcher ... it is known to be broken on Mac OS X.

brettwooldridge:
Sorry for the fire drill.  After some of our unit tests failed with v2.6.5, investigation led to the fact that apparently the Java FileSystem WatcherService code is known to be [broken on Mac OS X -- and potentially problematic on Linux](https://www.reddit.com/r/java/comments/3vtv8i/beware_javaniofilewatchservice_is_subtly_broken/), though we have observed no issues on Linux ourselves.

Reverted to the former standard polling, with minor optimizations.  All of our unit tests now pass on OS X, and continue to pass on Linux.


dkoszewnik:
Thanks @brettwooldridge, this fix is now released in `v2.6.6`.

PR #73, opened 2017-08-01
Allow HollowHashIndexes to listen for updates

ryangardner:
- current implementation does a full refresh of the index when there are changes

### DIFF
--- a/hollow/src/test/java/com/netflix/hollow/core/index/HollowHashIndexTest.java
+++ b/hollow/src/test/java/com/netflix/hollow/core/index/HollowHashIndexTest.java
@@ -18,42 +18,46 @@
 package com.netflix.hollow.core.index;
 
 import com.netflix.hollow.core.AbstractStateEngineTest;
-
-import com.netflix.hollow.core.index.HollowHashIndex;
-import com.netflix.hollow.core.index.HollowHashIndexResult;
-import com.netflix.hollow.core.write.objectmapper.HollowObjectMapper;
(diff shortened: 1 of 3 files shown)


dkoszewnik (approved):

hollow/src/test/java/com/netflix/hollow/core/index/HollowHashIndexTest.java
```diff
@@ -56,6 +56,52 @@ public void test() throws Exception {
...
+        mapper.add(new TypeA(3, 3.3d, new TypeB("three"), new TypeB("thirty"), new TypeB("three hundred")));
+        roundTripDelta();
+
+        HollowHashIndexResult result2 = index.findMatches(2, "forty");
+        HollowOrdinalIterator iter2 = result.iterator();
+
+        int matchedOrdinal2 = iter2.next();
+
+        while(matchedOrdinal2 != HollowOrdinalIterator.NO_MORE_ORDINALS) {
+            System.out.println(matchedOrdinal2);
```
dkoszewnik: I think these calls to System.out.println should be Assert statements.  

It appears that the prior test in this class was also written as an exploratory test -- but then the step to convert it to an automated unit test was missed.

ryangardner: No problem. I'll update both tests with assertions

dkoszewnik: Awesome, thanks @ryangardner

PR #74, opened 2017-08-03
Duplicate detection validator - initial version

lkancode:
Validator to detect duplicate keys based on Primary Key definition for an object. This is initial version. Will add more details soon.


hollow/src/main/java/com/netflix/hollow/core/validator/DuplicateDataDetectionValidator.java
```diff
@@ -0,0 +1,86 @@
+package com.netflix.hollow.core.validator;
```
dkoszewnik: Can we put this in a package more closely aligned with the Producer API?  For example `com.netflix.hollow.api.producer.validation`.

dkoszewnik (changes requested):
Looks great!  But can we please use a package under `api`?

dkoszewnik (approved):

PR #77, opened 2017-08-09
Feature/HollowSparseIntegerSet

kineshsatiya:


kineshsatiya:
Summary of the changes

- HollowSparseIntegerSet
This class can be used to build indexes based on a predicate to index sparse integer values. The underlying data structure is SparseBitSet (inner class). Changes include unit tests for both the classes.

- FieldPathUtil
Added a utility class that reads a field path, calculates the field types, positions and verifies that they are valid. It also has a utility method that is generic, which helps in getting all the values pointed by the field path, given the parent type. Also this code can be reused across multiple indexes.

- HollowPrefixIndex
Updated the code to use FieldPathUtil.

PR #79, opened 2017-08-15
Init flag value for producer initializer

kination:
Init isInitialized as false when instance are being create.


dkoszewnik:
Boolean fields are implicitly false, and stylistically we prefer not explicitly declaring this.

kination:
@dkoszewnik Ok. Thanks!

PR #81, opened 2017-08-16
Fixes #80 Ensure that the RefreshListener is called for every ...

### LINKED ISSUE #80: Unexpected/incorrect callback order for HollowConsumer.RefreshListener during init
We have a ``RefreshListener`` registered with the ``HollowConsumer``.  We are implementing ``snapshotUpdateOccurred()`` and ``deltaUpdateOccurred()``, whereby we act upon add/modify/delete of objects based on the ``populatedOrdinals``/``previousOrdinals``.

When starting from a clean state (no objects in hollow), everything works as expected.  The server produces a snapshot and announces it, then the client processes it and calls ``snapshotUpdateOccurred()``.  The next change on the server produces a delta, which is announced, and the client processes it and calls ``deltaUpdateOccurred()``.  So far, so good.

The trouble occurs when we restart the client.

We construct the ``HollowConsumer`` and call ``triggerAsyncRefresh()``.  In order to restore the state from the locally persisted files, the consumer ends up calling ``HollowDataHolder.applyInitialTransitions()``.  [That method](https://github.com/Netflix/hollow/blob/master/hollow/src/main/java/com/netflix/hollow/api/client/HollowDataHolder.java#L97) ends up applying all of the transitions, starting with the snapshot then proceeding through the deltas, and only after all of that [calls ``snapshotUpdateOccurred()``](https://github.com/Netflix/hollow/blob/master/hollow/src/main/java/com/netflix/hollow/api/client/HollowDataHolder.java#L120) with the *final* version arrived at (which is actually a delta, not a snapshot, btw).

At that point, we have missed our chance to process the transitions via ``populatedOrdinals``/``previousOrdinals``, because the ``previousOrdinals`` only contains those of the final transition.

What we *need* is for the ``RefreshListener`` to be called for each transition, just as if they were occurring at runtime.

brettwooldridge:
... transition when applying initial transitions.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/client/HollowDataHolder.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/client/HollowDataHolder.java
@@ -96,42 +96,35 @@ public class HollowDataHolder {
 
     private void applyInitialTransitions(HollowUpdatePlan updatePlan) throws Throwable {
         for(HollowConsumer.Blob transition : updatePlan) {
-            InputStream is = transition.getInputStream();
-
-            try {
+            try(InputStream is = transition.getInputStream()) {
                 applyTransition(is, transition);
+
+                for(HollowConsumer.RefreshListener refreshListener : refreshListeners) {
+                    if(transition.isSnapshot()) {
+                        if(objLongevityConfig.enableLongLivedObjectSupport()) {
+                            HollowProxyDataAccess dataAccess = new HollowProxyDataAccess();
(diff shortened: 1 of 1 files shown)


dkoszewnik:
Hi @brettwooldridge please see comments on #80 -- this proposal would unfortunately break our implementations of `RefreshListener`.

We're hopeful that you can provide feedback about how we might reduce confusion around the distinction between the `RefreshListener` methods `blobLoaded` and `(snapshot|delta)UpdateOccurred`

PR #82, opened 2017-08-17
Fixes #80 Ensure that the RefreshListener is called for every transit…

### LINKED ISSUE #80: Unexpected/incorrect callback order for HollowConsumer.RefreshListener during init

brettwooldridge:
…ion when applying initial transitions.

@dkoszewnik Of course another option, non-breaking, would be to keep existing ``snapshotUpdateOccurred()`` behavior and introduce new methods that are called on every transition.  *The challenge here is in the naming...*

In the *"who reads javadocs?"* world, in which I admit residing, based on the existing method naming of ``RefreshListener``, I *thought* that I understood the contract.  The fact that the behavior was not evident in the naming is, to this coder, a code smell.

But as noted above, a non-breaking option would be to introduce additional methods.  That is encompassed in this pull request.

As an aside, is there any intention to move to Java 8?  Declaring all of the methods as ``default`` with empty bodies would make transitions like this easier.  Instead, I have added an ``AbstractRefreshListener`` with empty method bodies -- useful not only internally but also for users (but better still would be Java 8).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 **/build/
 **/bin/
+**/out/
 **/.classpath
 **/.project
 **/.settings
--- a/hollow-diff-ui/src/main/java/com/netflix/hollow/history/ui/HollowHistoryRefreshListener.java
+++ b/hollow-diff-ui/src/main/java/com/netflix/hollow/history/ui/HollowHistoryRefreshListener.java
@@ -17,14 +17,14 @@
  */
 package com.netflix.hollow.history.ui;
 
+import com.netflix.hollow.api.consumer.HollowConsumer;
 import com.netflix.hollow.api.consumer.HollowConsumer.Blob;
-import com.netflix.hollow.api.consumer.HollowConsumer.RefreshListener;
 import com.netflix.hollow.api.custom.HollowAPI;
 import com.netflix.hollow.core.read.engine.HollowReadStateEngine;
 import com.netflix.hollow.tools.history.HollowHistory;
 
-public class HollowHistoryRefreshListener implements RefreshListener {
-	
+public class HollowHistoryRefreshListener extends HollowConsumer.AbstractRefreshListener {
+
     private final HollowHistory history;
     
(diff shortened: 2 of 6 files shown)


brettwooldridge:
@dkoszewnik Updated.  Now with 100% less API breakage.

dkoszewnik:
Released in `v2.7.1`.  Thanks again @brettwooldridge.

PR #84, opened 2017-08-22
Enhance field path

kineshsatiya:
Updated FieldPath class with following changes
- Changed the functionality of field path when referencing a Map schema. It does not auto-expand the field path. When using a Map type in field path, it is necessary to add "key" or "value" in the path.
- Added a flag to turn-off auto-expand
- Updated documentation for the class
- Added many unit tests for testing the functioning of FieldPath.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/index/FieldPath.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/index/FieldPath.java
@@ -1,8 +1,27 @@
+/*
+*
+*  Copyright 2017 Netflix, Inc.
+*
+*     Licensed under the Apache License, Version 2.0 (the "License");
+*     you may not use this file except in compliance with the License.
+*     You may obtain a copy of the License at
+*
+*         http://www.apache.org/licenses/LICENSE-2.0
+*
+*     Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


hollow/src/main/java/com/netflix/hollow/core/index/FieldPath.java
```diff
@@ -203,38 +312,47 @@ protected String getLastRefTypeInPath() {
...
                 }
+                return new Object[]{};
+            } else {
+                return new Object[]{readFromObject(objectTypeDataAccess, ordinal, fieldTypes[fieldIndex], fieldPositions[fieldIndex])};
             }
         }
         return values;
     }
+
+    private Object readFromObject(HollowObjectTypeDataAccess objectTypeDataAccess, int ordinal, FieldType fieldType, int fieldPosition) {
```
davesu, line 331: This seems like common code and it does not seem to handle checking for special values.

See HollowReadFieldUtils.fieldValueObject to determine whether it makes sense to reuse that.

kineshsatiya: Hi David, will check up on this, if possible I will reuse this.

PR #85, opened 2017-08-22
HollowProducer.ValidationConfig proposal - for disabling/enabling Validators at runtime

dkoszewnik:


ghost:
Closing this.

Our current intention (that needs documenting) is for users to leverage the Producer event rework where validators are just specialized kinds of producer listeners. For instance, a validator implementation can implement both `ValidatorListener` and `CycleListener`. Then in `onCycleStart` it can check runtime properties (or similar configuration source) and freeze it's current thresholds or validation configs for the remainder of the cycle and to be used in `onValidate`

PR #86, opened 2017-08-23
Feature/boolean field ergonomics

davesu:
Introduce boolean field ergonomics for generated client api

- created accessors without "get" prefix when boolean fields has the following prefix: "is", "has", "do", "should", "was", "contains", "enable", "disable", "get"

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/codegen/HollowCodeGenerationUtils.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/codegen/HollowCodeGenerationUtils.java
@@ -310,7 +310,8 @@ public class HollowCodeGenerationUtils {
      *
      *      boolean isPrimary - isPrimary()
      *      boolean hasStreams - hasStreams()
-     *      boolean playable - isPlayable()
+     *      boolean playable - getPlayable()
+     *      boolean value - getValue()
      *
      * other field type: prepend "get" + upper case first char
      *
--- a/hollow/src/main/java/com/netflix/hollow/api/codegen/HollowCodeGenerationUtils.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/codegen/HollowCodeGenerationUtils.java
@@ -32,21 +32,12 @@ import com.netflix.hollow.core.schema.HollowObjectSchema;
 import com.netflix.hollow.core.schema.HollowObjectSchema.FieldType;
 import com.netflix.hollow.core.schema.HollowSchema;
 import com.netflix.hollow.core.schema.HollowSetSchema;
-import java.io.File;
-import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import javax.tools.JavaCompiler;
-import javax.tools.ToolProvider;
 
 /**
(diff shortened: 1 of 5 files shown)


hollow/src/main/java/com/netflix/hollow/api/codegen/HollowCodeGenerationUtils.java
```diff
@@ -293,4 +297,49 @@ public static String getJavaScalarType(FieldType fieldType) {
         }
         throw new IllegalArgumentException("Java scalar type is not known for FieldType." + fieldType.toString());
     }
-}
+
+    private static final Set<String> booleanMethodPrefixes = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(
+            "is", "has", "do", "should", "was", "contains", "enable", "disable", "get")));
+    public static Set<String> getBooleanMethodPrefixes() { return Collections.unmodifiableSet(booleanMethodPrefixes); }
```
dkoszewnik: This is already a `Collections.unmodifiableSet()`.

hollow/src/test/java/com/netflix/hollow/api/codegen/HollowAPIGeneratorTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+import com.netflix.hollow.core.write.HollowWriteStateEngine;
+import com.netflix.hollow.core.write.objectmapper.HollowObjectMapper;
+import java.io.File;
+import java.io.IOException;
+import org.junit.Before;
+import org.junit.Test;
+
+public class HollowAPIGeneratorTest {
+    private String baseDir = "/tmp/HollowAPIGeneratorTest";
```
dkoszewnik: Might be best to use `System.getProperty("java.io.tmpdir")` instead of hardcoding `"/tmp"`

hollow/src/main/java/com/netflix/hollow/api/codegen/HollowCodeGenerationUtils.java
```diff
@@ -293,4 +297,49 @@ public static String getJavaScalarType(FieldType fieldType) {
...
+
+    /**
+     * Rules: prepend "get" / "is" + upper case first char of field name
+     *
+     * boolean/Boolean field:
+     *    - has a boolean prefix (@see {@link #booleanMethodPrefixes}), just return it; otherwise, prepend "get" + upper case first char
+     *
+     *      boolean isPrimary - isPrimary()
+     *      boolean hasStreams - hasStreams()
+     *      boolean playable - getPlayable()
```
dkoszewnik, line 313: I wonder if this should also default to `isPlayable()`?

dkoszewnik (approved):
Looks good.  Minor comments.

kineshsatiya (approved):

hollow/src/main/java/com/netflix/hollow/api/codegen/HollowCodeGenerationUtils.java
davesu: I was going with that as well originally but "value"  field became isValue() vs getValue().

hollow/src/test/java/com/netflix/hollow/api/codegen/HollowAPIGeneratorTest.java
davesu: Good point, will do that.  thx

hollow/src/main/java/com/netflix/hollow/api/codegen/HollowCodeGenerationUtils.java
davesu: ah, yeah.  forgot to fix that.

PR #91, opened 2017-09-01
Add HollowIncrementalProducer support for withThreadsPerCpu

rpalcolea:
This introduces the `threadsPerCpu ` for `HollowIncrementalProducer`.

```
HollowIncrementalProducer incrementalProducer = new HollowIncrementalProducer(producer, 2.0d);
```

Devs can still use the constructor as before 

```
HollowIncrementalProducer incrementalProducer = new HollowIncrementalProducer(producer);
```

It also introduces the usage of `SimultaneousExecutor` in `addRecords` while running a cycle. By default it uses `1.0d` threads per CPU

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
@@ -27,6 +27,7 @@ import com.netflix.hollow.core.read.engine.HollowTypeReadState;
 import com.netflix.hollow.core.schema.HollowObjectSchema;
 import com.netflix.hollow.core.schema.HollowSchema;
 import com.netflix.hollow.core.schema.HollowSchema.SchemaType;
+import com.netflix.hollow.core.util.SimultaneousExecutor;
 import com.netflix.hollow.core.write.HollowTypeWriteState;
 import com.netflix.hollow.core.write.objectmapper.RecordPrimaryKey;
 import com.netflix.hollow.tools.traverse.TransitiveSetTraverser;
@@ -46,12 +47,18 @@ public class HollowIncrementalProducer {
 
     private final HollowProducer producer;
     private final ConcurrentHashMap<RecordPrimaryKey, Object> mutations;
+    private final double threadsPerCpu;
     
     public HollowIncrementalProducer(HollowProducer producer) {
+        this(producer, 1.0d);
+    }
+
+    public HollowIncrementalProducer(HollowProducer producer, double threadsPerCpu) {
         this.producer = producer;
+        this.threadsPerCpu = threadsPerCpu;
         this.mutations = new ConcurrentHashMap<RecordPrimaryKey, Object>();
     }
-    
+
     public void restore(long versionDesired, BlobRetriever blobRetriever) {
         producer.hardRestore(versionDesired, blobRetriever);
     }
@@ -69,7 +76,34 @@ public class HollowIncrementalProducer {
     public void delete(RecordPrimaryKey key) {
         mutations.put(key, DELETE_RECORD);
     }
-    
+
+    public static HollowIncrementalProducer.Builder withProducer(HollowProducer producer) {
+        HollowIncrementalProducer.Builder builder = new HollowIncrementalProducer.Builder();
+        return builder.withProducer(producer);
+    }
+
+    public static class Builder {
+        private HollowProducer producer;
+        private double threadsPerCpu = 1.0d; //Default to 1 thread per CPU
+
+        public Builder withProducer(HollowProducer hollowProducer) {
+            this.producer = hollowProducer;
+            return this;
+        }
+
+        public Builder withThreadsPerCpu(double threadsPerCpu) {
+            this.threadsPerCpu = threadsPerCpu;
+            return this;
+        }
+
+        public HollowIncrementalProducer build() {
+            if(producer == null)
+                throw new IllegalArgumentException("HollowProducer must be specified");
+
+            return new HollowIncrementalProducer(producer, threadsPerCpu);
+        }
+    }
+
     public long runCycle() {
         return producer.runCycle(new Populator() {
             public void populate(WriteState newState) throws Exception {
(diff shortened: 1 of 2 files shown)


dkoszewnik (approved):

ghost (changes requested):
The feature looks good.

Can you revert some of the stylistic changes? Looks perhaps like automatic *Format Code* changes from an IDE. For instance, some whitespace changes and the introduction of an explicit `final` keyword on effectively final variables. I'll point out a few examples with line comments.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
```diff
@@ -86,10 +121,10 @@ private void removeRecords(WriteState newState) {
 
             private Map<String, BitSet> findTypesWithRemovedRecords(ReadState readState) {
                 Map<String, BitSet> recordsToRemove = new HashMap<String, BitSet>();
-                for(RecordPrimaryKey key : mutations.keySet()) {
-                    if(!recordsToRemove.containsKey(key.getType())) {
+                for (RecordPrimaryKey key : mutations.keySet()) {
```
ghost: Though unconventional, we prefer omitting the space between `for`, `if`, `while`, etc. and the following open paren.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
```diff
@@ -100,49 +135,63 @@ private void removeRecords(WriteState newState) {
...
+            private void markTypeRecordsToRemove(HollowReadStateEngine priorStateEngine, final String type, final BitSet typeRecordsToRemove) {
                 HollowTypeReadState priorReadState = priorStateEngine.getTypeState(type);
                 HollowSchema schema = priorReadState.getSchema();
-                if(schema.getSchemaType() == SchemaType.OBJECT) {
-                    HollowPrimaryKeyIndex idx = new HollowPrimaryKeyIndex(priorStateEngine, ((HollowObjectSchema) schema).getPrimaryKey()); ///TODO: Should we scan instead?  Can we create this once and do delta updates?
-                    
-                    for(Map.Entry<RecordPrimaryKey, Object> entry : mutations.entrySet()) {
-                        if(entry.getKey().getType().equals(type)) {
+                if (schema.getSchemaType() == SchemaType.OBJECT) {
+                    final HollowPrimaryKeyIndex idx = new HollowPrimaryKeyIndex(priorStateEngine, ((HollowObjectSchema) schema).getPrimaryKey()); ///TODO: Should we scan instead?  Can we create this once and do delta updates?
```
ghost: Avoid explicit `final` on local variables unless required for inner classes.

We *do* prefer `final` on immutable fields, however.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
```diff
@@ -27,49 +27,84 @@
...
 public class HollowIncrementalProducer {
-    
+
     private static final Object DELETE_RECORD = new Object();
 
     private final HollowProducer producer;
     private final ConcurrentHashMap<RecordPrimaryKey, Object> mutations;
-    
+
+    private double threadsPerCpu;
```
ghost: This field can be `final` since the builder is how you mutate the config.

ghost (changes requested):
Added a few specific review comments (still learning the right way to use GitHub's review feature).

hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
```diff
@@ -36,40 +37,73 @@
...
         mutations.put(key, DELETE_RECORD);
     }
-    
+
+    public static HollowIncrementalProducer.Builder withProducer(HollowProducer producer) {
+        HollowIncrementalProducer.Builder builder = new HollowIncrementalProducer.Builder();
+        return builder.withProducer(producer);
+    }
+
+    public static class Builder {
```
kineshsatiya: Do we need a builder just for one additional argument? Also, constructor for creating incremental producer is public. Just in my opinion does not justify builder pattern. Consider removing it.

rpalcolea:
Thanks for the feedback @toolbear!

I fixed the formatting and cleaned up the `final` attributes

hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
```diff
@@ -69,6 +76,33 @@ public void delete(Object obj) {
     public void delete(RecordPrimaryKey key) {
         mutations.put(key, DELETE_RECORD);
     }
+
+    public static HollowIncrementalProducer.Builder withProducer(HollowProducer producer) {
+        HollowIncrementalProducer.Builder builder = new HollowIncrementalProducer.Builder();
+        return builder.withProducer(producer);
+    }
+    
+    public static class Builder {
```
kineshsatiya: Do we need a builder just for one additional argument? Also, constructor for creating incremental producer is public. Just in my opinion it does not justify builder pattern. Consider removing it.

kineshsatiya (changes requested):

hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
rpalcolea: @kineshsatiya I added the `Builder` pattern to keep consistence with the way `HollowProducer` and `HollowConsumer` are built. It's a fair point to say that it isn't enough to use the pattern. 

Regarding the constructor, I kept it public to avoid breaking existing consumers if they upgrade once this is merged. Since it's a experimental API, I think it makes sense to make the constructors private since breaking changes could happen

My question would be, Do you want to use the builder pattern or just have the constructors and remove the builder pattern?

rpalcolea: Just changed the visibility to `protected` for the constructor

kineshsatiya: I see why you added the `Builder` pattern. I think for now having a public constructor with `threadsPerCpu` will suffice. If in future we add more build parameters then we can add the `Builder` pattern. Keeping it simple.

rpalcolea: Makes sense @kineshsatiya 

I'll modify the code and let you know

rpalcolea: @kineshsatiya Applied your suggestion. Thanks for the feedback

rpalcolea:
ready for review again @toolbear @kineshsatiya

kineshsatiya (approved):

rpalcolea:
ready for review again @toolbear @dkoszewnik

Issue #93, opened 2017-09-06
Delta / Snapshot cleanup [question]


rpalcolea:
I brought this last month on Gitter:

> Hi, curious about how people is managing historical of snapshots.
> We’ve been running a hollow producer for about 2 months and hit 1.27 TiB of snapshots.

On that time 2 of the chat members replied with some ideas to clean this up: 3 lifecycle after few days of use or  lambda on a timer to manage our s3 with a ttl on last updated and keeping the newest few snapshots.

While I like the idea of having something external like S3 lifecycle or a Google Cloud Function, there's always the risk of losing files since this might not be aware of the usage by Hollow.

I was wondering is there is a desire to have something built in the hollow producer to remove old deltas and snapshots. Allow the producer to remove files from the blob storage every time there's a new snapshot or delta and let users to set the numbers of files to keep or just let them plug their own implementation so they can determine if they want to keep 'x' number of files or based on dates.

Something in the lines of `BlobStorageCleaner`.

```java
MyBlobStorageCleaner blobStorageCleaner = new MyBlobStorageCleaner();
HollowProducer producer = HollowProducer.withBlobStorageCleaner(blobStorageCleaner);
```

The `BlobStorageCleaner` could be 

```java
public abstract class BlobStorageCleaner {
   public void clean(Blob.Type blobType) {
        switch(blobType) {
        case SNAPSHOT:
            cleanSnapshots();
            break;
        case DELTA:
            cleanDeltas();
            break;
        case REVERSE_DELTA:
            cleanReverseDeltas();
            break;
        }
    }
   abstract void cleanSnapshots();
   abstract void cleanDeltas();
   abstract void cleanReverseDeltas();
}
```

And this could be triggered after `fireArtifactPublish` is triggered -> https://github.com/Netflix/hollow/blob/master/hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java#L476

The idea would be that developers could implement their own `BlobStorageCleaner` and decide how they want to remove their snapshots, deltas and reverse deltas. Perhaps you want to clean snapshots but you don;t want to clean deltas. Up to them

mafa73:
I would like to have that feature. Currently we have implemented this functionality in the application that produces the snapshots/deltas by keeping x number of files.

rpalcolea:
Any thoughts on this @dkoszewnik @toolbear ?

rpalcolea:
Implemented this in https://github.com/Netflix/hollow/pull/94

PR #94, opened 2017-09-11
Added support for Hollow Producer BlobStorageCleaner

rpalcolea:
This introduces `BlobStorageCleaner` as proposed in https://github.com/Netflix/hollow/issues/93.

By default, the `HollowProducer` uses a `DummyBlobStorageCleaner` since must of the users won't like the idea of a default behavior for removing snapshots.

A `HollowFilesystemBlobStorageCleaner` is provided as an example on how to remove the last `n` snapshots.

The `clean` is triggered after `fireArtifactPublish` is triggered -> https://github.com/Netflix/hollow/pull/94/files#diff-46bae8959f9797f4fdae3d795addb5ccR480


rpalcolea:
Any thoughts on this? :)

brettwooldridge (commented):
I'm a big fan of the newer  java.nio.Path/Paths/Files classes, they can make a lot of this kind of file manipulation and directory traversal cleaner.  Care to see how it looks employing them here?

brettwooldridge:
Overall, I really like this.

rpalcolea:
Updated to use the changes in https://github.com/Netflix/hollow/pull/104

hollow/src/main/java/com/netflix/hollow/api/producer/DummyBlobStorageCleaner.java
```diff
@@ -0,0 +1,33 @@
...
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.hollow.api.producer;
+
+/**
+ * This Dummy blob storage cleaner does nothing
+ */
+public class DummyBlobStorageCleaner extends HollowProducer.BlobStorageCleaner {
```
kineshsatiya: Should this be package private, if its the default implementation that does nothing?

hollow/src/main/java/com/netflix/hollow/api/producer/fs/HollowFilesystemBlobStorageCleaner.java
```diff
@@ -0,0 +1,92 @@
...
+    @Override
+    public void cleanDeltas() { }
+
+    @Override
+    public void cleanReverseDeltas() { }
+
+    private void sortByLastModified(File[] files) {
+        Arrays.sort(files, new Comparator<File>() {
+            public int compare(File f1, File f2) {
+                return Long.compare(f1.lastModified(), f2.lastModified());
```
kineshsatiya: Remove the reversal sort order. It seems that the most recent files will be cleaned up first. The oldest file should be the first to be cleaned up.

hollow/src/test/java/com/netflix/hollow/api/producer/HollowProducerBlobStorageCleanerTest.java
```diff
@@ -0,0 +1,122 @@
...
+        incrementalProducer.addOrModify(new TypeA(3, "three", 1000));
+        incrementalProducer.runCycle();
+        incrementalProducer.addOrModify(new TypeA(4, "two", 100));
+        incrementalProducer.runCycle();
+        incrementalProducer.addOrModify(new TypeA(5, "three", 1000));
+        incrementalProducer.runCycle();
+
+        File[] files = listFiles(HollowProducer.Blob.Type.SNAPSHOT.prefix);
+        List<String> fileNames = getFileNames(files);
+        assert files.length == 5;
```
kineshsatiya, line 78: The unit test should use Junit Assert. Can you switch them?

kineshsatiya (changes requested):

hollow/src/main/java/com/netflix/hollow/api/producer/fs/HollowFilesystemBlobStorageCleaner.java
rpalcolea: Tackled in another way:  modified the for loop to always keep the latest 'n' files -> https://github.com/Netflix/hollow/pull/94/files#diff-4dbcd2ad1aede1e6ad3380cca689a0a0R58

Added a custom version minter to have control over the test and make sure it doesn't keep the first snapshot -> https://github.com/Netflix/hollow/pull/94/files#diff-b7f49725b4f12dbaf20d35b021c1fea2R91

hollow/src/test/java/com/netflix/hollow/api/producer/HollowProducerBlobStorageCleanerTest.java
rpalcolea: done, thanks for point this out

rpalcolea:
@kineshsatiya addressed your comments, thanks for the feedback. Ready for review again

hollow/src/main/java/com/netflix/hollow/api/producer/DummyBlobStorageCleaner.java
rpalcolea: Moved this into the `HollowProducer` as a private class. Yes it is the default implementation that does nothing.

ghost:
Reviewing latest changes now, @rpalcolea.

Regarding @brettwooldridge comments on `java.nio`, I'm also a big fan of the `Path/Paths/Files` apis. The current Hollow codebase is very `java.io` oriented currently. I would be receptive to a separate PR that consistently moved to `java.nio` APIs (while maintaining `java.io` compatibility in public APIs). In the context of this PR, in some ways a piecemeal adoption of `java.nio` for files/paths would be worse.

hollow/src/main/java/com/netflix/hollow/api/producer/fs/HollowFilesystemBlobStorageCleaner.java
ghost: Love the custom version minter for unit test.

Since you're keeping the reverse sort order, you can avoid sorting twice by altering the custom comparator slightly to `compare(f2.lastModified(), f1.lastModified());`, i.e. swap the order from (f1,f2) to (f2,f1).

It's a subtle difference, but in my experience relatively idiomatic.

ghost (changes requested):

Issue #76, opened 2017-08-07
Hollow Metrics


rpalcolea:
Hi, 

Currently we’re running background jobs with quartz or simple future scheduling  on spring and ratpack to get metrics such as domain count object, heap used, refresh failed and so on and expose them via Prometheus. This has helped us to build grafana dashboards (screenshot attached)

<img width="1435" alt="screen shot 2017-08-06 at 8 36 52 pm" src="https://user-images.githubusercontent.com/1625920/29009286-9b8e592c-7ae7-11e7-9905-ad0d1504d4ec.png">

I'm opening this issue because I think it would be nice to have metrics out of the box from hollow and to see if the following  could fit on your design principles and contribute with it:

1. Introduce a `HollowMetrics` object as part of `HollowClientUpdater` to keep metrics for things such as domain object count by type (`readEngine.typeStates`), heap usage (`typeState.getApproximateHeapFootprintInBytes()`), current version, refresh succeeded/failed.  Every time a refresh fails or it's completed,  the metrics would be updated.

2. Introduce a `HollowMetricsCollector` interface: This could be part of the Consumer API ex. `HollowConsumer.withMetricsCollector(myMetricsCollector)`. Provide this as part of the `HollowClientUpdater` and invoke a method ex. `collect` when an update happens (refresh fails or succeeds - https://github.com/Netflix/hollow/blob/ecabde9e763ce4a7a6e128c86e3bf4c8e7a34f2c/hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java#L100 / https://github.com/Netflix/hollow/blob/ecabde9e763ce4a7a6e128c86e3bf4c8e7a34f2c/hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java#L104). This way, users could include their own mechanism to use those metrics such as writing to logs or exposing via JMX but that shouldn't be part of Hollow.

All of the above could be part of Hollow core, in addition to this I think it would be nice to have a `hollow-metrics` module (similar to `hollow-ui-explorer` and others). This module could have default implementations for `HollowMetricsCollector` using Micrometer (http://micrometer.io/)... think SLF4J, but for application metrics. Micrometer allows to expose metrics via Prometheus, datadog, atlas, influx, graphite and others.  This could be part of a second iteration.

your thoughts?

dkoszewnik:
[...]

PR #95, opened 2017-09-18
Metrics and metrics collector

rpalcolea:
This is for https://github.com/Netflix/hollow/issues/76

This is only the first phase which contains the implementation for the core metrics.

The external module would be implemented in a 2nd phase


kineshsatiya:
Quick question, any particular reason, `HollowProducerListener` and `HollowUpdateListener` are not being used to collect metrics. We could have default implementations and people could hook them up when creating producers and consumers.

rpalcolea:
That is a great question @kineshsatiya 

When I thought about this, my first idea was to create it independent of the listeners. I didn't know if it would be a good idea to have a default listener and allow users to add their own ones.. or provide a base implementation and let them override it as needed. That's why I created it as a separate thing that always happens, because more likely, everyone would like to have metrics around the consumers and/or producer. 

That was the main driver to go into this direction where metrics are collected always no matter if you add or not listeners. 

Is there a preference for this? I could definitely refactor based on what you think is better for this use case. Also, if you have more feedback around the metrics itself and the collector interface would be great.

rpalcolea:
ready for review @dkoszewnik

dkoszewnik (approved):

rpalcolea:
Updated to use the changes in https://github.com/Netflix/hollow/pull/104

kineshsatiya:
@rpalcolea So the reason I was inclined to use listeners for recording metrics, is so that the code in `HollowProducer` for making the right callbacks at completion of each stage could be reused. 

However, I do not strongly feel about using metrics as listeners, since by default collecting those metrics might be a good idea, compared to providing a choice. So going ahead and merging this change.

costimuraru:
@rpalcolea nice job on adding the metrics.

I'm getting an error though on:
`HollowProducerMetrics.java:48`
```java
HollowReadStateEngine hollowReadStateEngine = producerStatus.getReadState().getStateEngine();
```
It seems the `producerStatus.getReadState()` is null.

```java
java.lang.NullPointerException
	at com.netflix.hollow.api.metrics.HollowProducerMetrics.updateCycleMetrics(HollowProducerMetrics.java:48)
	at com.netflix.hollow.api.producer.HollowProducer.runCycle(HollowProducer.java:299)
	at com.netflix.hollow.api.producer.HollowProducer.runCycle(HollowProducer.java:299)
```

rpalcolea:
Mm interesting, I wouldn’t expect the read state to be null.

I’ll open a pr soon to check for this.

Any details you could share on this? And specific state?

Thanks

rpalcolea:
Opened a PR to fix this @costimuraru  -> https://github.com/Netflix/hollow/pull/110

costimuraru:
Thanks for quick response, @rpalcolea!
I'm a bit puzzled myself onto why the state is null (it happens when I run the first runCycle on the producer) - I'll have to dig in.

hollow

rpalcolea:
@costimuraru 

Yeah, looks like it is a valid scenario https://github.com/Netflix/hollow/blob/master/hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducerListener.java#L209

Totally omitted that possibility. Hopefully the fix will be merged and released soon!

My bad 😔

costimuraru:
Cool, mystery solved :D
Thanks! Keep up the great work.

PR #103, opened 2017-09-26
Feature/codegen datachange api

davesu:
Generate [Type]DataAccessor which supports the following:

Constructor:
-  [Type]DataAccessor(HollowConsumer consumer, String type) 
-  ...
-  [Type]DataAccessor(HollowReadStateEngine rStateEngine, String type, PrimaryKey primaryKey)

API:
-   public abstract [Type] getRecord(int ordinal);
-   public Collection<[Type]> getAllRecords();
-   public Collection<[Type]> getAddedRecords();
-   public Collection<[Type]> getRemovedRecords();
-   public Collection<UpdatedRecord<[Type]>> getUpdatedRecords();


hollow/src/main/java/com/netflix/hollow/api/codegen/api/HollowDataAccessorGenerator.java
```diff
@@ -0,0 +1,114 @@
...
+        genConstructors(builder);
+        genPublicAPIs(builder);
+
+        builder.append("}");
+
+        return builder.toString();
+    }
+
+    protected void genConstructors(StringBuilder builder) {
+        builder.append("    public " + classname + "(HollowConsumer consumer, String type) {\n");
```
dkoszewnik: Why do we need to specify a type here -- isn't the type implied based on the schema from which this is generated?

davesu: You are right, I don't need that in generated code, I was just blindly overriding base class ;)

Good catch.

ghost:
I'm late to the review (it's already merged), but I'll share my idea anyway...

What about using an inner class for each data accessor on its corresponding type, e.g. `new Movie.DataAccessor(consumer)` instead of `new MovieDataAccessor`? Ultimately it's a stylistic choice. I like the inner class because it declutters the generated API and avoids the need to create a sub-package to house the data accessors. If we followed that pattern for other classes (like primary key indexes?) it groups all the classes for the type together instead of having each spread over a package hierarchy.

PR #111, opened 2017-10-03
Keeping constructor for HollowConsumer backwards incompatible 

kineshsatiya:
In order to keep constructor for HollowConsumer backwards incompatible, we are passing a null for HollowMetricsCollector. HollowMetricsCollector was recently added to hollow. It used by HollowClientUpdater as a way to use metrics collected by consumer. Passing null is quick fix for now to fix breaking changes (HollowClientUpdater has null checks before using the collector). There will be quick follow up to replace null with a default implementation for HollowMetricsCollector.


rpalcolea:
oops :( my bad

davesu (approved):

PR #116, opened 2017-10-06
Singleproducerenforcer

r6b61756c:
Hook for HollowProducer to prevent multiple active producers in a distributed system


hollow/src/main/java/com/netflix/hollow/api/producer/enforcer/BasicSingleProducerEnforcer.java
```diff
@@ -0,0 +1,38 @@
...
+ *     limitations under the License.
+ *
+ */
+package com.netflix.hollow.api.producer.enforcer;
+
+public class BasicSingleProducerEnforcer extends AbstractSingleProducerEnforcer {
+    private boolean isPrimary = true;
+
+    @Override
+    protected void _enable() {
```
kineshsatiya, line 24: Any reason "_" is the prefix for the method names?

PR #120, opened 2017-10-16
Add validation listener to provide more visibility into validations per cycle

lkancode:
- Add Hollow validation listener
- Add validation status and per validator status per cycle
- Modify system validators to add more information in toString method for this iteration
- Surface each validator information in ValidationStatus which is passed to the listener


hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidatorStatus.java
```diff
@@ -0,0 +1,147 @@
...
+ * In next iteration this might be directly returned by validators.
+ */
+class ValidatorStatus {
+    private final long version;
+    private final Status status;
+    private final String message;
+    private final Throwable throwable;
+    private final Map<String, String> additionalInfo;
+
+	@Generated("SparkTools")
```
kineshsatiya: "SparkTools" seems unfamiliar, can you add some comment how we use this?

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -178,19 +181,22 @@ protected HollowProducer(BlobStager blobStager,
             int numStatesBetweenSnapshots,
             long targetMaxTypeShardSize,
             HollowMetricsCollector<HollowProducerMetrics> metricsCollector) {
-        this(blobStager, publisher, announcer, validators, listeners, versionMinter, snapshotPublishExecutor, numStatesBetweenSnapshots, targetMaxTypeShardSize, metricsCollector, new DummyBlobStorageCleaner(), new BasicSingleProducerEnforcer());
+        this(blobStager, publisher, announcer, validators, listeners, Collections.<HollowValidationListener>emptyList(), versionMinter, snapshotPublishExecutor, numStatesBetweenSnapshots, targetMaxTypeShardSize, metricsCollector, new DummyBlobStorageCleaner(), new BasicSingleProducerEnforcer());
```
kineshsatiya, line 187: Do we expect multiple validation listeners? I think one per producer should suffice.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -982,6 +999,17 @@ public Builder withListeners(HollowProducerListener... listeners) {
             return this;
         }
         
+        public Builder withValidationListener(HollowValidationListener listener) {
```
kineshsatiya: Probably no needed, since there is an existing method with var args.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/HollowValidationListener.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+public interface HollowValidationListener extends EventListener{
+	
+
+    /**
+     * Called when the {@code HollowProducer} has begun validating the new data state produced this cycle.
+     *
+     * @param version Version to be validated
+     */
+    public void onValidationStart(long version);
```
kineshsatiya, line 44: I'd recommend to use the existing `HollowProducerListener` to include `ValidationStatus`. The methods defined in the interface `HollowValidationListener` are already defined with the same intent.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidationStatus.java
```diff
@@ -0,0 +1,151 @@
...
+	@Generated("SparkTools")
+	public static Builder builder() {
+		return new Builder();
+	}
+
+	/**
+	 * Builder to build {@link ValidationStatus}.
+	 */
+	@Generated("SparkTools")
+	public static final class Builder {
```
kineshsatiya: This should probably be private.

kineshsatiya (changes requested):

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
lkancode: I followed same pattern as Producer listeners. Is there any downside to allowing multiple?

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
lkancode: Just making it easy to register a single listener at a time instead of collecting all into one call. This pattern is similar to the HollowProducer listener method in builder.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/HollowValidationListener.java
lkancode: As we discussed producer listener is going to provide visibility into various steps in a cycle. Validation listener is going to provide more details on validation. And in next iteration there is proposal to provide even more events like per validator start/ end. For this reason validation having its separate listener I think will be valuable.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidationStatus.java
lkancode: I think providing a way to build ValidationStatus using a builder is valuable. Is there a specific reason why this should be private?

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidatorStatus.java
lkancode: It is just indicating a plugin I used to generate the code. As far as I know this annotation does not impact the behavior. I will add comment explaining or remove it.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidationStatus.java
kineshsatiya: Sorry I meant package-private. I don't see the use case for creating a this builder outside of Hollow. I see that the constructor is already private, so probably fine. Also, the constructor for `ValidationStatus` is public. Can you make that private?

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidatorStatus.java
kineshsatiya: We should probably not check this in.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
kineshsatiya: If you don't expect multiple validation listener, then it should probably not be a list.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidatorStatus.java
lkancode: I checked the @Generate annotation. The documentation clearly states that it is used to indicate generated code  and a tool name can be provided. With that documentation I do not feel this annotation is an issue.

kineshsatiya (approved):

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -178,19 +181,22 @@ protected HollowProducer(BlobStager blobStager,
...
-        this(blobStager, publisher, announcer, validators, listeners, versionMinter, snapshotPublishExecutor, numStatesBetweenSnapshots, targetMaxTypeShardSize, metricsCollector, new DummyBlobStorageCleaner(), new BasicSingleProducerEnforcer());
+        this(blobStager, publisher, announcer, validators, listeners, Collections.<HollowValidationListener>emptyList(), versionMinter, snapshotPublishExecutor, numStatesBetweenSnapshots, targetMaxTypeShardSize, metricsCollector, new DummyBlobStorageCleaner(), new BasicSingleProducerEnforcer());
     }
 
     protected HollowProducer(BlobStager blobStager,
                              Publisher publisher,
                              Announcer announcer,
                              List<Validator> validators,
                              List<HollowProducerListener> listeners,
+                             List<HollowValidationListener> vallisteners,
```
davesu: minor but might be good idea to use camel case like valListeners

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -630,15 +640,18 @@ private void applyDelta(Blob blob, HollowReadStateEngine stateEngine) throws IOE
 
     private void validate(HollowProducer.ReadState readState) {
         ProducerStatus.Builder status = listeners.fireValidationStart(readState);
+        ValidationStatus.Builder valStatus = ValidationStatus.builder().withReadState(readState).withVersion(readState.getVersion());
         try {
             List<Throwable> validationFailures = new ArrayList<Throwable>();
             
             for(Validator validator : validators) {
+            	Throwable throwable = null;
```
davesu: Wonder whether it might be better to have validate return a result instead of throwing exception.  Something to consider for the future.   Returning Result seems a bit more structured than throw exception and make sure validator.toString() has enough context.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
davesu: yeah, existing pattern no so good.

When there is an method that takes "..." params, it already supports 0 or more so adding specific method to take 1 does not seem necessary.    Looks like we need to cleanup all this extra methods once we release Hollow 3.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducerListener.java
```diff
@@ -482,6 +482,6 @@ public Throwable getCause() {
     }
 
     public enum Status {
-        SUCCESS, FAIL
+        SUCCESS, FAIL, SKIP
```
davesu, line 485: Hum, is SKIP specific to validation?  Or is it also applicable to general HollowProducerListeners?  In that case, does it mean that the cycle was skipped?

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -32,13 +32,14 @@
 
 /**
  *
```
davesu, line 36: Might be good idea to provide some javadoc on what this validator does.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -64,17 +65,33 @@ public DuplicateDataDetectionValidator(String dataTypeName, String[] fieldPathNa
...
-		log.log(Level.INFO, "Running DuplicateDataDetectionValidator for type "+dataTypeName);
-		PrimaryKey primaryKey = getPrimaryKey(readState);
-		HollowPrimaryKeyIndex hollowPrimaryKeyIndex = new HollowPrimaryKeyIndex(readState.getStateEngine(), primaryKey);
+		Builder statusBuilder = initializeForValidation(readState);
+		
+		PrimaryKey primaryKey = getPrimaryKey(readState, statusBuilder);
+		String fieldPaths = Arrays.toString(primaryKey.getFieldPaths());
+		statusBuilder.addAdditionalInfo(FIELD_PATH_NAME , fieldPaths);
+		
+		HollowPrimaryKeyIndex hollowPrimaryKeyIndex = new HollowPrimaryKeyIndex(readState.getStateEngine(), primaryKey );
```
davesu: Wonder if it should try to use existing before creating new one for example: HollowPrimaryKeyIndex primaryKeyIndex = typeState.getListener(HollowPrimaryKeyIndex.class);

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -630,15 +640,18 @@ private void applyDelta(Blob blob, HollowReadStateEngine stateEngine) throws IOE
...
                 try {
                     validator.validate(readState);
```
davesu: Validate.validate interface does not seem to indicate it should throw exception on failure.   Might be good idea to make it clearer.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -85,23 +102,42 @@ private String getDuplicateIDsString(Collection<Object[]> dupKeysCollection) {
...
-					"Primary key validation defined but unable to find primary key for data type " + dataTypeName);
+			handleEndValidation(statusBuilder, Status.FAIL,  String.format(NO_PRIMARY_KEY_ERRRO_MSG_FORMAT, dataTypeName));
 
 		return primaryKey;
 	}
-
+	
+	private void handleEndValidation(Builder statusBuilder, Status status, String message) {
+		statusBuilder.withStatus(status);
+		if(message != null){
```
davesu: Wonder if this can be improved since without context and just reading this code it feels a bit odd to throw exception if message is not null without checking on the status.    E.g. status == SUCCESS and message != null.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/RecordCountVarianceValidator.java
```diff
@@ -52,33 +52,71 @@ public RecordCountVarianceValidator(String typeName, float allowableVariancePerc
...
+		if(throwException){
+			validationException = new ValidationException(message);
+			builder.withThrowable(validationException);
+			this.status = builder.build();
+			throw validationException;
+		}
+		this.status = builder.build();
+	}
+	
+	private static final String DATA_TYPE_NAME = "Typename";
```
davesu: Minor: but should this be at the top of the class?   Don't recall seeing class fields defined at the bottom of class before so thought it might be good to be consistent.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidationStatus.java
```diff
@@ -0,0 +1,151 @@
...
+			return this;
+		}
+		
+		public Builder fail(Throwable th) {
+			this.status = Status.FAIL;
+			this.throwable = th;
+			return this;
+		}
+
+		public Builder withThrowable(Throwable throwable) {
```
davesu: is there a use case for adding a throwable without fail?

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidationStatus.java
```diff
@@ -0,0 +1,151 @@
...
+		}
+
+		public Builder withReadState(HollowProducer.ReadState readState) {
+			this.readState = readState;
+			return this;
+		}
+
+		public Builder addValidatorStatus(Throwable th, String message) {
+			Status success = (th == null)? Status.SUCCESS: Status.FAIL;
+			this.validatorStatusList.add(ValidatorStatus.builder().withStatus(success).withMessage(message).withThrowable(th).withVersion(version).build());
```
davesu: should this method change this.status as well?

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidationStatus.java
```diff
@@ -0,0 +1,151 @@
...
+			return this;
+		}
+
+		public ValidationStatus build() {
+			return new ValidationStatus(this);
```
davesu: Not a big fan of circular dependency ValidationStatus -> Builder -> ValidationStatus

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidatorStatus.java
```diff
@@ -0,0 +1,147 @@
...
+ * NOTE: Beta API subject to change.   *
+ * *************************************
+ * 
+ * @author lkanchanapalli {@literal<lavanya65@yahoo.com>}
+ * 
+ * ValidationStatus has one instance per each validator that ran. 
+ * For now ValidationStatus builds these and sets toString on validator as the message (where validators provide any details).
+ * In next iteration this might be directly returned by validators.
+ */
+class ValidatorStatus {
```
davesu: Wonder whether we can improve the names between ValidationStatus and ValidatorStatus.  They were so close that I thought they were duplicate classes.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidatorStatus.java
davesu: It would be great if the plugin has option to omit this ;)  Similar to Eclipse building code generator toString(), equals() which does not add annotation which keeps things cleaner.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
lkancode: @davesu your point is reverse of @kineshsatiya. I agree with only providing multiple for the new API that I added. I will make that change.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidationStatus.java
lkancode: Based on the package structure either ValidatorStatus or ValidationStatu has to be public. I lean towards adding all these object to Validator package and keeping current declarations.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidatorStatus.java
lkancode: Hmm. But the @generate seems pretty common annotation used in Java specifically for this. Since I don't feel too strongly about this, I will remove it based on more than one person finding this confusing.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
lkancode: Ok. Just was a bit lazy I guess. :)

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
lkancode: Yes, that approach is being considered for next iteration.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducerListener.java
lkancode: My thinking is that indicating skipped status when possible provides richer information. For example, if there are no changes publish, announce, validate phases could certainly say skipped.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
lkancode: Ok.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
lkancode: Based on what I remember that is what the call does underneath. I will double check.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
lkancode: That is true. That is also being considered for next iteration.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
lkancode: Ok.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/RecordCountVarianceValidator.java
lkancode: Typically they are defined at the top and I am not sure why they are defined at the top. I feel these constants clutter the top and distract from the class itself and so moved it down.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidationStatus.java
lkancode: There is no case. Good point. I will recheck and make change as needed.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidationStatus.java
lkancode: At this point these two are being set separately. One is set in producer while the other in ValidationStatus. They happen to be same in all cases. Let me see if there is a way to ensure it.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidationStatus.java
lkancode: It is pure builder pattern as far as I remember. Builder is static. Let me recheck.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidatorStatus.java
lkancode: Yes I felt the same. Suggestions are welcome.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
lkancode: I think that allowing multiple listeners for validation gives more flexibility to users. If there was as reason like performance or some other to restrict listener to 1 then I believe it makes sense. Without that in this case, I feel inclined to leave it as it is.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
lkancode: Done.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
lkancode: Done.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
lkancode: Done. Added some for also RecordCountValidator

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
lkancode: I remembered wrong. Modified to reused if one is existing.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
lkancode: Done.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidationStatus.java
lkancode: Done. Removed withThrowable as it is never used.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidationStatus.java
lkancode: Made some changes. More changes to come to make these interactions clearer.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidationStatus.java
lkancode: Confirmed that there is no circular dependency. It is a static builder returning an object.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidatorStatus.java
lkancode: Made some changes. More changes to come to make interaction between HollowProducer, individual validator status and overall validation status better.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ValidatorStatus.java
lkancode: Done. Removed the annotations.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -647,15 +657,18 @@ private void applyDelta(Blob blob, HollowReadStateEngine stateEngine) throws IOE
 
     private void validate(HollowProducer.ReadState readState) {
         ProducerStatus.Builder status = listeners.fireValidationStart(readState);
+        OverallValidationStatus.OverallValidationBuilder valStatus = OverallValidationStatus.builder().withReadState(readState).withVersion(readState.getVersion());
```
kineshsatiya: Just wondering, if we could use a better name for this one - like CycleValidationStatus

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducerListener.java
kineshsatiya: Are the code changes done to use SKIP when there is nothing to publish/announce?

kineshsatiya (commented):
Added few comments.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducerListener.java
lkancode: No I have not modified other Hollow cycle stages to use Skip. Those changes can be made as needed when working in the area.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -215,6 +224,10 @@ public void execute(Runnable command) {
 
         for(HollowProducerListener listener : listeners)
             this.listeners.add(listener);
+        
+        for(HollowValidationListener vallistener: validationListeners){
```
davesu, line 228: comment is not specified to this change but wonder why we don just do addAll.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -646,33 +659,45 @@ private void applyDelta(Blob blob, HollowReadStateEngine stateEngine) throws IOE
...
-            }
-            
-            status.success();
-        } catch (Throwable th) {
-            status.fail(th);
-            throw th;
-        } finally {
-            listeners.fireValidationComplete(status);
-        }
+    	com.netflix.hollow.api.producer.HollowProducerListener.ProducerStatus.Builder psb = listeners.fireValidationStart(readState);
```
davesu, line 662: minor: might be good idea to use longer name.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -646,33 +659,45 @@ private void applyDelta(Blob blob, HollowReadStateEngine stateEngine) throws IOE
...
+    			try {
+    				validator.validate(readState);
+	    		} catch (Throwable th) {
+	    			throwable = th;
+	    			exceptions.add(th);
+	    		}
+    			valStatus.addSingelValidationStatus(getValidationStatus(readState, validator, throwable));
+    		}
+	    	
+	    	if(!exceptions.isEmpty()) {
```
davesu, line 678: wonder if this logic could be encapsulated a little bit different.  For example,

- have valStatus, keep track of all exceptions
- it will also be responsible for tracking fail/success automatically while processing individual validator results

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -872,10 +897,14 @@ public ValidationException(String msg, Throwable cause) {
                 super(msg, cause);
             }
             
-            public void setIndividualFailures(List<Throwable> individualFailures) {
+            public ValidationException(String msg, List<Throwable> individualFailures) {
+            	super(msg);
+            	this.individualFailures = individualFailures;
```
davesu, line 902: minor: it would be nice to fix alignment ;)

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -999,6 +1029,12 @@ public Builder withListeners(HollowProducerListener... listeners) {
             return this;
         }
         
+        public Builder withValidationListeners(HollowValidationListener... listeners) {
+            for(HollowValidationListener listener : listeners)
```
davesu, line 1033: addAll maybe?

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducerListener.java
davesu: Might be good idea to add it when there is an use case for it if there is none yet.  That way we minimize none used code.

hollow/src/main/java/com/netflix/hollow/api/producer/ListenerSupport.java
```diff
@@ -32,14 +36,20 @@
...
     ListenerSupport() {
         listeners = new CopyOnWriteArraySet<>();
+        validationListeners = new CopyOnWriteArraySet<>();
     }
 
     void add(HollowProducerListener listener) {
         listeners.add(listener);
     }
+    
+    void add(HollowValidationListener listener) {
```
davesu, line 50: Does it need a corresponding remove so it will be balanced similar to HollowProducerListener?

hollow/src/main/java/com/netflix/hollow/api/producer/validation/RecordCountVarianceValidator.java
```diff
@@ -52,33 +56,68 @@ public RecordCountVarianceValidator(String typeName, float allowableVariancePerc
...
+		this.status = builder.withStatus(status).build();
+	}
+	
+	private static final String DATA_TYPE_NAME = "Typename";
+	private static final String ALLOWABLE_VARIANCE_PERCENT_NAME = "AllowableVariancePercent";
+	private static final String LATEST_CARDINALITY_NAME = "LatestRecordCount";
+	private static final String PREVIOUS_CARDINALITY_NAME = "PreviousRecordCount";
+	private static final String ACTUAL_CHANGE_PERCENT_NAME = "ActualChangePercent";
+	private static final String ZERO_PREVIOUS_COUNT_WARN_MSG_FORMAT = "Previous record count is 0. Not running RecordCountVarianceValidator for type %s. "
+																		+"This scenario is not expected except when starting a new namespace." ;
```
davesu, line 119: minor: wonder if this auto breaking the string by IDE is necessary ;)

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
lkancode: Will give it another try in next PR.

lkancode: (saw your comment after merging)

PR #122, opened 2017-10-19
Generated API matches return Type

r6b61756c:


hollow/src/main/java/com/netflix/hollow/api/codegen/HollowAPIGenerator.java
```diff
@@ -170,6 +171,15 @@ public void setUseHollowPrimitiveTypes(boolean useHollowPrimitiveTypes) {
         this.useHollowPrimitiveTypes = useHollowPrimitiveTypes;
     }
 
+    /**
```
davesu, line 188: This only impact Scalar type so might be a good idea to have the javadoc and code to reflect that ;)

hollow/src/main/java/com/netflix/hollow/api/codegen/HollowAPIGenerator.java
```diff
@@ -334,6 +345,15 @@ public Builder reservePrimaryKeyIndexForTypeWithPrimaryKey(boolean reservePrimar
             return this;
         }
 
+        /**
```
davesu, line 364: Wonder if it would be good idea to internally invoke withErgonimicsShortcusts and reflect that in javadoc

hollow/src/main/java/com/netflix/hollow/api/codegen/delegate/HollowObjectDelegateCachedImplGenerator.java
```diff
@@ -173,62 +173,132 @@ public String generate() {
...
-            builder.append("        if(").append(fieldName).append(" == null)\n");
-            builder.append("            return false;\n");
-            builder.append("        return ").append(fieldName).append(".booleanValue();\n");
-            builder.append("    }\n\n");
-            builder.append("    public Boolean get").append(uppercase(fieldName)).append("Boxed(int ordinal) {\n");
-            builder.append("        return ").append(fieldName).append(";\n");
-            builder.append("    }\n\n");
+            if(restrictApiToFieldType) {
+                if(!reference) {
+                    builder.append("    public boolean get").append(uppercase(fieldName)).append("(int ordinal) {\n");
```
davesu: Perhaps we can improve this a bit without duplicating builder.append lines. For example:

if (!restrictApiToFieldType || !reference) {
   ... primitive
}

if (!restrictApiToFieldType || reference) {
   ... boxed
}

hollow/src/main/java/com/netflix/hollow/api/codegen/delegate/HollowObjectDelegateInterfaceGenerator.java
```diff
@@ -38,8 +38,8 @@
  */
 public class HollowObjectDelegateInterfaceGenerator extends HollowObjectDelegateGenerator {
 
-    public HollowObjectDelegateInterfaceGenerator(String packageName, HollowObjectSchema schema, HollowErgonomicAPIShortcuts ergonomicShortcuts, boolean usePackageGrouping, boolean useHollowPrimitiveTypes) {
-        super(packageName, schema, ergonomicShortcuts, usePackageGrouping, useHollowPrimitiveTypes);
+    public HollowObjectDelegateInterfaceGenerator(String packageName, HollowObjectSchema schema, HollowErgonomicAPIShortcuts ergonomicShortcuts, boolean usePackageGrouping, boolean useHollowPrimitiveTypes, boolean restrictApiToFieldType) {
```
davesu: Might be better to leave delegates along and more flexible since they are meant for internal/advance use cases.

hollow/src/main/java/com/netflix/hollow/api/codegen/delegate/HollowObjectDelegateLookupImplGenerator.java
```diff
@@ -39,8 +39,8 @@
  */
 public class HollowObjectDelegateLookupImplGenerator extends HollowObjectDelegateGenerator {
 
-    public HollowObjectDelegateLookupImplGenerator(String packageName, HollowObjectSchema schema, HollowErgonomicAPIShortcuts ergonomicShortcuts, boolean usePackageGrouping, boolean useHollowPrimitiveTypes) {
-        super(packageName, schema, ergonomicShortcuts, usePackageGrouping, useHollowPrimitiveTypes);
+    public HollowObjectDelegateLookupImplGenerator(String packageName, HollowObjectSchema schema, HollowErgonomicAPIShortcuts ergonomicShortcuts, boolean usePackageGrouping, boolean useHollowPrimitiveTypes, boolean restrictApiToFieldType) {
```
davesu: Might be better to leave delegates along and more flexible since they are meant for internal/advance use cases.

hollow/src/main/java/com/netflix/hollow/api/codegen/objects/HollowObjectJavaGenerator.java
```diff
@@ -185,12 +187,18 @@ private String generateReferenceFieldAccessor(int fieldNum) {
...
                 String methodName = (shortcut.getType()==FieldType.BOOLEAN) ? generateBooleanAccessorMethodName(fieldName, useBooleanFieldErgonomics) : "get" + uppercase(fieldName);
 
-                builder.append("    public ").append(HollowCodeGenerationUtils.getJavaBoxedType(shortcut.getType())).append(" ").append(getterPrefix).append(methodName + "Boxed() {\n");
-                builder.append("        return delegate().get" + uppercase(fieldName) + "Boxed(ordinal);\n");
-                builder.append("    }\n\n");
-                builder.append("    public ").append(HollowCodeGenerationUtils.getJavaScalarType(shortcut.getType())).append(" ").append(getterPrefix).append(methodName + "() {\n");
-                builder.append("        return delegate().get" + uppercase(fieldName) + "(ordinal);\n");
-                builder.append("    }\n\n");
+                if(restrictApiToFieldType) {
+                    builder.append("    public ").append(HollowCodeGenerationUtils.getJavaBoxedType(shortcut.getType())).append(" ").append(getterPrefix).append(methodName + "() {\n");
```
davesu: It would be good if this can be done without dupping build.append?

hollow/src/main/java/com/netflix/hollow/api/codegen/delegate/HollowObjectDelegateLookupImplGenerator.java
r6b61756c: That is what I was wondering as well... at which layer do these api changes stop.  I like the idea of not porting the changes all the way into the delegates.

hollow/src/main/java/com/netflix/hollow/api/codegen/delegate/HollowObjectDelegateCachedImplGenerator.java
r6b61756c: This is hopefully a stop gap till V3, when we default to this flag, (and/or separate out the client-api generation).

kineshsatiya (approved):

PR #135, opened 2017-11-17
Better HashIndex behavior with null fields

### LINKED ISSUE #128: Add ability to use a hash index on nullable fields
When creating a hash index on a nullable field, `ArrayOutOfBoundsException` is thrown from `FixedLengthElementArray:94`.

ghost:
* [x] fix exception when building index over fields with null values
* [ ] ~~fix exception when attempting to find matches with null query value~~
* [x] more meaningful exception when attempting to find matches with null query values

Fixes #128.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
@@ -158,6 +158,8 @@ public class HollowHashIndex implements HollowTypeStateListener {
 
                 HollowObjectTypeReadState objectAccess = (HollowObjectTypeReadState)readState;
                 int fieldIdx = fieldPath[fieldPath.length-1];
+                if(hashOrdinal == -1 && query[i] == null)
+                    continue;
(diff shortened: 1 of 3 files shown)


hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
```diff
@@ -158,6 +158,8 @@ private boolean matchIsEqual(FixedLengthElementArray matchHashTable, long hashBu
 
                 HollowObjectTypeReadState objectAccess = (HollowObjectTypeReadState)readState;
                 int fieldIdx = fieldPath[fieldPath.length-1];
+                if(hashOrdinal == -1 && query[i] == null)
```
ghost, line 163: Possibly belongs in `HollowReadFieldUtils.fieldValueEquals` instead.

ghost:
The changes I made for allowing `findMatches` with null query values doesn't cover all cases. Instead of trying to add that as part of this PR, I'm going to instead have a better exception that querying for null isn't supported.

PR #138, opened 2017-11-29
Add validator name, clean up success and failure validator messages.

lkancode:
Add validator name, clean up success and failure validator messages.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/validation/AllValidationStatus.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/validation/AllValidationStatus.java
@@ -32,24 +32,17 @@ import com.netflix.hollow.api.producer.HollowProducerListener.Status;
  *
  */
 public class AllValidationStatus {
-    private final long version;
     private final Status status;
     private final List<SingleValidationStatus> validationStatusList;
 
 	private AllValidationStatus(AllValidationStatusBuilder builder) {
-		this.version = builder.version;
 		this.status = builder.status;
 		this.validationStatusList = builder.validationStatusList;
 	}
 
-	public AllValidationStatus(long version, Status status, List<SingleValidationStatus> validatorStatusList) {
-		this.version = version;
+	public AllValidationStatus(Status status, List<SingleValidationStatus> validatorStatusList) {
 		this.status = status;
 		this.validationStatusList = validatorStatusList;
-	}
-
-	public long getVersion() {
-		return version;
 	}
 
 	public Status getStatus() {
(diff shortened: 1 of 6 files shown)


hollow/src/main/java/com/netflix/hollow/api/producer/validation/AllValidationStatus.java
```diff
@@ -32,26 +32,19 @@
  *
  */
 public class AllValidationStatus {
```
ghost, line 34: I know this class was already added prior to this PR, but wonder if `AggregateValidationStatus` would be a better name. I see in the `toString` you use "OverallValidationStatus" so maybe that's pointing out we want a different name? I think "aggregate" indicates intent well, but even "overall" is an improvement over "all".

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -153,4 +159,9 @@ private void handleEndValidation(SingleValidationStatusBuilder statusBuilder, St
 				+ "is not Object. This validation cannot be done.";
 	private static final String FIELD_PATH_NAME = "FieldPaths";
 	private static final String DATA_TYPE_NAME = "Typename";
+
+	@Override
+	public String getName() {
+		return NAME+"_"+dataTypeName;
```
ghost, line 165: I wonder if instead of needing `Nameable` if what we really want is something like a type-specific validator:

```java
interface TypeSpecificValidator extends Validator {
  String getTypeName();
}
```

Then in the `HollowProducer::validate` construct the name by combining the validator class name with the type name, as you're doing here.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -44,7 +45,9 @@
  * this validator will not fail as the two objects are deduped.
  *
  */
-public class DuplicateDataDetectionValidator implements Validator {
+public class DuplicateDataDetectionValidator implements Nameable, Validator {
+	private final String NAME = "DuplicateDataDetectionValidator";
```
ghost, line 49: Small smell that we're duplicating the class name here. Maybe we can just use short class name instead of a constant string? Or change the mechanism for constructing the name mentioned below).

lkancode: Using short class name works too. The validator can choose an appropriate name to return. I just chose this but point taken.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
lkancode: What you suggest is one way to achieve this visibility. But at this point, leaving it to validators whether to even have a name is fine. In next iteration we can focus on how we would like HollowProducer and validator relation to change.

PR #139, opened 2017-11-30
Minor: change name of overall validation status, messages

lkancode:
Minor: change name of overall validation status, messages


hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -678,7 +678,7 @@ private void applyDelta(Blob blob, HollowReadStateEngine stateEngine) throws IOE
     private void validate(HollowProducer.ReadState readState) {
     	com.netflix.hollow.api.producer.HollowProducerListener.ProducerStatus.Builder psb = listeners.fireValidationStart(readState);
     	List<Throwable> exceptions = new ArrayList<>();
-    	AllValidationStatusBuilder valStatus = AllValidationStatus.builder();
+    	OverAllValidationStatusBuilder valStatus = OverAllValidationStatus.builder();
```
ghost, line 681: Small nitpick, I think overall is one word, so `OverallValidation...` instead?

lkancode: Sounds good. Thank you.

ghost:
We'll stick with the current names for now. This is a breaking change we should have merged in December before much usage. There are enough users (at least internally) of these class names that breaking them without a proper deprecation cycle doesn't have much value. Current names are good enough.

PR #142, opened 2017-12-27
clear mutations in hollow incremental producer between states

rpalcolea:
Hi @toolbear @dkoszewnik 

We found that the mutations map in the hollow inc producer was eating up memory. Since `mutations` is shared between states if you only create one instance of `HollowIncrementalProducer`, you keep all the `mutations` over time and carrying over references to objects from previous cycles, growing memory exponentially. I'm not a garbage collector expert but, do you think this could be causing java to avoid collecting old instances from previous cycles? Since you keep the reference in the `mutations` map, java thinks that they are in use so it doesn't collect them.

As a workaround, we created an instance of `HollowIncrementalProducer` on every cycle but we thought that it would be nice to just cleanup the mutations at the end of `populate` because you still keep several `HollowIncrementalProducer` instances for a while. 

Please let me know your thoughts

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
@@ -81,9 +81,7 @@ public class HollowIncrementalProducer {
         this.mutations.clear();
     }
 
-    protected ConcurrentHashMap<RecordPrimaryKey, Object> getMutations() {
-        return this.mutations;
-    }
+    public boolean hasMutations() { return this.mutations.size() > 0; }
 
     public long runCycle() {
         return producer.runCycle(new Populator() {
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
@@ -77,11 +77,11 @@ public class HollowIncrementalProducer {
         mutations.put(key, DELETE_RECORD);
     }
 
-    public void clearMutations() {
+    public void clearChanges() {
         this.mutations.clear();
     }
 
-    public boolean hasMutations() { return this.mutations.size() > 0; }
+    public boolean hasChanges() { return this.mutations.size() > 0; }
 
     public long runCycle() {
         return producer.runCycle(new Populator() {
@@ -89,7 +89,7 @@ public class HollowIncrementalProducer {
                 newState.getStateEngine().addAllObjectsFromPreviousCycle();
                 removeRecords(newState);
                 addRecords(newState);
-                clearMutations();
+                clearChanges();
             }
 
             private void removeRecords(WriteState newState) {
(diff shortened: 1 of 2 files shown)


hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
```diff
@@ -76,13 +76,22 @@ public void delete(Object obj) {
     public void delete(RecordPrimaryKey key) {
         mutations.put(key, DELETE_RECORD);
     }
-    
+
+    public void clearMutations() {
```
ghost: I like that this is public. A user might decide to abandon a set of mutations previously queued up. This allows them to do that without having to discard the producer to gain that effect.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
```diff
@@ -76,13 +76,22 @@ public void delete(Object obj) {
...
+        this.mutations.clear();
+    }
+
+    protected ConcurrentHashMap<RecordPrimaryKey, Object> getMutations() {
```
ghost: I dislike methods that exist just for unit testing. I don't have an alternative suggestion that I like that much more, however.

What about replacing this with a public `boolean hasMutations()` method? I don't think we lose much by changing the unit test from asserting that size == 8 versus just asserting that it's non-empty.

It seems plausible that a `hasMutations()` method might be useful to users. For instance, if you choose to run a cycle on a cadence you could skip cycles for which there haven't been changes (as opposed to running a cycle that ostensibly won't produce a new data state).

ghost (changes requested):
I like. A few specific suggestions.

Also, I wonder if `clearChanges()` (and `hasChanges`) might serve as good names for the public API. I don't feel strongly either way, though.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
rpalcolea: 👍

hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
rpalcolea: Agree, I wasn't very comfortable with adding a method just for testing. Your suggestion is great, added the `hasMutations` as public also so it could be useful for users as you mentioned.

Also decided to do 

```
    public boolean hasMutations() { return this.mutations.size() > 0; }
```

instead of 

```
    public boolean hasMutations() { return !this.mutations.isEmpty(); }
```

I'm not a big fan of doing a negative condition... wish Java had `isNotEmpty` like Kotlin or other languages 👎 

Let me know your thoughts

rpalcolea: @toolbear ^^

rpalcolea:
Also your suggestion around `changes` instead of `mutations` is very good, cleaner for the public API!

Issue #137, opened 2017-11-21
Confuse on using HollowIncrementalProducer


maiakhoa:
Hi @dkoszewnik 
Can you help me explain this case?
I've tried to rewrite addOrModify method in **HollowIncrementalProducer** like this

```
public void addOrModify(Object obj) {
    RecordPrimaryKey pk = newState.getObjectMapper().extractPrimaryKey(obj);
    int oldOrdinal = getOrdinalRecord(newState.getPriorState().getStateEngine(), pk);

    int newOrdinal = newState.add(obj);
    if (oldOrdinal != -1 && newOrdinal != oldOrdinal) {
        delete(pk);
    }
}
```

Just run deleting at the final step.
Meaning I want to avoid store objects in mutations map because it cost a lot of memory. But when I did that, the data is messed up like fields become a primary key. Could you help me explain this or do you have a solution to avoid store so many objects in mutation map?

Thanks

ghost:
Hi @maiakhoa. Our thinking for the `HollowIncrementalProducer` was to streamline the use-case where you would otherwise follow the guidance in [Delta-Based Producer Input](http://hollow.how/advanced-topics/#delta-based-producer-input) by accumulating changes in RAM. To avoid too much accumulation, you would either run cycles frequently enough (e.g. every N minutes on a set cadence) or keep track of modifications and run cycles once a threshold was reached (e.g. run a cycle every 1000 mutations).

We were also guarding against lost mutations. A naive implementation we discarded early on would have essentially done this:

1. finish a cycle, then immediately call `newState.getStateEngine().addAllObjectsFromPreviousCycle()`
2. call `addOrModify` or `delete` for events as they came in
3. call `runCycle` every so often
4. go to 1

One problem is cycle failure. It's possible for the cycle to fail, in which case we've lost all of the mutation events that came in. Unless there's a mechanism to reliably replay those events we've either lost data or we have to a traditional `HollowProducer` *add the world* cycle from the source-of-truth.

A second problem has to do with multiple events for the same primary key. These have to be deduped.

Both problems are sidestepped by using a `Map<PrimaryKey,POJO>` though it does limit utility to those cases which can fit all mutations for a single cycle on the heap.

ghost:
This doesn't work as expected:

```
    int oldOrdinal = getOrdinalRecord(newState.getPriorState().getStateEngine(), pk);

    int newOrdinal = newState.add(obj);
[...]

PR #144, opened 2018-01-09
Hollow incremental populator cleanup

rpalcolea:
Did some code re-arrangement (introduced `HollowIncrementalCyclePopulator` and now `HollowIncrementalProducer` is simpler. Also found that clearChanges was executed before producing a new state https://github.com/Netflix/hollow/blob/master/hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java#L435, so a failure there could end-up losing all the mutations for the next cycle. I read your comments on https://github.com/Netflix/hollow/issues/137#issuecomment-356174169 and found that out.

Now `clearChanges` gets executed after we get a version from `runCycle` in `HollowIncrementalProducer` https://github.com/rpalcolea/hollow/blob/c22a096b408a5de234c72bfdfa219acd23babeee/hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java#L73

The tests didn’t change at all, just moving code here and there.

I was thinking on adding a cleanup method to the `Populator` interface and invoke it at the end of `runCycle` -> https://github.com/Netflix/hollow/blob/master/hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java#L482 but didn’t know if that was something you’d want. Wondering if people would like to cleanup populator’s states… for now I think it makes more sense to just clean the mutations map as before.

Let me know your thoughts

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalCyclePopulator.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalCyclePopulator.java
@@ -1,0 +1,114 @@
+package com.netflix.hollow.api.producer;
+
+import com.netflix.hollow.core.index.HollowPrimaryKeyIndex;
+import com.netflix.hollow.core.read.engine.HollowReadStateEngine;
+import com.netflix.hollow.core.read.engine.HollowTypeReadState;
+import com.netflix.hollow.core.schema.HollowObjectSchema;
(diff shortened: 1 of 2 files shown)


rpalcolea:
@toolbear, do you have thoughts/feedback on this?

ghost:
It was good the first time I looked at it, just forgot to merge.

Have some other changes in the works that I'd like to get in the next release, so holding off on cutting a release right now.

Issue #145, opened 2018-01-10
Pinning Consumers - integration


rpalcolea:
Hi @toolbear,

I was wondering if there are plans to introduce the ["Pinning Consumers"](http://hollow.how/infrastructure/#pinning-consumers) mechanism as a default in Hollow.

I think it would be nice if we could know when a consumer is pinned and expose this via the metrics or the API. something like `consumer.isPinned()` and/or `consumer.getPinnedVersion()`. This could help to triage issues when you have tens/hundreds of consumers with the same dataset but because of the nature of eventual consistency, a few of them could be stuck. 

I don't have a clear idea on how this could be accomplish, my guess is that the `AnnouncementWatcher` could be modified to support the pinning mechanism and `HollowConsumer` expose the information, similar to:

```java
public long getCurrentVersionId() {
        return updater.getCurrentVersionId();
}
```

Something like

```java
public long getPinnedVersion() {
        return announcementWatcher.getPinnedVersion();
} 
```

It would be up to the users to store the pinned version in the `announcementWatcher` or read it from their database/blob storage every time they need this. `getPinnedVersion` should be another abstract method in `AnnouncementWatcher`.

From the `HollowProducer` perspective, I think Hollow could provide a `versionPinner` in the builder that takes an `VersionPinner` (new interface). e.g.

```java
HollowProducer producer = HollowProducer.withPublisher(new FakeBlobPublisher())
                                                .withAnnouncer(new MyAnnouncer(tmpFolder))
                                                .withVersionPinner(new MyVersionPinner());
```

Then, probably the `HollowProducer` could have a `pin`Version method that takes the version number as an argument and invokes ` versionPinner.pin(long pinnedVersion)`. This way users could programmatically pin version from the producer either by background jobs based on their own business rules or as simple as expose the `pinVersion` as a HTTP endpoint.

Also it could have the `unpin`. 

thoughts?

PR #146, opened 2018-01-10
Version pinning mechanism [proposal]

rpalcolea:
Introduces the pinning mechanism as part of `HollowProducer` and `HollowConsumer`

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
@@ -254,6 +254,18 @@ public class HollowConsumer {
     }
 
     /**
+     * @return the pinned version of the announcement watcher.
+     */
+    public long getPinnedVersion() {
+        return announcementWatcher.getPinnedVersion();
+    }
+
+
+    public boolean isPinned() {
+        return announcementWatcher.hasPinnedVersion();
+    }
+
+    /**
      * @return the api which wraps the underlying dataset.
      */
     public HollowAPI getAPI() {
@@ -405,14 +417,20 @@ public class HollowConsumer {
      */
     public static interface AnnouncementWatcher {
 
-        public static final long NO_ANNOUNCEMENT_AVAILABLE = Long.MIN_VALUE;
+        public static final long NO_VERSION_AVAILABLE = Long.MIN_VALUE;
 
         /**
-         * Return the latest announced version.
-         *
-         * @return
+         * Returns the latest announced version.
          */
         public long getLatestVersion();
+
+        /**
+         * Returns the latest announced version.
+         */
+        public long getPinnedVersion();
+
+
+        public boolean hasPinnedVersion();
 
         /**
          * Implementations of this method should subscribe a HollowConsumer to updates to announced versions.
--- a/hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
@@ -253,14 +253,6 @@ public class HollowConsumer {
         return updater.getCurrentVersionId();
     }
 
-    /**
-     * @return the pinned version of the announcement watcher.
-     */
-    public long getPinnedVersion() {
-        return announcementWatcher.getPinVersion();
-    }
-
-
     public boolean isPinned() {
         return announcementWatcher.hasPinVersion();
     }
@@ -423,11 +415,6 @@ public class HollowConsumer {
          * Returns the latest announced version.
          */
         public long getLatestVersion();
-
-        /**
-         * Returns the latest announced version.
-         */
-        public long getPinVersion();
 
 
         public boolean hasPinVersion();
--- a/hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
@@ -257,12 +257,12 @@ public class HollowConsumer {
      * @return the pinned version of the announcement watcher.
      */
     public long getPinnedVersion() {
-        return announcementWatcher.getPinnedVersion();
+        return announcementWatcher.getPinVersion();
     }
 
 
     public boolean isPinned() {
-        return announcementWatcher.hasPinnedVersion();
+        return announcementWatcher.hasPinVersion();
     }
 
     /**
@@ -427,10 +427,10 @@ public class HollowConsumer {
         /**
          * Returns the latest announced version.
          */
-        public long getPinnedVersion();
+        public long getPinVersion();
 
 
-        public boolean hasPinnedVersion();
+        public boolean hasPinVersion();
 
         /**
          * Implementations of this method should subscribe a HollowConsumer to updates to announced versions.
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
@@ -17,7 +17,7 @@
  */
 package com.netflix.hollow.api.producer;
 
-import static com.netflix.hollow.api.consumer.HollowConsumer.AnnouncementWatcher.NO_ANNOUNCEMENT_AVAILABLE;
+import static com.netflix.hollow.api.consumer.HollowConsumer.AnnouncementWatcher.NO_VERSION_AVAILABLE;
 import static java.lang.System.currentTimeMillis;
 
 import java.io.File;
@@ -43,7 +43,6 @@ import com.netflix.hollow.api.producer.HollowProducerListener.RestoreStatus;
 import com.netflix.hollow.api.producer.enforcer.BasicSingleProducerEnforcer;
 import com.netflix.hollow.api.producer.enforcer.SingleProducerEnforcer;
 import com.netflix.hollow.api.producer.fs.HollowFilesystemBlobStager;
-import com.netflix.hollow.api.producer.fs.HollowFilesystemVersionPinner;
 import com.netflix.hollow.api.producer.validation.AllValidationStatus;
 import com.netflix.hollow.api.producer.validation.AllValidationStatus.AllValidationStatusBuilder;
 import com.netflix.hollow.api.producer.validation.HollowValidationListener;
@@ -444,7 +443,7 @@ public class HollowProducer {
      * at an optimal state.
      * 
      * @param config specifies what criteria to use to determine whether a compaction is necessary
-     * @return the version identifier of the produced state, or AnnouncementWatcher.NO_ANNOUNCEMENT_AVAILABLE if compaction was unnecessary.
+     * @return the version identifier of the produced state, or AnnouncementWatcher.NO_VERSION_AVAILABLE if compaction was unnecessary.
      */
     public long runCompactionCycle(HollowCompactor.CompactionConfig config) {
         if(config != null && readStates.hasCurrent()) {
@@ -459,7 +458,7 @@ public class HollowProducer {
             }
         }
         
-        return NO_ANNOUNCEMENT_AVAILABLE;
+        return NO_VERSION_AVAILABLE;
     }
 
     protected void runCycle(Populator task, ProducerStatus.Builder cycleStatus, long toVersion) {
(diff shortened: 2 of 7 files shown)


rpalcolea:
Hi @toolbear,

This is an idea for https://github.com/Netflix/hollow/issues/145

Please let me know your thoughts and if this is desired

ghost:
The quick answer might be "yes, but not right this moment."

There are some features we have implemented internally (pinning, restore on startup, namespaces) that we believe might be worth surfacing in the public Hollow API, but we aren't 100% confident.

Specifically for pinning, I'm not certain what shape that API should take. The one you propose doesn't fit how we do it, but maybe it should. The way we do it might not make a good fit as a public API or it might. Let's use #145 or #146 as a long-lived discussion on how that takes shape.

The other reason for deferring for a bit is that Q1 or Q2 we'll be doing some active development on our pinning implementation and from that might surface a better understanding.

That said, I'll respond with some concrete thoughts to the PR as written. Don't treat them as requested changes, though, as even with those changes we likely aren't ready to accept this feature.

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
```diff
@@ -405,15 +409,16 @@ public long getToVersion() {
      */
     public static interface AnnouncementWatcher {
 
-        public static final long NO_ANNOUNCEMENT_AVAILABLE = Long.MIN_VALUE;
+        public static final long NO_VERSION_AVAILABLE = Long.MIN_VALUE;
```
ghost, line 412: I wonder if we can keep the *announcement* nomenclature?

For cloud environment, we use two topics: main topic and the pin topic. In both cases we "announce" a version to that topic. Though maybe it's more strained to talk in terms of "announcing a pinned version" versus talking about "announcing" and "pinning" as two separate verbs. Or maybe the verb "pinning" is shorthand for "announcing a pinned version" and that's equally clear.

Of course, an equally valid implementation is to retrieve the pinned version from a side channel unrelated to how your main announcements are received (we also do this in some of our use cases). For that design maybe it feels like *pinning* would be shoehorned into an *announcement* metaphor. Or maybe it's fine that the Hollow's metaphor for receiving a version always come from an *announcement* even if it's a different channel altogether?

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -366,6 +387,26 @@ public boolean enablePrimaryProducer(boolean doEnable) {
         return (singleProducerEnforcer.isPrimary() == doEnable);
     }
 
+    /**
+     * Allows to pin a specific version, won't produce data, only modifies the pinned version file
+     *
+     */
+    public void pinVersion(long pinnedVersion) {
```
ghost, line 394: Our tooling for performing a pin is operationally separate from our producer. There may be some good reasons for coupling the two, but for our use-case we'd really have a system that only had the `VersionPinner` and eschewed using a `HollowProducer` at all.

Said another way, if you wanted to build a microservice that was only responsible for pinning, it would be heavy weight to have to instantiate a producer just to do it. Similarly, there's value in not having to rely on the producer that's producing your data states to do the actual pinning; one case could be if the producer somehow announced a bad state just prior to becoming unresponsive. Operationally you would like to pinback quickly while taking your time on the stuck producer remediation.

That said, I can see the motivation of wanting to consolidate the idea of pinning in both the `HollowConsumer` and `HollowProducer` interfaces.

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
```diff
@@ -405,15 +409,16 @@ public long getToVersion() {
...
         /**
-         * Return the latest announced version.
-         *
-         * @return
+         * Returns the latest announced version.
          */
         public long getLatestVersion();
 
+
+        public boolean hasPinVersion();
```
ghost, line 420: I wonder if the code is telling us that we should be exposing an `Announcement` or `AnnouncementMetadata` class?

We would keep `getLatestVersion()` for the simple case where you only cared about the numeric version, but add a `getLatestAnnouncement()` that returned a richer object (or maybe just a Map). Elsewhere, at least for the *listener* and *metrics* type interfaces we'd include the richer object instead of just `long version`. Then whether or not the version is pinned would just be a facet of information about that announcement.

ghost: One example, we might include a key-value pair of the announced version that was overridden by the pinned version.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
rpalcolea: I thought about this too.

I agree that most of the times you'd like to have a `VersionPinner` to do the job without knowing `HollowProducer` at all.

I think it makes more sense to consolidate the pinning in the consumer rather than the producer. 

One thing I can think of is that if the `HollowProducer` knows about the `VersionPinner` and that a `pin` happened, you could potentially prevent the producer for executing more cycles while you fix any potential issue.  I know this can be done via tools to pin version and stop the producer (if needed)... so I was wondering if this could be a potential use case for adding the pinner to the producer. 

I think at the end, having a `VersionPinner` could be helpful for both cases... probably it shouldn't be part of the `HollowProducer` package at all, but the producer could know about it. Also, I didn't get into the details of listeners for the pinning mechanism on this initial proposal

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
rpalcolea: I literally spent a long time going on circles on this. I think you could announce two type of versions, the "pinned" one and the "normal"? (don't know how to call it) and `NO_ANNOUNCEMENT_AVAILABLE` could prevail. 

To be honest, while doing these changes, I thought `NO_ANNOUNCEMENT_AVAILABLE` wasn't the proper one because I'm not used to see the pinning as part of the Hollow internals, so you always think of announcement as a new version that came in. I think is more of semantics of announcement on a daily-basis usage rather than if it fits or not... hope I'm being clear on this. Now it makes more sense to keep it as it is.

Also, I think `NO_ANNOUNCEMENT_AVAILABLE` could be part of some sort of constants class that could be shared across `VersionPinner`, `HollowProducer` and `HollowConsumer`; in particular for the Filesystem implementations.

 Overall, I think pinning needs more exposure, either at docs level or as part of the public API so people could adopt easily that pattern which IMO is very useful. In my experience, I talked to other teams that didn't think about it and didn't read the [Pinning Consumer](http://hollow.how/infrastructure/#pinning-consumers) section and this could help them while triaging issues

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
rpalcolea: I think this is a great idea, that way you could know the pinned version and the last version that was applied before the pinning. I don't know if it would be useful or not, but having a metric on how much time you spent on pinning could be another point to introduce a `AnnouncementMetadata` class.

rpalcolea:
Hi @toolbear, 

Yes, I realized that the proposal could be very different from what you do for pinning since I don't know the internals on your side and to be honest, we haven't spent that much of time on it to come up with a robust implementation of pinning mechanism.

With that said, I appreciate your time and comments and made me think about things like `AnnouncementMetadata` or `Announcement` objects and so on :)

I'll keep this open just in case you want to keep a discussion in the future (months from now I guess) .

On a side note, while I understand the workflow (or not?, you implement features that are needed at Netflix internally and then open source them), is there a desire to have some sort of list of "good contributions" that are probably low priority for Netflix but desired for Hollow in a short-mid term that external people (like me) could help with? I think in our case, we are doing a heavy investment in Hollow and it's getting adopted by several teams and while we spend time on building tools and APIs that are domain specific to us, I'd like to contribute back as much as possible 👍, let me know your position on this :)

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
ghost: > so you always think of announcement as a new version that came in. I think is more of semantics of announcement on a daily-basis usage

That's a good insight. Even if you can learn that Hollow treats pinning like a flavor of announcing, is that a natural way of thinking about it or will you always find yourself going "oh yeah, pins are announcements, too". It's one thing if that happens to an implementation detail and another if it's the metaphor in the public API.

> Overall, I think pinning needs more exposure, either at docs level or as part of the public API so people could adopt easily that pattern which IMO is very useful.

👍

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
ghost: > One thing I can think of is that if the HollowProducer knows about the VersionPinner and that a pin happened, you could potentially prevent the producer for executing more cycles while you fix any potential issue. I know this can be done via tools to pin version and stop the producer (if needed)... so I was wondering if this could be a potential use case for adding the pinner to the producer.

That's also a good point. You might not always want to stop producer cycles when pinning. In our experience it depends on whether you're responding to a bug in the producer versus responding to bad data from your source-of-truth. In the case of bad source data you may want to keep running cycles because of the tooling (history, explorer) you get on published blobs.

Nevertheless, I can see that there might be potential value in having pinning visibility in the producer. As mentioned in the pinning section of the docs (and if you don't halt the producer when pinned) that you produce a bunch of bad delta states and you want the producer to patch over them by restoring from the pinned version. That could be justify tighter coupling between producer and pinner.

ghost:
@rpalcolea 

> Yes, I realized that the proposal could be very different from what you do for pinning since I don't know the internals on your side and to be honest, we haven't spent that much of time on it to come up with a robust implementation of pinning mechanism.

This PR/conversation is really helpful because a lot of what we're looking for is how other people are doing/thinking-about pinning in order to inform how we should shape the API (or whether we should). We'd like to avoid adding an abstraction that turns out not to be useful for others (even if it, say, matches how we do it internally).

> On a side note, while I understand the workflow (or not?, you implement features that are needed at Netflix internally and then open source them), is there a desire to have some sort of list of "good contributions" that are probably low priority for Netflix but desired for Hollow in a short-mid term that external people (like me) could help with?

That's a good suggestion for a couple of reasons. One, we do have changes like that which would be a good fit for community contributions. Two, I've been thinking a lot about how we want to manage Hollow 3.0 which we expect to have a fair number of breaking changes (a lot in terms of different defaults, some in terms of removing APIs).

I'll take this as feedback that we can do a better job being more visible with our plans. Thanks!

> I think in our case, we are doing a heavy investment in Hollow and it's getting adopted by several teams and while we spend time on building tools and APIs that are domain specific to us, I'd like to contribute back as much as possible 👍, let me know your position on this :)

Contributions are definitely welcome, even those which might have impedance mismatch with how we use Hollow at Netflix. I'll work to make us better about being clear and prompt when a particular change isn't desired so that an issue/PR that we ultimately won't merge isn't left in limbo. Contributions that require us to change how we use hollow aren't expressly discouraged, but at a minimum the pace with which they're made will need to align with our schedules and so might move slower than if it was just a pure contribution from the community. I hope that makes sense. BTW, we don't run a fork of Hollow, we build internally against the same releases you use. This has both pros and cons for the pace of change.

ghost:
> or not?, you implement features that are needed at Netflix internally and then open source them

I realized that I missed addressing an implied question there.

We do both things:

1. implement functionality internally that eventually makes its way into Hollow
2. implemented desired changes directly in Hollow to meet needs we have internally

In fact, because we don't fork Hollow and want to avoid drifting or forming a sedimentary layer over the top of Hollow, we have a strong bias towards (2) now.

A good example of (1) is our code for automatically restoring a producer from the last version announced. We recognize that some support for this in Hollow would be useful and would make our implementation cleaner, we just haven't invested in that yet.

A good example of (2) is hooks we added for enforcing that a single producer is publishing and announcing. We added that to Hollow at the same time as we were implementing the zookeeper-based system we use.

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
kineshsatiya: I like the idea. Announcement naturally means a new version is available, whether it's pinned version or a new version that producer published.

PR #147, opened 2018-01-14
Allow IncrementalProducer to discard changes 

rpalcolea:
Hi @toolbear,

Was playing around with the `HollowIncrementalProducer` and thought that it would be nice if `discard` was introduced. This way if a user adds/modify/delete something and later executes some business logic and determines that the change shouldn't be reflected, they could just `discard` that particular record. 

With this, users now could clear all by using `cleanChanges` or clear/discard a specific record with `discard`, either with an object that contains the primary key or a `RecordPrimaryKey`... same was as `delete`.

Let me know your thoughts.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
@@ -57,9 +57,18 @@ public class HollowIncrementalProducer {
         RecordPrimaryKey pk = producer.getObjectMapper().extractPrimaryKey(obj);
         delete(pk);
     }
+
+    public void discard(Object obj) {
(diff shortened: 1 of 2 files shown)


rpalcolea:
Hi @toolbear, I ended implementing the support for serial/parallel collections mentioned in the description... worst case scenario, undo that commit.

Ideally, the `callbacks` should be refactored to lambdas when Java 8 is introduced

ghost (changes requested):
I like where your thinking on collection processing, but I'd rather consider that as a separate PR. Can you rebase those commits away (and submit them as another PR) and leave the *discard* changes in this PR?

rpalcolea:
Hi @toolbear ,

Done and opened https://github.com/Netflix/hollow/pull/152 with the collection in serial/parallel support

ghost (approved):

PR #150, opened 2018-01-18
fix isListenToDataRefreah typo

rpalcolea:


hollow/src/main/java/com/netflix/hollow/api/consumer/index/AbstractHollowHashIndex.java
```diff
@@ -63,20 +63,20 @@ private API castAPI(HollowAPI api) {
         return (API) api;
     }
 
-    public boolean isListenToDataRefreah() {
-        return isListenToDataRefreah;
+    public boolean isListenToDataRefresh() {
```
ghost: As annoying as it seems, how about keeping a deprecated `isListenToDataRefreah` around that delegates to the correct method to avoid this being a breaking change for anyone who coded to the version with the typo?

ghost (changes requested):

ghost (approved):

PR #151, opened 2018-01-18
enable isListenToDataRefresh in HollowUniqueKeyIndexGenerator by default

rpalcolea:


hollow/src/main/java/com/netflix/hollow/api/codegen/indexes/HollowUniqueKeyIndexGenerator.java
```diff
@@ -81,7 +81,7 @@ protected void genConstructors(StringBuilder builder) {
 
     protected void genSimpleConstructor(StringBuilder builder) {
         builder.append("    public " + className + "(HollowConsumer consumer) {\n");
-        builder.append("        this(consumer, false);");
+        builder.append("        this(consumer, true);");
```
ghost: Close ;)

Should neither be hardcoded to `false` nor `true`, but concatenate `isAutoListenToDataRefresh`.

ghost (dismissed):

PR #152, opened 2018-01-18
support for add/modify/delete collections in serial and parallel way

rpalcolea:
regarding `addOrModify` and `delete` one thing we do on our side is use Java 8 `parallelStream` to process a list of thousands of records. While we could argue that Hollow shouldn't be in charge of multi-threading add/modify/delete, I think it could be useful to allow the `HollowIncrementalProducer` to accept collections of objects and add/modify/delete them from the `mutations` in a parallel way. Still there is the assumption that the `List` should contain unique keys, otherwise the user could have race condition issues. This could be something for Hollow 3.x and Java 8 or use `SimultaneousExecutor` to keep threading consistency across hollow (?).  Also, what are your thoughts on adding a simple iterator for add/modify/remove, something in a serial fashion... let the user decide if they want to process serial or parallel (have both):

`addOrModify(List<Object> objs)`  and `addOrModifyInParallel(List<Object> objs)` 
`delete(List<Object> objs)`  and `deleteInParallel(List<Object> objs)` 

For now, the `addOrModifyInParallel` and `deleteInParallel` methods could re-use the `threadsPerCpu` provided for the `populator` and create a `SimultaneousExecutor` with it.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
@@ -20,6 +20,7 @@ package com.netflix.hollow.api.producer;
 import com.netflix.hollow.api.consumer.HollowConsumer;
 import com.netflix.hollow.api.consumer.HollowConsumer.BlobRetriever;
 import com.netflix.hollow.api.consumer.fs.HollowFilesystemAnnouncementWatcher;
+import com.netflix.hollow.core.util.SimultaneousExecutor;
 import com.netflix.hollow.core.write.objectmapper.RecordPrimaryKey;
 
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


kineshsatiya:
Just curious if addOrModify method could be used with multiple threads, then why add addOrModifyInParallel?

rpalcolea:
Hi @kineshsatiya,

While I agree that `addOrModify` could/should be used with multiple threads, I didn't want to assume that everyone wants this behavior. For example, if a user decides to do "micro" cycles with a few objects (less than 100 let's say), probably scheduling the threads would be costly in comparison of processing that list of tens of elements in a serial way.

I'm totally up for making it parallel by default but wanted to bring that up.


thoughts?

ghost:
Maybe when we remove support for Java 7 and can use Java 8 streams we can revisit these APIs and find a way to generalize the serial/parallel under a common API. For now I'm okay with separate methods, though I do feel similar to @kineshsatiya that it adds weight to the public API.

ghost (changes requested):
Okay with the changes as-is. Resolve conflicts and can merge.

PR #153, opened 2018-01-19
IncrementalProducer listeners (IncrementalCycleListener) and cycleMetadata

rpalcolea:
Introduces `IncrementalCycleListener` and also the builder pattern for `HollowIncrementalProducer`


kineshsatiya:
I wonder if we should have an interface for the producer with two implementations - CycleProducer & IncrementalCycleProducer. Some abstraction would be nice. I think the same thing would apply to listeners. I see that we have IncrementalListener in ListenerSupport, and they are only used in IncrementalHollowProducer. We should probably use separate them out. What your thoughts on this @toolbear and @rpalcolea ?

rpalcolea:
Hi @kineshsatiya ,

I think this aligns with https://github.com/Netflix/hollow/issues/154. What are your thoughts on that? I wanted to re-use `ListenerSupport` because is part of the `producer` package... and yes, seems like the `IncrementalListener` is only used by the incremental producer. I think consolidating things into one interface with two implementations would be nice... don't know if that's something in a short term though.

kineshsatiya:
Hi @rpalcolea , I'll review the current PR. If we can come up with a better design now, we can start with that. However, we can re-visit the design, when we start on next major release of Hollow.

rpalcolea:
sounds good @kineshsatiya

rpalcolea:
Hi @toolbear,

Just added something to this PR that could be helpful for others: `cycleMetadata`

The idea is that users could attach metadata to each cycle:

```
 HashMap<String, Object> cycleMetadata = new HashMap<>();
        cycleMetadata.put("foo", "bar");
        incrementalProducer.addCycleMetadata(cycleMetadata);
```

then in the listener we will have:

```void onCycleComplete(IncrementalCycleStatus status, long elapsed, TimeUnit unit, HashMap<String, Object> cycleMetadata)```

Then, users could use that metadata to take action at listener level. 

In our case we would like to attach some metadata from our source system and trigger actions depending on success or failure.

File: hollow/src/main/java/com/netflix/hollow/api/producer/AbstractIncrementalCycleListener.java [hunk 1]
```diff
@@ -0,0 +1,32 @@
...
+
+import java.util.HashMap;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Beta API subject to change.
+ */
+public class AbstractIncrementalCycleListener implements IncrementalCycleListener {
+    @Override
+    public void onCycleComplete(IncrementalCycleStatus status, long elapsed, TimeUnit unit, HashMap<String, Object> cycleMetadata) { }
```
ghost: Expose `Map<String,Object>` instead of `HashMap` in the public API

ghost: Instead of adding new argument to the listener interface, we've tended to make additional info available through the already passed in cycle status class (in this case `IncrementalCycleStatus`). This has the added bonus of avoiding a breaking change (not that that is necessary given this is an experimental API).

hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
```diff
@@ -28,18 +33,33 @@
...
 public class HollowIncrementalProducer {
     
+    private static final long FAILED_VERSION = Long.MIN_VALUE;
+
     private final HollowProducer producer;
     private final ConcurrentHashMap<RecordPrimaryKey, Object> mutations;
     private final HollowProducer.Populator populator;
-    
+    private final ListenerSupport listeners;
+    private final HashMap<String, Object> cycleMetadata;
```
ghost: This can be declared as `Map`.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
```diff
@@ -28,18 +33,33 @@
...
         this.mutations = new ConcurrentHashMap<RecordPrimaryKey, Object>();
         this.populator = new HollowIncrementalCyclePopulator(mutations, threadsPerCpu);
+        this.listeners = new ListenerSupport();
+        this.cycleMetadata = new HashMap<String, Object>();
+
+        for(IncrementalCycleListener listener : listeners)
+            this.listeners.add(listener);
+    }
+
+    public void addCycleMetadata(HashMap<String, Object> metadata) {
```
ghost: Should this be `addAllCycleMetadata` with the addition of a `addCycleMetadata(String key, Object value)` method?

hollow/src/main/java/com/netflix/hollow/api/producer/IncrementalCycleListener.java
```diff
@@ -0,0 +1,157 @@
...
+    /**
+     * Called after {@code HollowIncrementalProducer} has completed a cycle normally or abnormally. A {@code SUCCESS} status indicates that the
+     * entire cycle was successful and the producer is available to begin another cycle.
+     *
+     * @param status IncrementalCycleStatus of this cycle. {@link IncrementalCycleListener.IncrementalCycleStatus#getStatus()} will return {@code SUCCESS}
+     *   when the a new data state has been announced to consumers or when there were no changes to the data.
+     * @param elapsed duration of the cycle in {@code unit} units
+     * @param unit units of the {@code elapsed} duration
+     */
+    public void onCycleComplete(IncrementalCycleStatus status, long elapsed, TimeUnit unit, HashMap<String, Object> cycleMetadata);
```
ghost: `Map` instead of `HashMap`

hollow/src/main/java/com/netflix/hollow/api/producer/IncrementalCycleListener.java
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * Called after {@code HollowIncrementalProducer} has completed a cycle normally or abnormally. A {@code FAIL} status indicates that the
+     * entire cycle failed.
+     *
+     * @param status IncrementalCycleStatus of this cycle. {@link IncrementalCycleListener.IncrementalCycleStatus#getStatus()} will return {@code FAIL}
+     * @param elapsed duration of the cycle in {@code unit} units
+     * @param unit units of the {@code elapsed} duration
+     */
+    public void onCycleFail(IncrementalCycleStatus status, long elapsed, TimeUnit unit, HashMap<String, Object> cycleMetadata);
```
ghost: `Map` instead of `HashMap`

hollow/src/main/java/com/netflix/hollow/api/producer/ListenerSupport.java
```diff
@@ -169,4 +182,16 @@ void fireAnnouncementComplete(ProducerStatus.Builder psb) {
         ProducerStatus st = psb.build();
         for(final HollowProducerListener l : listeners) l.onAnnouncementComplete(st, psb.elapsed(), MILLISECONDS);
     }
+
+    void fireIncrementalCycleComplete(long version, long recordsAddedOrModified, long recordsRemoved, HashMap<String, Object> cycleMetadata) {
```
ghost: `Map` instead of `HashMap`

hollow/src/main/java/com/netflix/hollow/api/producer/ListenerSupport.java
```diff
@@ -169,4 +182,16 @@ void fireAnnouncementComplete(ProducerStatus.Builder psb) {
...
+        IncrementalCycleStatus.Builder icsb = new IncrementalCycleStatus.Builder().success(version, recordsAddedOrModified, recordsRemoved);
+        for(final IncrementalCycleListener l : incrementalCycleListeners)
+            l.onCycleComplete(icsb.build(), icsb.elapsed(), MILLISECONDS, cycleMetadata);
+    }
+
+    void fireIncrementalCycleFail(Throwable cause, long recordsAddedOrModified, long recordsRemoved, HashMap<String, Object> cycleMetadata) {
```
ghost: `Map` instead of `HashMap`

hollow/src/main/java/com/netflix/hollow/api/producer/ListenerSupport.java
```diff
@@ -169,4 +182,16 @@ void fireAnnouncementComplete(ProducerStatus.Builder psb) {
         ProducerStatus st = psb.build();
         for(final HollowProducerListener l : listeners) l.onAnnouncementComplete(st, psb.elapsed(), MILLISECONDS);
     }
+
+    void fireIncrementalCycleComplete(long version, long recordsAddedOrModified, long recordsRemoved, HashMap<String, Object> cycleMetadata) {
+        IncrementalCycleStatus.Builder icsb = new IncrementalCycleStatus.Builder().success(version, recordsAddedOrModified, recordsRemoved);
```
ghost: This is where an `icsb.withMetadata(cycleMetadata)` would go instead of passing it directly to listener callback.

ghost (changes requested):
Looks good. Some small changes and one medium change around how to expose the metadata to listeners.

rpalcolea:
Hi @toolbear ,

Thanks for the feedback. 

Made the changes to use `Map` and also to include `cycleMetadata` as part of `IncrementalCycleStatus`

rpalcolea:
Also, I'd like to get this one (bigger) in first in order to fix conflicts on https://github.com/Netflix/hollow/pull/152 and https://github.com/Netflix/hollow/pull/182 if that's ok with you @toolbear

hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
```diff
@@ -99,70 +113,133 @@ public void clearChanges() {
...
-    public boolean hasChanges() { return this.mutations.size() > 0; }
+    public boolean hasChanges() {
+        return this.mutations.size() > 0;
+    }
+
+    public void addCycleMetadata(String key, Object value) {
+        this.cycleMetadata.put(key, value);
+    }
+
+    public void addAllCycleMetadata(HashMap<String, Object> metadata) {
```
ghost: Missed one 😉 

Use `Map` here, too.

ghost (changes requested):

PR #160, opened 2018-01-24
Incremental producer - delete orphan tests

rpalcolea:
Added unit test to verify orphan objects get removed from Hollow using the incremental producer.

The idea is to get this fixed -> https://github.com/Netflix/hollow/issues/162


rpalcolea:
Wrote some tests for https://github.com/Netflix/hollow/issues/162

I think it would be nice to keep tests to guarantee the orphan deletion in the incremental producer

PR #163, opened 2018-01-25
add test for resuming work when announcement fails

rpalcolea:
Found an issue in our side while trying to resume work on incremntal producer after an announcement failure.

I think it would be nice to have a test to cover this

thoughts?


ghost:
@akhaku recently added Mockito dependency. This PR has conflicts, but looks good to go. The conflict is likely on the mockito addition. I can merge once conflicts are resolved.

rpalcolea:
the conflicts should be resolved now @toolbear , thanks for looking at it

PR #169, opened 2018-02-08
Allow passing a custom Writer to HollowRecordStringifier

akhaku:
For cases where a record has a large amount of data,
stringifying it can result in the underlying StringBuilder
running out of memory. Expose a version of the stringify
methods that take Writer objects instead and write records
using the provided Writer.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/tools/stringifier/HollowRecordJsonStringifier.java
+++ b/hollow/src/main/java/com/netflix/hollow/tools/stringifier/HollowRecordJsonStringifier.java
@@ -78,75 +78,59 @@ public class HollowRecordJsonStringifier implements HollowStringifier<HollowReco
         return this;
     }
 
-
-    /**
-     * Create a JSON representation of the provided {@link HollowRecord}.
-     */
     @Override
     public String stringify(HollowRecord record) {
         return stringify(record.getTypeDataAccess().getDataAccess(), record.getSchema().getName(), record.getOrdinal());
     }
 
-
-    /**
-     * Create a JSON representation of the record in the provided dataset, of the given type, with the specified ordinal.
-     */
     @Override
-    public String stringify(HollowDataAccess dataAccess, String type, int ordinal) {
-        StringWriter builder = new StringWriter();
-
-        try {
-            appendStringify(builder, dataAccess, type, ordinal, 0);
-        } catch(IOException e) {
-            throw new RuntimeException(e);
-        }
-
-        return builder.toString();
-    }
-
-    /**
-     * Write a JSON representation of the provided {@link HollowRecord} to the provided Writer.
-     */
     public void stringify(Writer writer, HollowRecord record) throws IOException {
         stringify(writer, record.getTypeDataAccess().getDataAccess(), record.getSchema().getName(), record.getOrdinal());
     }
 
-    /**
-     * Write a JSON representation of the record in the provided dataset, of the given type, with the specified ordinal.
-     *
-     * The representation will be written to the provided Writer.
-     */
+    @Override
+    public String stringify(HollowDataAccess dataAccess, String type, int ordinal) {
+        try {
+            StringWriter writer = new StringWriter();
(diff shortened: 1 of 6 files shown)


hollow/src/main/java/com/netflix/hollow/tools/stringifier/HollowRecordJsonStringifier.java
```diff
@@ -78,83 +78,67 @@ public HollowRecordJsonStringifier addExcludeObjectTypes(String... types) {
         return this;
     }
 
-
-    /**
-     * Create a JSON representation of the provided {@link HollowRecord}.
-     */
```
akhaku, line 84: Removed duplicate javadocs - the interface method that we override has the javadocs, no sense in duplicating them

hollow/src/main/java/com/netflix/hollow/tools/stringifier/HollowRecordJsonStringifier.java
```diff
@@ -78,83 +78,67 @@ public HollowRecordJsonStringifier addExcludeObjectTypes(String... types) {
...
-     *
-     * The representation will be written to the provided Writer.
-     */
+    @Override
     public void stringify(Writer writer, HollowDataAccess dataAccess, String type, int ordinal) throws IOException {
         appendStringify(writer, dataAccess, type, ordinal, 0);
     }
 
-    private void appendStringify(Writer builder, HollowDataAccess dataAccess, String type, int ordinal, int indentation) throws IOException {
+    private void appendStringify(Writer writer, HollowDataAccess dataAccess, String type, int ordinal, int indentation) throws IOException {
```
akhaku, line 107: the only diff here is renaming building to writer

hollow/src/main/java/com/netflix/hollow/tools/stringifier/HollowRecordStringifier.java
```diff
@@ -67,60 +69,70 @@ public HollowRecordStringifier addExcludeObjectTypes(String... types) {
...
 
-        appendStringify(builder, dataAccess, type, ordinal, 0);
+    @Override
+    public String stringify(HollowDataAccess dataAccess, String type, int ordinal) {
+        try {
+            StringWriter writer = new StringWriter();
+            stringify(writer, dataAccess, type, ordinal);
+            return writer.toString();
+        } catch (IOException e) {
+            throw new IllegalStateException("Unexpected exception using StringWriter", e);
```
ghost: I tend to reserve `IllegalStateException` for when a class has gotten into a bad internal state, e.g. where assumptions or invariants are violated. For demoting a checked exception I often use `RuntimeException` unless a better subclass presents itself.

ghost (commented):
Looks good. Small quibble below.

Do you want to also switch the Explorer UI over to the `Writer` flavor, or do that in a separate PR?

hollow/src/main/java/com/netflix/hollow/tools/stringifier/HollowRecordStringifier.java
akhaku: Fair enough, updated

akhaku:
Thanks for the review! Don't know anything about VelocityContext at the moment, but it looks like updating the Explorer UI is probably not a trivial change, so would prefer to tackle that in a follow-up PR.

hollow/src/main/java/com/netflix/hollow/tools/stringifier/HollowStringifier.java
```diff
@@ -18,24 +18,42 @@
...
 
     /**
-     * Exclude specified object types (replace output with null)
+     * Exclude specified object types (replace output with null).
      */
     public T addExcludeObjectTypes(String... types);
 
     /**
-     * create a String representation of the specified {@link HollowRecord}.
+     * Create a String representation of the specified {@link HollowRecord}.
```
kfahy, line 36: marsad1

PR #172, opened 2018-02-23
Add com.netflix.hollow.core.type support for Byte

akhaku:
Rather than generating Byte classes when we encouunter a
Byte as part of another Type, introduce a top-level Byte
type as part of the core hollow types. Bytes are stored
as integers under the hood, and the generated classes will
use integers, similar to how a primitive byte gets converted
to an int.
Adding support for byte to the hollow schema is out of scope
for this PR.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/type/ByteTypeAPI.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/type/ByteTypeAPI.java
@@ -1,0 +1,60 @@
+/*
+ *
+ *  Copyright 2018 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.hollow.core.type;
+
+import com.netflix.hollow.api.custom.HollowAPI;
+import com.netflix.hollow.api.custom.HollowObjectTypeAPI;
+import com.netflix.hollow.core.read.dataaccess.HollowObjectTypeDataAccess;
+import com.netflix.hollow.core.type.delegate.ByteDelegateLookupImpl;
+
+public class ByteTypeAPI extends HollowObjectTypeAPI {
+
+    private final ByteDelegateLookupImpl delegateLookupImpl;
+
+    public ByteTypeAPI(HollowAPI api, HollowObjectTypeDataAccess typeDataAccess) {
+        super(api, typeDataAccess, new String[] {
+            "value"
+        });
+        this.delegateLookupImpl = new ByteDelegateLookupImpl(this);
+    }
+
+    public int getValue(int ordinal) {
+        if(fieldIndex[0] == -1)
+            return missingDataHandler().handleByte("Byte", ordinal, "value");
+        // bytes are actually stored as ints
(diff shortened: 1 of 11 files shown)


hollow/src/main/java/com/netflix/hollow/core/type/ByteTypeAPI.java
```diff
@@ -0,0 +1,60 @@
...
+    private final ByteDelegateLookupImpl delegateLookupImpl;
+
+    public ByteTypeAPI(HollowAPI api, HollowObjectTypeDataAccess typeDataAccess) {
+        super(api, typeDataAccess, new String[] {
+            "value"
+        });
+        this.delegateLookupImpl = new ByteDelegateLookupImpl(this);
+    }
+
+    public int getValue(int ordinal) {
```
akhaku, line 36: One possibility is to have these methods return byte and Byte, to make it easier if we ever add byte support to the schema (probably a breaking change, so maybe with Hollow 3). Thoughts?

akhaku:
Feedback on the approach would be great, as well as on good tests to add. I looked through PrimitiveDataAccessorTest and adding a ByteDataAccessorTest, but that seems to involve adding bytes to the type schema, as far as I can tell.

hollow/src/main/java/com/netflix/hollow/core/type/ByteTypeAPI.java
davesu: To be consistent it seems best to return byte, regardless how internally it is stored.

hollow/src/main/java/com/netflix/hollow/core/type/ByteTypeAPI.java
```diff
@@ -0,0 +1,60 @@
...
+        if(fieldIndex[0] == -1)
+            return missingDataHandler().handleByte("Byte", ordinal, "value");
+        // bytes are actually stored as ints
+        return getTypeDataAccess().readInt(ordinal, fieldIndex[0]);
+    }
+
+    public Integer getValueBoxed(int ordinal) {
```
davesu, line 43: To be consistent it seems best to return Byte, regardless how internally it is stored.

davesu:
I don't believe it is trivial to properly introduce new hollow primitive types and will require much more changes to properly treat byte as a hollow primitive type such as changes to:

- HollowObjectTypeDataAccess
- HollowObjectSchema.FieldType
- Hollow Code Generator(s)

MORE IMPORTANTLY, it will likely break lots of external code that walks through the hollow schema such as :

         switch(schema.getFieldType(i)) {
            case BOOLEAN:
                return "false";
            case DOUBLE:
                return "java.lang.Double.NaN";
            case FLOAT:
                return "java.lang.Float.NaN";
            case INT:
                return "java.lang.Integer.MIN_VALUE";
            case LONG:
                return "java.lang.Long.MIN_VALUE";
            case BYTES:
            case STRING:
            case REFERENCE:
                return "null";
        }

Take a look at code references to HollowObjectSchema.FieldType.BYTES and that should give you an idea additional those places that needs to change to support FieldType.BYTE.  And likely more.

NOTE:  As currently proposed in this PR to store byte as INT in hollow blob without thorough support throughout hollow will likely cause issues.

davesu:
Actually, I forgot to mention that hollow primitive type means data can be inlined (as well as being referenced) and having both int and byte being stored as "int" in inlined mode, I am not sure how hollow will be able to differentiate between them.

akhaku:
> I am not sure how hollow will be able to differentiate between them

Yeah, that's a problem we have today. If you have a `byte` type in your model, it comes out the generated side as an `int`, since we don't have a `byte` type in the hollow schema. However, if you have a `Byte`, we generate Byte{Delegate*,DataAccessor,HollowFactory,TypeAPI} (what we would normally do for any user-defined objects). Generally this is fine, but Byte in those files conflict with java.lang.Byte, because of a star import.
I'll put up a PR to handle this another way, please hold.
EDIT #177 now up as an alternative solution. Also created #176 to describe the problem in more detail.

akhaku:
Declining in favor of #177 to fix #176

PR #173, opened 2018-02-23
Fix byte array ordinal map load factor

kineshsatiya:
Correction in calculating the load factor for the hash table in ByteArrayOrdinalMap. This will increase the capacity of the hash table before running into NegativeSizeArrayException. The old capacity was around 187M, after this change, the new capacity will be around 325M.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
@@ -433,7 +433,7 @@ public class ByteArrayOrdinalMap {
        populateNewHashArray(newKeys, valuesToAdd);
 
        /// 70% load factor
-       sizeBeforeGrow = (newKeys.length() * 7) / 10;
+       sizeBeforeGrow = (int)(((float)newKeys.length()) * 0.7);
        pointersAndOrdinals = newKeys;
    }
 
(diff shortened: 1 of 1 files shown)


hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
```diff
@@ -433,7 +433,7 @@ private void growKeyArray() {
        populateNewHashArray(newKeys, valuesToAdd);
 
        /// 70% load factor
-       sizeBeforeGrow = (newKeys.length() * 7) / 10;
+       sizeBeforeGrow = (int)(((float)newKeys.length()) * 0.7);
```
akhaku: Great catch!
Probably a bad idea to make the array size a `long` too, yeah?

akhaku:
Thanks for the formatting fixes, and for having those in a separate commit! Made reviewing a lot easier.

akhaku (approved):

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
```diff
@@ -433,7 +433,7 @@ private void growKeyArray() {
        populateNewHashArray(newKeys, valuesToAdd);
 
        /// 70% load factor
-       sizeBeforeGrow = (newKeys.length() * 7) / 10;
+       sizeBeforeGrow = (int) (((float) newKeys.length()) * 0.7);
```
akhaku, line 442: Might also be a good idea to throw an IllegalStateException (or something) with a useful error message if we overflow

kineshsatiya: Yea, good idea. I've added the changes to throw an `IllegalStateException` when the new size overflows. This will help provide more clarity on the error.

PR #174, opened 2018-02-24
Show `[null]` for null nodes in diff/history UI

### LINKED ISSUE #166: Diff tool should differentiate between null values and empty collections
When a collections field changes from `null` to a reference to Set|List|Map that is empty, the Diff and History UIs detect this as a difference, but then present no difference in the UI. Surface in the UI that the field changed from null to an empty collection (or vice versa).

akhaku:
For cases where we have a node that isn't a leaf (i.e. it represents an object), call out the fact that it is null, instead of ignoring that in the history and diff UIs. This is important for seeing the difference between a node that is null and a node that has no children (e.g. an empty list set, or map).
Fixes #166

### DIFF
--- a/hollow-diff-ui/src/main/java/com/netflix/hollow/diffview/DiffViewOutputGenerator.java
+++ b/hollow-diff-ui/src/main/java/com/netflix/hollow/diffview/DiffViewOutputGenerator.java
@@ -17,11 +17,13 @@
  */
 package com.netflix.hollow.diffview;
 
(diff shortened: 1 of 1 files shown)


akhaku:
Example going from empty collection to null:

screen shot 2018-02-26 at 1 40 02 pm

kineshsatiya (approved):
Nice! This issue had been outstanding for a long time.

PR #175, opened 2018-02-28
Add v2 producer listener and use it to fire "cycle skipped" event

akhaku:
HollowProducerListener cleanup and documentation update
    
    - remove unnecessary scope modifiers - everything in an
      interface is public by default
    - add missing param javadoc
    - document that implementing classes should extend
      BaseHollowProducerListener

Add v2 producer listener and use it to fire "cycle skipped" event
    
    - instead of breaking library users' compilation, create a
      subinterface of HollowProducerListener to add a new
      listener method
    - create a no-op version of our new HollowProducerListenerV2
      that users of our library should utilize to allow adding
      producer listener hooks without us breaking their
      compilation
    - when calling producer listener hooks, call a new hook when
      a cycle is skipped
    - add mockito as a testCompile dependency, and add unit
      tests for our new functionality

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducerListenerV2.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducerListenerV2.java
@@ -1,0 +1,36 @@
+/*
+ *
+ *  Copyright 2018 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.hollow.api.producer;
+
+/**
+ * An extension of {@link HollowProducerListener} to allow adding new methods without
+ * breaking backwards compatability.
+ * Subclasses should extend {@link BaseHollowProducerListener} to avoid having to override
+ * every method in this interface.
+ * TODO(hollow3): Collapse this into HollowProducerListener.
+ */
+public interface HollowProducerListenerV2 extends HollowProducerListener {
+    enum CycleSkipReason {
+        NOT_PRIMARY_PRODUCER;
+    }
+
+    /**
+     * Called when a cycle is skipped.
+     */
+    void onCycleSkip(CycleSkipReason reason);
+}
--- a/hollow/src/test/java/com/netflix/hollow/api/producer/HollowProducerListenerTest.java
+++ b/hollow/src/test/java/com/netflix/hollow/api/producer/HollowProducerListenerTest.java
@@ -1,0 +1,61 @@
+package com.netflix.hollow.api.producer;
+
+import com.netflix.hollow.api.producer.enforcer.SingleProducerEnforcer;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.ArgumentMatchers;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Unit tests to verify that HollowProducerListener objects provided to HollowProducers
+ * are invoked at the right times.
+ */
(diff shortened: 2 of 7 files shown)


hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducerListenerV2.java
```diff
@@ -0,0 +1,36 @@
...
+package com.netflix.hollow.api.producer;
+
+/**
+ * An extension of {@link HollowProducerListener} to allow adding new methods without
+ * breaking backwards compatability.
+ * Subclasses should extend {@link NoOpHollowProducerListener} to avoid having to override
+ * every method in this interface.
+ * TODO(hollow3): Collapse this into HollowProducerListener.
+ */
+public interface HollowProducerListenerV2 extends HollowProducerListener {
```
ghost: I hate `V{N}` in class names, yet this is probably the best way to address this. Making a new name like `EnhancedHollowProducerListener` falls short when you need to enhance it again.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducerListenerV2.java
```diff
@@ -0,0 +1,36 @@
...
+    enum CycleSkipReason {
+        NOT_PRIMARY_PRODUCER;
```
ghost, line 29: I like it. We've talked about having the ability to keep a producer from running cycles without shutting it down (for the style where you run cycles on a cadence). If we added support for that eventually, that could be another skip reasons.

hollow/src/main/java/com/netflix/hollow/api/producer/NoOpHollowProducerListener.java
```diff
@@ -0,0 +1,106 @@
...
+ */
+package com.netflix.hollow.api.producer;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A trivial implementation of {@link HollowProducerListenerV2} which does nothing.
+ * Implementations of HollowProducerListenerV2 should subclass this class for convenience.
+ */
+public class NoOpHollowProducerListener implements HollowProducerListenerV2 {
```
ghost: Small quibble. I wonder if `BaseHollowProducerListener` or `AbstractHollowProducerListener` would be better names? The only reason is that when I subclass it and do something in response to one or more signals, I'm not really a NoOp listener anymore.

hollow/src/test/java/com/netflix/hollow/api/producer/HollowProducerListenerTest.java
```diff
@@ -0,0 +1,61 @@
...
+    public void setUp() {
+        MockitoAnnotations.initMocks(this);
+        this.producer = new HollowProducer.Builder()
+            .withListeners(listener, listenerV2)
+            .withSingleProducerEnforcer(singleProducerEnforcer).build();
+    }
+
+    @Test
+    public void testCycleSkip() {
+        Mockito.when(singleProducerEnforcer.isPrimary()).thenReturn(false);
```
ghost, line 37: 👍  Yay. Mockito!

ghost (approved):
Looks good. The suggested name change for the no-op base class is up to your judgement. Good as-is if you prefer to keep the name.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducerListenerV2.java
akhaku: Yeah agreed, not ideal, but works for our use case. In theory, in the next version of hollow this file will go away.

hollow/src/main/java/com/netflix/hollow/api/producer/NoOpHollowProducerListener.java
akhaku: Good idea - I like BaseHollowProducerListener, since I wasn't planning on making it an abstract class.

hollow/src/test/java/com/netflix/hollow/api/producer/HollowProducerListenerTest.java
akhaku: so_good

kineshsatiya (approved):
So it looks like, a cycle in Hollow could be either success, fail or skip. Should this be reflected in `ProducerStatus`?

akhaku:
...it bothers me that github doesn't let me respond to messages in this thread.
Kinesh - good point. I'm not completely sure what the right thing is here - other cases that involve a ProducerStatus all involve an onCycleStart, but in this case we don't even fire a start. For that reason, I'm inclined to leave out ProducerStatus here entirely. No strong feelings on this though, I could be convinced either way.

akhaku:
renamed to BaseHollowProducerListener, updated commit message

PR #177, opened 2018-03-01
Use explicit imports instead of star imports for generated types

### LINKED ISSUE #176: Compilation of generated files fails if data model object has `Byte`
If you try and generate an API for object that contains a Byte, the generated files to do not compile. Any generated files that to try reference the type (eg <APIName>HashIndex.java) conflict between my.generated.package.Byte and java.lang.Byte, since we import my.generated.*.

A couple ways to fix this:
* have similar treatment for Byte as we do for String, Integer, etc (include them in com.netflix.hollow.core.type and replace with HByte) #172
* import the generated Byte classes directly, instead of using a star import (WIP)

akhaku:
For cases where our generated type matches a java.lang type
(eg Byte), using a star import causes the compiler to get
confused between the generated type and the java.lang type.
Switching to explicit imports makes it clear to the
compiler, so do that.
Fixes #176.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/codegen/HollowConsumerJavaFileGenerator.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/codegen/HollowConsumerJavaFileGenerator.java
@@ -15,6 +15,16 @@
  */
 package com.netflix.hollow.api.codegen;
 
+import com.netflix.hollow.core.schema.HollowListSchema;
+import com.netflix.hollow.core.schema.HollowMapSchema;
+import com.netflix.hollow.core.schema.HollowSchema;
+import com.netflix.hollow.core.schema.HollowSetSchema;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
 /**
  * Not intended for external consumption.
  *
@@ -54,7 +64,18 @@ public abstract class HollowConsumerJavaFileGenerator implements HollowJavaFileG
     }
 
     protected void appendPackageAndCommonImports(StringBuilder builder) {
-        String fullPackageName = createFullPackageName(packageName, subPackageName, config.isUsePackageGrouping());
+        appendPackageAndCommonImports(builder, null, new ArrayList<HollowSchema>());
+    }
+
+    protected void appendPackageAndCommonImports(StringBuilder builder,
+            String apiClassname) {
+        appendPackageAndCommonImports(builder, apiClassname, new ArrayList<HollowSchema>());
+    }
+
+    protected void appendPackageAndCommonImports(StringBuilder builder,
+            String apiClassname, List<HollowSchema> schemasToImport) {
+        String fullPackageName =
+            createFullPackageName(packageName, subPackageName, config.isUsePackageGrouping());
         if (!isEmpty(fullPackageName)) {
             builder.append("package ").append(fullPackageName).append(";\n\n");
 
@@ -63,9 +84,42 @@ public abstract class HollowConsumerJavaFileGenerator implements HollowJavaFileG
             }
 
             if (config.isUsePackageGrouping()) {
-                builder.append("import ").append(packageName).append(".*;\n");
-                builder.append("import ").append(packageName).append(".core.*;\n");
-                if (useCollectionsImport) builder.append("import ").append(packageName).append(".collections.*;\n\n");
+                if (apiClassname != null) {
+                    appendImportFromBasePackage(builder, apiClassname);
+                }
+                Set<String> schemaNameSet = new HashSet<>();
+                for (HollowSchema schema : schemasToImport) {
+                    switch (schema.getSchemaType()) {
+                        case OBJECT:
+                            addToSetIfNotPrimitive(schemaNameSet, schema.getName());
+                            break;
+                        case SET:
+                            addToSetIfNotPrimitive(schemaNameSet,
+                                    ((HollowSetSchema) schema).getElementType());
+                            break;
+                        case LIST:
+                            addToSetIfNotPrimitive(schemaNameSet,
+                                    ((HollowListSchema) schema).getElementType());
+                            break;
+                        case MAP:
+                            HollowMapSchema mapSchema = (HollowMapSchema) schema;
+                            addToSetIfNotPrimitive(schemaNameSet, mapSchema.getKeyType(),
+                                    mapSchema.getValueType());
+                            break;
+                        default:
+                            throw new IllegalArgumentException(
+                                    "Unexpected HollowSchema to import: " + schema);
+
+                    }
+                }
+                for (String schemaName : schemaNameSet) {
+                    appendImportFromBasePackage(builder, schemaName);
+                }
+                appendImportFromBasePackage(builder, "core.*");
+                if (useCollectionsImport) {
+                    appendImportFromBasePackage(builder, "collections.*");
+                }
+                builder.append("\n");
             }
         }
     }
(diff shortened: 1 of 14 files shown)


hollow/src/main/java/com/netflix/hollow/api/codegen/HollowConsumerJavaFileGenerator.java
```diff
@@ -82,4 +137,17 @@ private String createFullPackageName(String packageName, String subPackageName,
...
     }
+
+    private void appendImportFromBasePackage(StringBuilder builder, String leaf) {
+        builder.append("import ").append(packageName).append(".").append(leaf)
+            .append(";\n");
+    }
+
+    private void addToSetIfNotPrimitive(Set<String> set, String... schemaNames) {
+        for (String schemaName : schemaNames) {
+            if (!HollowCodeGenerationUtils.isPrimitiveType(schemaName)) {
```
akhaku: I initially checked `!config.isUseHollowPrimitiveTypes() || ` here, assuming that isUseHollowPrimitiveTypes being `false` meant that we were generating even primitive types, but the tests failed 😕

hollow/src/main/java/com/netflix/hollow/api/codegen/HollowConsumerJavaFileGenerator.java
```diff
@@ -30,14 +40,15 @@
     protected String className;
     protected boolean useCollectionsImport=false;
 
-    public HollowConsumerJavaFileGenerator(String packageName, String subPackageName, CodeGeneratorConfig config) {
+    public HollowConsumerJavaFileGenerator(String packageName, String subPackageName,
```
davesu: I assume this is auto formatting.   If so, it would be best only to auto format on lines that were changed.  This way it will minimize diff change noises ;)

hollow/src/main/java/com/netflix/hollow/api/codegen/HollowConsumerJavaFileGenerator.java
```diff
@@ -63,9 +85,42 @@ protected void appendPackageAndCommonImports(StringBuilder builder) {
...
+                            HollowMapSchema mapSchema = (HollowMapSchema) schema;
+                            addToSetIfNotPrimitive(imports, mapSchema.getKeyType(),
+                                    mapSchema.getValueType());
+                            break;
+                        default:
+                            throw new IllegalArgumentException(
+                                    "Unexpected HollowSchema to import: " + schema);
+
+                    }
+                    for (String anImport : imports) {
```
davesu: Not sure if I am reading this correctly, but should this be outside of the for(HollowSchema) loop?

hollow/src/main/java/com/netflix/hollow/api/codegen/HollowConsumerJavaFileGenerator.java
akhaku: Oops, will revert, thanks!

hollow/src/main/java/com/netflix/hollow/api/codegen/HollowConsumerJavaFileGenerator.java
akhaku: Ahh you're right, this is incorrectly inside the for loop. Probably resulted in extra imports, which I didn't notice :/

hollow/src/main/java/com/netflix/hollow/api/codegen/HollowConsumerJavaFileGenerator.java
```diff
@@ -82,4 +137,17 @@ private String createFullPackageName(String packageName, String subPackageName,
     private boolean isEmpty(String value) {
         return value == null || value.trim().isEmpty();
     }
+
+    private void appendImportFromBasePackage(StringBuilder builder, String leaf) {
+        builder.append("import ").append(packageName).append(".").append(leaf)
+            .append(";\n");
+    }
+
+    private void addToSetIfNotPrimitive(Set<String> set, String... schemaNames) {
```
davesu: Minor since this is a small method but it might be better to rename "set" to importSet or something more descriptive ;)

akhaku: Can't hurt, will do

davesu: Actually, I am not clear on this method.  is it supposed to convert schemaName into Hollow Package Import String?

hollow/src/main/java/com/netflix/hollow/api/codegen/HollowConsumerJavaFileGenerator.java
```diff
@@ -63,9 +85,42 @@ protected void appendPackageAndCommonImports(StringBuilder builder) {
             }
 
             if (config.isUsePackageGrouping()) {
-                builder.append("import ").append(packageName).append(".*;\n");
-                builder.append("import ").append(packageName).append(".core.*;\n");
-                if (useCollectionsImport) builder.append("import ").append(packageName).append(".collections.*;\n\n");
+                if (apiClassname != null) {
+                    appendImportFromBasePackage(builder, apiClassname);
+                }
+                Set<String> imports = new HashSet<>();
```
davesu: Perhaps rename "imports" to schemaNameSet since I thought that contains the full import strings.  Perhaps it was just me ;)

akhaku: Ah, yes, will do.

hollow/src/main/java/com/netflix/hollow/api/codegen/HollowConsumerJavaFileGenerator.java
```diff
@@ -82,4 +136,23 @@ private String createFullPackageName(String packageName, String subPackageName,
     private boolean isEmpty(String value) {
         return value == null || value.trim().isEmpty();
     }
+
+    private void appendImportFromBasePackage(StringBuilder builder, String leaf) {
```
davesu, line 140: Do we need to make sure the firstChar of the leaf is upper case?  

I believe the Code generator upper case the first char of the schema name.

hollow/src/main/java/com/netflix/hollow/api/codegen/HollowConsumerJavaFileGenerator.java
```diff
@@ -63,9 +84,42 @@ protected void appendPackageAndCommonImports(StringBuilder builder) {
...
+                                    "Unexpected HollowSchema to import: " + schema);
+
+                    }
+                }
+                for (String schemaName : schemaNameSet) {
+                    appendImportFromBasePackage(builder, schemaName);
+                }
+                appendImportFromBasePackage(builder, "core.*");
+                if (useCollectionsImport) {
+                    appendImportFromBasePackage(builder, "collections.*\n");
```
davesu: looks like it has extra "\n".

hollow/src/main/java/com/netflix/hollow/api/codegen/HollowConsumerJavaFileGenerator.java
```diff
@@ -82,4 +136,23 @@ private String createFullPackageName(String packageName, String subPackageName,
     private boolean isEmpty(String value) {
         return value == null || value.trim().isEmpty();
     }
+
+    private void appendImportFromBasePackage(StringBuilder builder, String leaf) {
+        builder.append("import ").append(packageName).append(".").append(leaf)
+            .append(";\n");
```
davesu: minor but it seems like your line wrap is a bit aggressive.  Perhaps in the future it would be nice to be a bit more lenient ;)

hollow/src/main/java/com/netflix/hollow/api/codegen/HollowConsumerJavaFileGenerator.java
akhaku: The schema name comes in cased correctly in my testing. Also, we can't do it here, since we pass in things like `core.*` and `collections.*` to this method (see :118 and :120)

hollow/src/main/java/com/netflix/hollow/api/codegen/HollowConsumerJavaFileGenerator.java
akhaku: indeed, fixing

hollow/src/main/java/com/netflix/hollow/api/codegen/HollowConsumerJavaFileGenerator.java
akhaku: yeah, I've been wrapping at 90 for a few years, a little more difficult with 4 space indents, though

ghost (approved):

akhaku:
Squashed commits, merging after build passes.

PR #179, opened 2018-03-01
Remove new BaseHollowProducerListener in favor of existing class

akhaku:
Looks like we already have a no-op base version of a
HollowProducerListener, so remove our newly added one.

We haven't released this new class yet, so it's okay to remove.


ghost:
And...I'm the author of that base class. Oops.

PR #180, opened 2018-03-03
Don't fire validation start listener handle twice

akhaku:
For cases in which we're given a HollowValidationListener
and a HollowProducerListener that happen to be the same
object, don't fire the listener twice. This is only needed
for onValidationStart since
HollowProducerListener#onValidationStart has the same
signature as HollowValidationListener#onValidationStart.


akhaku:
A little torn by this... the other way to fix this is to change the signature of one of the methods, possibly by simply renaming the method. Thoughts?

lkancode (approved):

PR #178, opened 2018-03-01
IncrementalCyclePopulator will not attempt to remove records from a n…

duro1:
…ull previous state

Allows incremental producer to directly create the first snapshot without relying on a special run of the underlying classic producer
This should also address issue 154 (https://github.com/Netflix/hollow/issues/154)

PR #181, opened 2018-03-03
HollowIncrementalProducer - added restoreFromLastState [proposal]

rpalcolea:
This pull request:

* Introduces Builder pattern for `HollowIncrementalProducer`
* Users could provide `AnnouncementWatcher, BlobRetriever and data model`
* Users can now restore from last state using the `HollowIncrementalProducer.restoreFromLastState)`.
* Kept public constructors: these could be removed later. I know `HollowIncrementalProducer` is a BETA API but for now I think it makes sense to keep them and perhaps deprecate in Hollow 3?

With the introduction of https://github.com/Netflix/hollow/pull/178 and this, users could now write the first cycle using the incremental directly and restore from previous states. This would significantly reduce their code by avoiding all that logic that basically everyone would have to write in order to restore an incremental producer.

In the future `HollowIncrementalProducer` could be just another type of `HollowProducer` but that's not in the scope of this pull request.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalProducer.java
@@ -17,7 +17,9 @@
  */
 package com.netflix.hollow.api.producer;
 
+import com.netflix.hollow.api.consumer.HollowConsumer;
 import com.netflix.hollow.api.consumer.HollowConsumer.BlobRetriever;
+import com.netflix.hollow.api.consumer.fs.HollowFilesystemAnnouncementWatcher;
 import com.netflix.hollow.core.write.objectmapper.RecordPrimaryKey;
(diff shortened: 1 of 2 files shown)


rpalcolea:
Hi @toolbear , please let me know your thoughts on this

duro1:
@rpalcolea This looks good.  It currently is a little involved to know how and in what order to initialize the classic producer in order to properly use the incremental producer.
What if instead of a builder, the newly proposed constructor arguments were mandatory, in the next major release?

This would allow the client to call the restoreFromLastState at any point, but know that the necessary objects should be present. This would also allow for controls to be added around running the first cycle only after restoring (or verifying a previous does not exist)

e.g.
private boolean isInitialized;

    restoreFromLastState(...)
        (attempt restore)
        isInitialized=true;

    public long runCycle() {
        if(!isInitialized)
            throw new IllegalStateException("You must restore state prior running an IncrementalProducer cycle");
        ...

PR #182, opened 2018-03-06
HollowIncrementalProducer - only clean changes when a cycle is written

rpalcolea:
Hi @toolbear,

This prevents the `HollowIncrementalProducer` from cleaning changes (`mutations`) when a `runCycle` happens and the producer isn't primary. Currently it cleans the mutations and in the case of an incremental producer ingested by kafka changes (in our case), we could lose that data. With this change, this will only clean the mutations only when a new version is written. I think in the future, the incremental producer could listen for `fireCycleSkipped`. However, `fireCycleSkipped` is part of `HollowProducerListenerV2` and I think it would be nice to introduce this once the listeners get consolidated.

Thoughts?


ghost:
I'm uncertain about the incremental producer having to rely on a listener hook to do the right thing. The listeners feel more like out-of-band signaling to me, and we should build something more explicit/direct. As a stop-gap, maybe the listener. Ultimately it may be that the incremental producer shouldn't compose a vanilla producer, but instead they share a common base class or we extract out components that can used in either type of producer.

ghost (changes requested):
This is good as a stopgap. Can you resolve the conflict? Then we can merge.

PR #192, opened 2018-04-08
HollowPOJOGenerator bugfixes and improvements

akhaku:


akhaku:
Probably best to look at the commits here individually - will make for an easier review.

PR #195, opened 2018-04-14
Introduce and throw IncompatibleSchemaException

akhaku:
As part of this, introduce the generic HollowException. The
idea is over time we add subclasses of HollowException for
different "expected" errors that can occur.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/error/IncompatibleSchemaException.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/error/IncompatibleSchemaException.java
@@ -1,0 +1,54 @@
+/*
+ *
+ *  Copyright 2018 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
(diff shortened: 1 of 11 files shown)


hollow/src/main/java/com/netflix/hollow/api/error/IncompatibleSchemaException.java
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * An exception thrown when trying to compare two versions of an object with incompatible schema.
+ */
+public class IncompatibleSchemaException extends HollowException {
+    private final String typeName;
+    private final String fieldName;
+
+    public IncompatibleSchemaException(String typeName, String fieldName) {
+        super("No common schema exists for " + typeName + ": field " + fieldName
```
davesu: It would be good to capture more context if available. For example, types changed.

davesu (approved):

hollow/src/main/java/com/netflix/hollow/api/error/IncompatibleSchemaException.java
akhaku: Hm good idea, I'll add some more info.

rpalcolea:
this is nice! great addition

Issue #140, opened 2017-12-08
It's not possible to start a consumer application when blob store directory is empty


tdudzik:
When you are trying to start an application and create indexes and your blob store dir is empty the following exceptions are thrown:

- for primary key index:
```
java.lang.NullPointerException
	at com.netflix.hollow.api.client.HollowClientUpdater.getStateEngine(HollowClientUpdater.java:166)
	at com.netflix.hollow.api.consumer.HollowConsumer.getStateEngine(HollowConsumer.java:246)
	at [your-api-package].[IndexName]PrimaryKeyIndex.<init>([IndexName]PrimaryKeyIndex.java:15)
```

- for hash index:
```
java.lang.NullPointerException
	at com.netflix.hollow.api.client.HollowClientUpdater.getAPI(HollowClientUpdater.java:170)
	at com.netflix.hollow.api.consumer.HollowConsumer.getAPI(HollowConsumer.java:260)
	at [your-api-package].[ApiName]APIHashIndex.<init>([ApiName]APIHashIndex.java:25)
```

Both exceptions are related to `HollowClientUpdater#hollowDataHolder` field which is `null` in that case.

johnrengelman:
Been hitting this one pretty regularly and can't seem to find a way to detect it other than wrapping this in `try/catch` for NPEs.

brettwooldridge:
We always just call mkDirs() on the store directory before starting the client.

johnrengelman:
Well, we are writing our own backing (using redis) and get the same problem. I guess I could look to see if it’s something in there. But it seems like the hollow state just doesn’t fully initialize unless there is data.

rpalcolea:
@brettwooldridge @johnrengelman,

I think this is a expected behavior, there was some conversation around this in the Gitter room. I guess the work on this would be a meaningful exception.  

In our case, we initialize the indexes once we triggered a refresh on the consumer and the `StateEngine` is ready. Our implementation pulls the blobs from Google Cloud Storage and keeps a local copy with `LocalBlobStore`. At the end we do the `mkdirs()` and wrap the index creation in a try/catch, something like:

```groovy
ProductPrimaryKeyIndex idx
        consumer = HollowConsumer.withBlobRetriever(gcsBlobRetriever)
                .withLocalBlobStore(new File("//blob-test").mkdirs())
                .withAnnouncementWatcher(announcementWatcher)
                .withGeneratedAPIClass(ProductAPI)
[...]

PR #197, opened 2018-04-16
Initialize consumer to empty read state if no announcements #140

akhaku:
If the AnnouncementWatcher has no published versions when
a HollowConsumer refreshes, initialize it to an empty read
state. Ensure that the next refresh forces a snapshot,
even if we are configured to disallow snapshots.
While we're here, also start the process of consolidating
sentinel values.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/HollowConstants.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/HollowConstants.java
@@ -1,0 +1,33 @@
+/*
+ *
+ *  Copyright 2018 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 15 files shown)


akhaku:
This PR contains a bit of a cleanup, but the meat of it is in `HollowClientUpdater` - initializing the data state to an empty `HollowReadStateEngine` if there are no published versions. If you `triggerRefresh` after something is published (or wait for `triggerRefreshAsync` to kick in on its own) you will be updated to the latest version via a snapshot update.
An important side-effect of the empty state is that there are _no_ types available in your consumer, so even initializing a `PrimaryKey` will fail. You can check if you have been initialized yet by calling `HollowConsumer#getCurrentVersionId` and comparing it to `HollowConstants#VERSION_NONE`.

hollow/src/main/java/com/netflix/hollow/api/HollowConstants.java
```diff
@@ -0,0 +1,33 @@
...
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.hollow.api;
+
+/**
+ * An interface to gather various sentinel constants used across hollow.
+ */
+public interface HollowConstants {
```
davesu, line 23: +1, nice.

davesu (approved):

PR #198, opened 2018-04-16
Add a class to facilitate creating read state engines for testing

akhaku:
This helper class is exposed in a new "hollow-test" jar.
This jar will contain utilities to aid in testing client
code that uses hollow.


hollow-test/src/main/java/com/netflix/hollow/test/HollowReadStateEngineBuilder.java
```diff
@@ -0,0 +1,67 @@
...
+    private final HollowObjectMapper objectMapper;
+
+    private boolean built;
+
+    /**
+     * Create a HollowReadStateEngineBuilder with an empty initial type state. Adding objects will
+     * add their types, if not already present.
+     */
+    public HollowReadStateEngineBuilder() {
+        this(new ArrayList<Class<?>>());
```
davesu: Minor: Collectons.emptyList() ?

akhaku: Sure, will do. Note to self: make sure to specify the type ie `Collections.<Class<?>>.emptyList()`

hollow-test/src/main/java/com/netflix/hollow/test/HollowReadStateEngineBuilder.java
```diff
@@ -0,0 +1,67 @@
...
+        for (Class<?> type : types) {
+            objectMapper.initializeTypeState(type);
+        }
+    }
+
+    /**
+     * Add an object that will be in our built HollowReadStateEngine. You cannot add any more
+     * objects after calling build().
+     */
+    public HollowReadStateEngineBuilder add(Object object) {
```
davesu: I wonder if add(Object ... objs) will be good to have instead.

davesu (approved):

hollow-test/src/main/java/com/netflix/hollow/test/HollowReadStateEngineBuilder.java
akhaku: Can't hurt, that supports the current usage pattern and others, too. Will do.

davesu (approved):

PR #199, opened 2018-04-16
allow to force enabling a producer

rpalcolea:


rpalcolea:
thoughts on this @akhaku @toolbear @dkoszewnik

hollow/src/main/java/com/netflix/hollow/api/producer/enforcer/SingleProducerEnforcer.java
```diff
@@ -39,4 +39,10 @@
      */
     boolean isPrimary();
 
+
+    /**
+     * Force Maring producer to enable processing cycles.
```
akhaku: `Maring`?

hollow/src/main/java/com/netflix/hollow/api/producer/enforcer/AbstractSingleProducerEnforcer.java
```diff
@@ -34,6 +34,8 @@
 
     protected abstract boolean _isPrimary();
 
+    protected abstract void _force();
```
akhaku: Let's implement this to throw an UnsupportedOperationException by default? This way we won't break existing clients?

akhaku (commented):
Generally seems good with the changes

rpalcolea:
Thanks for the feedback, @akhaku. Made the changes

akhaku (approved):
LGTM, but I'll let @toolbear take a look too

PR #202, opened 2018-04-18
Added SimultaneousExecutor for markTypeRecordsToRemove in HollowIncrementalProducer

rpalcolea:


rpalcolea:
This is to mark records to remove on each incremental cycle with a `SimultaneousExecutor`. Instead of going one at a time querying the index

@toolbear @akhaku

akhaku (approved):
LGTM, but I'l let @toolbear take a look too

Issue #159, opened 2018-01-24
ArrayOutOfBoundException: -1 when searching by Hash Index


dzmitryh:
Hey guys! Would like to share with you my issue.

## Description:

When I try to use hash index and then search by it I get `ArrayOutOfBoundException`

It is based on this sources https://github.com/Netflix/hollow/blob/7324212d7a892f5a04ee4d83885683042ad5a166/hollow/src/test/java/com/netflix/hollow/core/index/HollowHashIndexTest.java#L32
index example: 
```
HollowHashIndex index = new HollowHashIndex(readStateEngine, "TypeB", "", "b1.value");
```

## Stack trace:
```
java.lang.ArrayIndexOutOfBoundsException: -1

	at com.netflix.hollow.core.memory.encoding.FixedLengthElementArray.getElementValue(FixedLengthElementArray.java:94)
	at com.netflix.hollow.core.memory.encoding.FixedLengthElementArray.getElementValue(FixedLengthElementArray.java:85)
	at com.netflix.hollow.core.read.engine.object.HollowObjectTypeReadStateShard.readString(HollowObjectTypeReadStateShard.java:216)
	at com.netflix.hollow.core.read.engine.object.HollowObjectTypeReadState.readString(HollowObjectTypeReadState.java:188)
	at com.netflix.hollow.core.read.HollowReadFieldUtils.fieldValueEquals(HollowReadFieldUtils.java:193)
	at com.netflix.hollow.core.index.HollowHashIndex.matchIsEqual(HollowHashIndex.java:165)
	at com.netflix.hollow.core.index.HollowHashIndex.findMatches(HollowHashIndex.java:103)
	at com.netflix.hollow.core.index.HollowHashIndexTest.testIndexingColonIncludedValuesThrowsArrayOutOfBound(HollowHashIndexTest.java:95)
```

## Unit test:
I've created unit test in latest sources to reproduce it:

```
    @Test
    public void testIndexingColonIncludedValuesThrowsArrayOutOfBound() throws IOException {
        HollowObjectMapper mapper = new HollowObjectMapper(writeStateEngine);

        final String b1ContainsColon = "one:";
        mapper.add(new TypeB(null));

        roundTripSnapshot();

        HollowHashIndex index = new HollowHashIndex(readStateEngine, "TypeB", "", "b1.value");
        index.findMatches(b1ContainsColon);
    }
```
## Observations:
* if u delete `colon` symbol from `b1ContainsColon` variable the test passes;
[...]

PR #206, opened 2018-04-23
Fix for ArrayOutOfBoundException in HasfIndex (#159)

aidarko:
Fix for issue: #159 

Removed obsolete block:
`if(hashOrdinal == -1 && query[i] == null)
continue;`

Added prechecks to HollowObjectTypeReadStateShard.java in case of indexing objects.
Added tests (might be run without changes in order to determine the problem).

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectTypeReadStateShard.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectTypeReadStateShard.java
@@ -144,6 +144,9 @@ class HollowObjectTypeReadStateShard {
     }
 
     public Boolean readBoolean(int ordinal, int fieldIndex) {
+        if (ordinal < 0)
+            return null;
+
         HollowObjectTypeDataElements currentData;
         long value;
 
@@ -167,6 +170,9 @@ class HollowObjectTypeReadStateShard {
     }
 
     public byte[] readBytes(int ordinal, int fieldIndex) {
+        if (ordinal < 0)
+            return null;
+
         HollowObjectTypeDataElements currentData;
         byte[] result;
 
@@ -200,6 +206,9 @@ class HollowObjectTypeReadStateShard {
     }
 
     public String readString(int ordinal, int fieldIndex) {
+        if (ordinal < 0)
+            return null;
+
         HollowObjectTypeDataElements currentData;
         String result;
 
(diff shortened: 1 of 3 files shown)


hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectTypeReadStateShard.java
```diff
@@ -144,6 +144,9 @@ public long readLong(int ordinal, int fieldIndex) {
     }
 
     public Boolean readBoolean(int ordinal, int fieldIndex) {
+        if (ordinal < 0)
```
dzmitryh, line 147: should we apply the same check for other methods like `readLong` `readDouble` etc. ?

aidarko: No, in fact long and int works pretty well.
I've tried to run tests in order to crash, but without any result. 
Probably this scenario is the most tested.

akhaku: curious - what do we return for long and int? sentinel values?

aidarko: My bad, I've made the same mistake as previous developer - values go to different buckets. There's still the issue with readLong, etc... Here it's a bit more tricky due to the fact, that method uses primitives, but ordinal still might be -1.

aidarko: What about `if (ordinal < 0) return Long.MIN_VALUE;` in order to convert to a sentinel value?
The best option would be to return null, but that would mean changing method signature to use Long / Integer / Double , which would lead to quite a big change (and definitely to some side effects).

aidarko: I see, there's something like `DefaultMissingDataHandler.java`, which already have all this stuff.

akhaku: I think we may want to investigate why the callers of these methods are passing in empty ordinals - it may be that we make the change further upstream? It feels like calling readBoolean/readOtherType with an ordinal of -1 is an invalid usage pattern. We could decide to throw IllegalArgumentException there, for instance, and fix the caller to not call with -1.

aidarko: I've came up to the same decision. I'll open another PR, closing this one.

Issue #204, opened 2018-04-20
Allow tweaking RecordCountValidator variance at runtime


akhaku:
Requirements:
- users should be able to tweak the allowableVariancePercent at runtime, per instance of RecordCountVarianceValidator
- allowableVariancePercent should be fixed for the duration of a cycle - if you make a tweak mid-cycle, the new value doesn't take effect until next cycle

PR #208, opened 2018-04-24
introduce RecordCountValidator variance at runtime

rpalcolea:
This is for https://github.com/Netflix/hollow/issues/204

Introduces a breaking change in the `Validator` by adding `lock` method.

The idea would be that users could do some changes to their validators on runtime and once a cycle gets into the validation state, this would happen:

```
validator.lock();
validator.validate(readState);
```

From the `RecordCountVarianceValidator`, `lock` assigns the `allowableVariancePercent` to `cycleAllowableVariancePercent` which is used for the validation. This way, `allowableVariancePercent` could change any time via `updateVariancePercent` without affecting the current cycle.

Other idea in order to prevent breaking changes:

* Introduce a `ValidatorV2` interface (same way it happened with the `Listeners`) which would contain `validate(readState)` and `lock`. 
* Introduce a `DynamicRecordCountVarianceValidator` that uses `ValidatorV2` and has the functionality in https://github.com/Netflix/hollow/issues/204.

I'm also asking myself if `lock` is a good name or perhaps use something like `prepareForReadStateValidation` to keep the semantics

Thoughts @toolbear @akhaku ?


hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -917,6 +918,7 @@ public long getToVersion() {
     
     public static interface Validator {
         void validate(HollowProducer.ReadState readState);
+        void lock();
```
ghost: With Hollow 3, now we can use a default method and avoid the breaking change.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -89,6 +89,11 @@ public void validate(ReadState readState) {
 		handleEndValidation(statusBuilder, Status.SUCCESS, "");
 	}
 
+	@Override
```
ghost: ...and this can go away and rely on the default method

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -687,6 +687,7 @@ private void validate(HollowProducer.ReadState readState) {
     		for(Validator validator: validators) {
     			Throwable throwable = null;
     			try {
+    			    validator.lock();
```
ghost: Tabs. Bleh. ;)

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
ghost: I think a name like `lock()` should be reserved for concurrency type locks.

Let me discuss some naming ideas with @akhaku. My first thoughts are something like `freeze`, `freezeConfig`, `commitConfig`, `sealConfig`, …. Not real happy with any of them.

ghost (changes requested):

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
akhaku: we've used freeze in the past?

rpalcolea:
This is ready for review again @toolbear

ghost:
Sorry for there not being any action on this. We're revamping the validator API in a big way, so we're holding off on accepting changes like this until we settle on the approach we're going to use for runtime configuration.

@PaulSandoz has some good ideas on this.

rpalcolea:
Sounds good, thanks for the heads up. Closing this

PaulSandoz:
Two ideas:

1. Separate the listeners into component parts corresponding to stages (cycle, populate, publish, integrity check, validate, announce).  A validator is "just" another listener, an implementation can implement the validator listener interface and the cycle listener interface.  Using the latter a method implementation for `onCycleStart` can "snapshot"  validator state. 

2. Building on 1) it's possible to define a listener that has life-cycle.  A special supplier-based listener can supply a listener instance for a particular listener type (or set) it implements.  This enables integration with say Guice or Spring without binding those DI frameworks to Hollow.   The scope could be on demand or perhaps scoped to the cycle.

Issue #209, opened 2018-04-24
Exceptions thrown by HollowProducerListeners should be caught and logged


akhaku:
Currently, if a HollowProducerListener throws an exception in any of its methods, that prevents other listeners from being called and potentially fails the cycle. We should instead catch and log any RuntimeExceptions thrown by HollowProducerListeners.

rpalcolea:
This can be closed now (forgot to add a closes #209 in the commit :()

PR #211, opened 2018-04-25
Exceptions thrown by HollowProducerListeners should be caught and logged

rpalcolea:
This is a simple implementation for https://github.com/Netflix/hollow/issues/209

I think it could be better with Java 8+ and functions... at least to abstract the error handling logic.

thoughts @toolbear @akhaku

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/ListenerSupport.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/ListenerSupport.java
@@ -41,8 +41,8 @@ import java.util.logging.Logger;
  */
 final class ListenerSupport {
 
-    private final Logger log = Logger.getLogger(ListenerSupport.class.getName());
-
+    private static final Logger LOG = Logger.getLogger(ListenerSupport.class.getName());
+    
     private final Set<HollowProducerListener> listeners;
     private final Set<HollowValidationListener> validationListeners;
     private final Set<IncrementalCycleListener> incrementalCycleListeners;
@@ -78,7 +78,7 @@ final class ListenerSupport {
             try {
                 l.onProducerInit(elapsedMillis, MILLISECONDS);
             } catch(Throwable t) {
-                log.warning("Error executing listener" + t);
+                LOG.warning("Error executing listener" + t);
             }
         }
     }
@@ -88,7 +88,7 @@ final class ListenerSupport {
             try {
                 l.onProducerRestoreStart(version);
             } catch(Throwable t) {
-                log.warning("Error executing listener" + t);
+                LOG.warning("Error executing listener" + t);
             }
         }
     }
@@ -98,7 +98,7 @@ final class ListenerSupport {
             try {
                 l.onProducerRestoreComplete(status, elapsedMillis, MILLISECONDS);
             } catch(Throwable t) {
-                log.warning("Error executing listener" + t);
+                LOG.warning("Error executing listener" + t);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


hollow/src/main/java/com/netflix/hollow/api/producer/ListenerSupport.java
```diff
@@ -70,128 +74,260 @@ void remove(IncrementalCycleListener listener) {
     }
 
     void fireProducerInit(long elapsedMillis) {
-        for(final HollowProducerListener l : listeners) l.onProducerInit(elapsedMillis, MILLISECONDS);
+        for(final HollowProducerListener l : listeners) {
+            try {
+                l.onProducerInit(elapsedMillis, MILLISECONDS);
+            } catch(RuntimeException e) {
+                log.warning("Execution of " + l.getClass().getName() + "  during fireProducerRestoreStart failed | " + e);
```
akhaku: Let's use the [version](https://docs.oracle.com/javase/7/docs/api/java/util/logging/Logger.html#log(java.util.logging.Level,%20java.lang.String,%20java.lang.Throwable)) that takes a Throwable instead, so we can get stack traces.
At that point we could even simplify the message to something like "Error executing listener", since we'll have enough details in the stack trace.

hollow/src/test/java/com/netflix/hollow/api/producer/ListenerSupportTest.java
```diff
@@ -58,4 +60,97 @@ public void testFireValidationStart() {
...
+
+
+    @Test
+    public void fireAnnouncementStartDontStopWhenOneFails() {
+        long version = 31337;
+        HollowProducer.ReadState readState = Mockito.mock(HollowProducer.ReadState.class);
+        Mockito.when(readState.getVersion()).thenReturn(version);
+        Mockito.doThrow(RuntimeException.class).when(listener).onAnnouncementStart(version);
+        listenerSupport.fireAnnouncementStart(readState);
+        Mockito.verify(listener).onAnnouncementStart(version);
```
akhaku, line 154: Nice, thanks for the test coverage.

akhaku (commented):
Looks good, just the minor changes to get the stack trace. I agree that this could be a lot cleaner with Java 8, hopefully we drop Java 7 support soon!

hollow/src/main/java/com/netflix/hollow/api/producer/ListenerSupport.java
```diff
@@ -70,128 +74,260 @@ void remove(IncrementalCycleListener listener) {
     }
 
     void fireProducerInit(long elapsedMillis) {
-        for(final HollowProducerListener l : listeners) l.onProducerInit(elapsedMillis, MILLISECONDS);
+        for(final HollowProducerListener l : listeners) {
+            try {
+                l.onProducerInit(elapsedMillis, MILLISECONDS);
+            } catch(Throwable t) {
+                log.warning("Error executing listener" + t);
```
akhaku: Ah sorry I was unclear :( what I meant was to continue catching RuntimeException, but at the _log_ line update to something like:
`log.log(Level.WARN, "Error executing listener", e)`

hollow/src/main/java/com/netflix/hollow/api/producer/ListenerSupport.java
```diff
@@ -32,13 +32,17 @@
...
+import java.util.logging.Logger;
 
 /**
  * Beta API subject to change.
  *
  * @author Tim Taylor {@literal<tim@toolbear.io>}
  */
 final class ListenerSupport {
+
+    private final Logger log = Logger.getLogger(ListenerSupport.class.getName());
```
akhaku: Minor but can we make this static final?

`private static final Logger LOG = Logger.getLogger(ListenerSupport.class.getName());`

Similar to the logger in HollowClientUpdater.

rpalcolea:
done! @akhaku

PR #215, opened 2018-05-01
Throw useful exception in primary key init if no type

akhaku:
Instead of failing with an NPE, throw a descriptive
exception when constructing a primary key for a missing
type.
Note that if you call ReadStateEngine.getSchema(typeName)
yourself, you will continue to get back null. This new
exception is only thrown when creating a unique key, in
order to have callers not have to catch a
NullPointerException.


ghost (changes requested):
As discussed offline, extract a `getSchema` to a helper or abstract base class so that we don't have to generate the `getSchema` helper each index.

akhaku:
Added a getNotNullSchema to HollowDataSet and HollowStateEngine, using that now. Merging.

PR #216, opened 2018-05-02
ArrayOutOfBoundException: -1 when searching by Hash Index (#159)

aidarko:
Add tests with evidence of the issue, add relevant fixes, propose refactoring (in separate commit).

PR created in favor of https://github.com/Netflix/hollow/pull/206
Instead of making changes in HollowObjectTypeReadStateShard moved prechecks to HollowHashIndexBuilder and HollowHashIndex.

I've found many places, which are using magic -1 value :)
Here is the small list:
- com/netflix/hollow/tools/diff/count/HollowDiffObjectCountingNode.java:128
- com/netflix/hollow/tools/diff/count/HollowDiffObjectCountingNode.java:175
- com/netflix/hollow/tools/diff/exact/mapper/DiffEqualityObjectMapper.java:101
- com.netflix.hollow.diffview.HollowObjectDiffViewGenerator#fromOrdinal
- com/netflix/hollow/diffview/effigy/HollowEffigyFactory.java:47
- ...
So this PR should fit the philosophy.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
@@ -160,10 +160,9 @@ public class HollowHashIndex implements HollowTypeStateListener {
 
                 HollowObjectTypeReadState objectAccess = (HollowObjectTypeReadState)readState;
                 int fieldIdx = fieldPath[fieldPath.length-1];
-                if(hashOrdinal == -1 && query[i] == null)
-                    continue;
-                if(!HollowReadFieldUtils.fieldValueEquals(objectAccess, hashOrdinal, fieldIdx, query[i]))
+                if(hashOrdinal == -1 || !HollowReadFieldUtils.fieldValueEquals(objectAccess, hashOrdinal, fieldIdx, query[i])) {
                     return false;
+                }
             }
         }
 
--- a/hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
@@ -91,7 +91,6 @@ public class HollowHashIndex implements HollowTypeStateListener {
             hashCode ^= HashCodes.hashInt(keyHashCode(query[i], i));
         }
 
-        //System.out.println("QUERY HASH: " + hashCode);
         HollowHashIndexResult result;
         do {
             result = null;
@@ -109,7 +108,7 @@ public class HollowHashIndex implements HollowTypeStateListener {
                 }
 
                 bucket = (bucket + 1) & hashState.getMatchHashMask();
-                hashBucketBit = (long) bucket * hashState.getBitsPerMatchHashEntry();
+                hashBucketBit = bucket * hashState.getBitsPerMatchHashEntry();
                 bucketIsEmpty = hashState.getMatchHashTable().getElementValue(hashBucketBit, hashState.getBitsPerTraverserField()[0]) == 0;
             }
         } while (hashState != hashStateVolatile);
@@ -122,15 +121,15 @@ public class HollowHashIndex implements HollowTypeStateListener {
         case BOOLEAN:
             return HollowReadFieldUtils.booleanHashCode((Boolean)key);
         case DOUBLE:
-            return HollowReadFieldUtils.doubleHashCode(((Double)key).doubleValue());
+            return HollowReadFieldUtils.doubleHashCode((Double) key);
         case FLOAT:
-            return HollowReadFieldUtils.floatHashCode(((Float)key).floatValue());
+            return HollowReadFieldUtils.floatHashCode((Float) key);
         case INT:
-            return HollowReadFieldUtils.intHashCode(((Integer)key).intValue());
+            return HollowReadFieldUtils.intHashCode((Integer) key);
         case LONG:
-            return HollowReadFieldUtils.longHashCode(((Long)key).longValue());
+            return HollowReadFieldUtils.longHashCode((Long) key);
         case REFERENCE:
-            return ((Integer)key).intValue();
+            return (Integer) key;
         case BYTES:
             return HashCodes.hashCode((byte[])key);
         case STRING:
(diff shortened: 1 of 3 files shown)


hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
```diff
@@ -146,10 +145,10 @@ private boolean matchIsEqual(FixedLengthElementArray matchHashTable, long hashBu
             int hashOrdinal = (int)matchHashTable.getElementValue(hashBucketBit + hashState.getOffsetPerTraverserField()[field.getBaseIteratorFieldIdx()], hashState.getBitsPerTraverserField()[field.getBaseIteratorFieldIdx()]) - 1;
 
             HollowTypeReadState readState = field.getBaseDataAccess();
-            int fieldPath[] = field.getSchemaFieldPositionPath();
+            int[] fieldPath = field.getSchemaFieldPositionPath();
 
             if(fieldPath.length == 0) {
-                if(hashOrdinal != ((Integer)query[i]).intValue())
+                if(hashOrdinal != (Integer) query[i])
```
akhaku: This is correct with the current behaviour, but can we actually switch to using `.equals` just in case hashOrdinal ever becomes an Integer instead of a primitive int?

hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
```diff
@@ -160,10 +159,9 @@ private boolean matchIsEqual(FixedLengthElementArray matchHashTable, long hashBu
 
                 HollowObjectTypeReadState objectAccess = (HollowObjectTypeReadState)readState;
                 int fieldIdx = fieldPath[fieldPath.length-1];
-                if(hashOrdinal == -1 && query[i] == null)
-                    continue;
-                if(!HollowReadFieldUtils.fieldValueEquals(objectAccess, hashOrdinal, fieldIdx, query[i]))
+                if(hashOrdinal == -1 || !HollowReadFieldUtils.fieldValueEquals(objectAccess, hashOrdinal, fieldIdx, query[i])) {
```
akhaku, line 162: Hm, mind explaining this change? Mostly just for my benefit.

hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndexBuilder.java
```diff
@@ -376,14 +373,18 @@ private boolean intermediateMatchIsEqual(int matchIdx, long hashBucketBit) {
         return true;
     }
 
+    private boolean isAnyFieldNull(int matchOrdinal, int hashOrdinal) {
+        return matchOrdinal == -1 || hashOrdinal == -1;
```
akhaku, line 377: nice

akhaku (commented):
Nice work! Just a minor comment, otherwise LGTM.
Also thanks for the cleanup, much appreciated 👍

hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
aidarko: `query[i] == null` is never true due to precheck before:
com/netflix/hollow/core/index/HollowHashIndex.java:89

Another thing, if we leave `continue`, then null value (in our context ordinal with the -1 value) will be interpreted as any. So it will match to any value, like "-1" ~ "one", "-1" ~ "two", ...
For instance, if we'll try to leave `continue`, then `index.findMatches("one:")` from tests will return something, despite the fact, that we didn't put "one:" anywhere.

hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
aidarko: I guess we'd have to do that... 
BTW, this refactoring is relevant to Java 5.0 and higher. But hopefully it's ok.
This rule had been taken from [here](https://github.com/JetBrains/intellij-community/blob/master/plugins/InspectionGadgets/src/inspectionDescriptions/UnnecessaryUnboxing.html)

akhaku: Yup, definitely agree with the refactoring in general, thanks for that!

hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
akhaku: Awesome, got it, thanks!

hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
akhaku: To provide a bit of additional context: I only ask for the comparison change here because we've been bitten but this in the past, we had something like this but a refactoring changed the primitive to an Integer and equality checks started failing ([for values above 127](https://docs.oracle.com/javase/7/docs/api/java/lang/Integer.html#valueOf%28int%29), so unit tests didn't catch them).

aidarko: I see now, it's for the sake of the future :)
It might look like that in this case:
`if (query[i].equals(hashOrdinal))
                    return false;`

akhaku: Sure, that works

hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
```diff
@@ -148,7 +148,7 @@ private boolean matchIsEqual(FixedLengthElementArray matchHashTable, long hashBu
             int[] fieldPath = field.getSchemaFieldPositionPath();
 
             if(fieldPath.length == 0) {
-                if(hashOrdinal != (Integer) query[i])
+                if (query[i].equals(hashOrdinal))
```
akhaku: Looks like you forgot the `!`

hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
```diff
@@ -146,10 +145,10 @@ private boolean matchIsEqual(FixedLengthElementArray matchHashTable, long hashBu
             int hashOrdinal = (int)matchHashTable.getElementValue(hashBucketBit + hashState.getOffsetPerTraverserField()[field.getBaseIteratorFieldIdx()], hashState.getBitsPerTraverserField()[field.getBaseIteratorFieldIdx()]) - 1;
 
             HollowTypeReadState readState = field.getBaseDataAccess();
-            int fieldPath[] = field.getSchemaFieldPositionPath();
+            int[] fieldPath = field.getSchemaFieldPositionPath();
 
             if(fieldPath.length == 0) {
-                if(hashOrdinal != ((Integer)query[i]).intValue())
+                if (query[i].equals(hashOrdinal))
```
akhaku: needs a `!`

aidarko: Indeed! Thank you a lot !!
Tests didn't help here unfortunately

akhaku:
Solid work, thanks for the contribution!

PR #222, opened 2018-05-17
Fix overflow when calculating offset with large list ordinals

ghost:
When calculating an offset at [HollowListTypeReadStateShard.java#L38](https://github.com/Netflix/hollow/commit/bc7558c1b4dc84ceb2c415682e6103c11d2f6f5f#diff-7bf58a9ae2833102db141162363e5aa2R38) both operands are `int` and thus can overflow for sufficiently large ordinals:

```java
public int getElementOrdinal(int ordinal, int listIndex) {
    HollowListTypeDataElements currentData;
    int elementOrdinal;
    do {
        long startAndEndElement;
        do {
            currentData = this.currentData;
            long fixedLengthOffset = currentData.bitsPerListPointer * ordinal; // <- overflow
            // ...
        } while(readWasUnsafe(currentData));
        // ...
     } while(readWasUnsafe(currentData));
    return elementOrdinal;
}
```

Kudos to @manuelq for the report in [gitter](https://gitter.im/Netflix/hollow?at=5afadf33f04ce53632dbb526)

PR #223, opened 2018-05-17
fix overflow when calculating offset with large set ordinals

ghost:
Similar to #222 there's an integer overflow that could lead to `ArrayIndexOutOfBoundsException`. This one's more subtle:

```
OVERFLOW: (long)(ordinal * currentData.bitsPerFixedLengthSetPortion) + currentData.bitsPerSetPointer
CORRECT:  ((long)ordinal * currentData.bitsPerFixedLengthSetPortion) + currentData.bitsPerSetPointer,
```

Kudos to @adamkeyser for suggesting a closer look at sets.


ghost:
Redoing with a different branch and more than just Set types.

PR #225, opened 2018-05-18
Refactor magic number -1 to ORDINAL_NONE

ghost:
* [x] move `HollowConstants` into `core`
* [x] where `-1` is semantically an ordinal, replace with `HollowConstants.ORDINAL_NONE`
* [x] fix checkstyle import order failures


akhaku:
👍 you're a hero

PR #227, opened 2018-05-23
avoid integer overflow in hash table size calculation

ghost:
Previously overflowed when `numElements > Integer.MAX_VALUE / 10`.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/HollowConstants.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/HollowConstants.java
@@ -35,4 +35,10 @@ public interface HollowConstants {
      * An ordinal of NULL_ORDINAL signifies "null reference" or "no ordinal"
      */
     int ORDINAL_NONE = -1;
+
+    /**
+     * The maximum number of buckets allowed in a Hollow hash table. Empty space is reserved (based on 70% load factor),
+     * otherwise performance approaches O(n).
+     */
+    int HASH_TABLE_MAX_BUCKETS = (int)((1L << 30) * 7 / 10);
 }
--- a/hollow/src/main/java/com/netflix/hollow/core/memory/encoding/HashCodes.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/memory/encoding/HashCodes.java
@@ -17,15 +17,15 @@
  */
 package com.netflix.hollow.core.memory.encoding;
 
+import static com.netflix.hollow.core.HollowConstants.HASH_TABLE_MAX_BUCKETS;
+
 import com.netflix.hollow.core.memory.ArrayByteData;
 import com.netflix.hollow.core.memory.ByteData;
 import com.netflix.hollow.core.memory.ByteDataBuffer;
 
-
 public class HashCodes {
-
     private static final int MURMURHASH_SEED = 0xeab524b9;
-    
+
     public static int hashCode(ByteDataBuffer data) {
         return hashCode(data.getUnderlyingArray(), 0, (int) data.length());
     }
@@ -168,13 +168,37 @@ public class HashCodes {
         return key;
     }
 
-    public static int hashTableSize(int numElements) {
+    /**
+     * Determine size of hash table capable of storing the specified number of elements with a load
+     * factor applied.
+     *
+     * @param numElements number of elements to be stored in the table
+     * @return size of hash table, always a power of 2
+     * @throws IllegalArgumentException when numElements is negative or exceeds
+     *                                  {@link com.netflix.hollow.core.HollowConstants#HASH_TABLE_MAX_BUCKETS}
+     */
+    public static int hashTableSize(int numElements) throws IllegalArgumentException {
+        if (numElements < 0) {
+            throw new IllegalArgumentException("cannot be negative; numElements="+numElements);
+        } else if (numElements > HASH_TABLE_MAX_BUCKETS) {
+            throw new IllegalArgumentException("exceeds maximum number of buckets; numElements="+numElements);
+        }
+
         if (numElements == 0)
             return 1;
         if (numElements < 3)
             return numElements * 2;
-        int sizeAfterLoadFactor = numElements * 10 / 7;
-        return 1 << (32 - Integer.numberOfLeadingZeros(sizeAfterLoadFactor - 1));
-    }
 
+        // Apply load factor to number of elements and determine next
+        // largest power of 2 that fits in an int
+        int sizeAfterLoadFactor = (int)((long)numElements * 10 / 7);
+        final int size;
+        int bits = 32 - Integer.numberOfLeadingZeros(sizeAfterLoadFactor - 1);
+        if (bits == 31) {
+            size = HASH_TABLE_MAX_BUCKETS;
+        } else {
+            size = 1 << bits;
+        }
+        return size;
+    }
 }
--- a/hollow/src/test/java/com/netflix/hollow/core/util/HashCodesTest.java
+++ b/hollow/src/test/java/com/netflix/hollow/core/util/HashCodesTest.java
@@ -17,12 +17,16 @@
  */
 package com.netflix.hollow.core.util;
 
+import static com.netflix.hollow.core.HollowConstants.HASH_TABLE_MAX_BUCKETS;
+
 import com.netflix.hollow.core.memory.ByteDataBuffer;
 import com.netflix.hollow.core.memory.encoding.HashCodes;
 import com.netflix.hollow.core.memory.encoding.VarInt;
 import com.netflix.hollow.core.memory.pool.WastefulRecycler;
 import java.util.Random;
+import junit.framework.AssertionFailedError;
 import org.junit.Assert;
+import org.junit.Ignore;
 import org.junit.Test;
 
 public class HashCodesTest {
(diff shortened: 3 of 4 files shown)


hollow/src/main/java/com/netflix/hollow/core/memory/encoding/HashCodes.java
```diff
@@ -173,7 +173,7 @@ public static int hashTableSize(int numElements) {
             return 1;
         if (numElements < 3)
             return numElements * 2;
-        int sizeAfterLoadFactor = numElements * 10 / 7;
+        int sizeAfterLoadFactor = (int)((long)numElements * 10 / 7);
```
akhaku: I was going to suggest dividing by 7 first and them multiplying by 10, but this gives you a better spread.

We can still overflow if numElements * 10 / 7 > Integer.MAX_VALUE - how about something like `Math.min(Integer.MAX_VALUE, numElements * 10L / 7)`?

akhaku: Also, TIL (can't wait for java8 with Hollow!) https://stackoverflow.com/questions/1590831/safely-casting-long-to-int-in-java

rpalcolea:
so pulled this branch and released a local snapshot to try out with our dataset.

The fix seems to work :) Think I was going on the right direction but the debugger played with me 😓 

```
typeState = {HollowObjectTypeReadState@6359} 
 unfilteredSchema = {HollowObjectSchema@8357} “MyClass @PrimaryKey(key1, key2)"
 sampler = {HollowObjectSampler@8358} 
 shardNumberMask = 0
 shardOrdinalShift = 0
 shards = {HollowObjectTypeReadStateShard[1]@8359} 
 maxOrdinal = 217435254
```

~217 Million ordinals.

hollow/src/main/java/com/netflix/hollow/core/memory/encoding/HashCodes.java
rpalcolea: Wouldn't `numElements * 10 / 7 > Integer.MAX_VALUE ` be totally out of Hollow's limits? I believe the max was ~500 million, right? if the limit for the number of ordinals is `Integer.MAX_VALUE` which would be the limit of any array in Java, then we might want a safer way to cast the ints? like the one proposed by @akhaku 

`Math.toIntExact(long)` would be phenomenal once Hollow is in Java 8

rpalcolea (approved):

ghost:
Made some more changes to avoid other overflow scenarios and added some checks on `numElements` that I think are sane (i.e. would fail in other, less clear ways later on). Some questions for @dkoszewnik inline...

hollow/src/main/java/com/netflix/hollow/core/memory/encoding/HashCodes.java
```diff
@@ -168,13 +168,40 @@ public static int hashInt(int key) {
         return key;
     }
 
-    public static int hashTableSize(int numElements) {
+    /**
+     * Determine size of hash table capable of storing the specified number of elements with a load
+     * factor applied.
+     *
+     * @param numElements number of elements to be stored in the table
+     * @return size of hash table, always a power of 2
```
ghost, line 176: @dkoszewnik is power-of-two a hard requirement or just a desirable trait?

hollow/src/main/java/com/netflix/hollow/core/memory/encoding/HashCodes.java
```diff
@@ -168,13 +168,40 @@ public static int hashInt(int key) {
...
+     * @throws IllegalArgumentException when numElements is negative
+     */
+    public static int hashTableSize(int numElements) throws IllegalArgumentException {
+        if (numElements < 0) {
```
ghost: This is always an error. If `numElements` ever overflows this will surface it sooner and with a little more meaningful failure.

hollow/src/main/java/com/netflix/hollow/core/memory/encoding/HashCodes.java
```diff
@@ -168,13 +168,40 @@ public static int hashInt(int key) {
...
+            throw new IllegalArgumentException("cannot be negative; numElements="+numElements);
+        } else if (numElements > HASH_TABLE_MAX_SIZE) {
```
ghost: I *think* it's safe to always treat this as an error. For example, `HollowPrimaryKeyIndex` uses this to construct a `FixedLengthElementArray` for an open-addressing hash table and we cannot store more elements in it than the maximum hash table size.

hollow/src/main/java/com/netflix/hollow/core/memory/encoding/HashCodes.java
```diff
@@ -168,13 +168,40 @@ public static int hashInt(int key) {
...
             return 1;
         if (numElements < 3)
             return numElements * 2;
-        int sizeAfterLoadFactor = numElements * 10 / 7;
-        return 1 << (32 - Integer.numberOfLeadingZeros(sizeAfterLoadFactor - 1));
-    }
 
+        // Apply load factor to number of elements and determine next
+        // largest power of 2 that fits in an int
+        int sizeAfterLoadFactor = (int)((long)numElements * 10 / 7);
```
ghost, line 194: Avoid overflowing `numElements * 10`. Allow `sizeAfterLoadFactor` to overflow because we explicitly check for overflow a few lines down.

hollow/src/main/java/com/netflix/hollow/core/memory/encoding/HashCodes.java
```diff
@@ -168,13 +168,40 @@ public static int hashInt(int key) {
...
+        final int size;
+        if (sizeAfterLoadFactor < 0) { // overflow
```
ghost: I tried a variant with `sizeAfterLoadFactor` as a `long` and then testing whether it *would* overflow. This version where we just check to see if it did overflow ended up looking cleaner to me.

hollow/src/main/java/com/netflix/hollow/core/memory/encoding/HashCodes.java
```diff
@@ -168,13 +168,40 @@ public static int hashInt(int key) {
...
+            size = HASH_TABLE_MAX_SIZE;
+        } else {
+            int bits = 32 - Integer.numberOfLeadingZeros(sizeAfterLoadFactor - 1);
+            if (bits == 31) {
```
ghost: Previously, with a large enough `sizeAfterLoadFactor` the `1 << bits` would also overflow. Now we clamp at `HASH_TABLE_MAX_SIZE` i.e `1 << 30`

hollow/src/test/java/com/netflix/hollow/core/util/HashCodesTest.java
```diff
@@ -41,7 +45,91 @@ public void testStringHashCode() {
...
+        Assert.assertEquals(1 << 29, HashCodes.hashTableSize(N));
+        Assert.assertEquals(1 << 29, HashCodes.hashTableSize(N + 1));
+
+        // first quantity that requires hash table size at maximum
+        N = 375_809_640;
+        Assert.assertEquals(1 << 29, HashCodes.hashTableSize(N - 1));
+        Assert.assertEquals(HASH_TABLE_MAX_SIZE, HashCodes.hashTableSize(N));
+
+        // first quantity (after load factor) that exhausts maximum hash table size
+        N = 751_619_278;
```
ghost: I have a TODO to add a "Known Limitations" section in [hollow.how](http://hollow.how). This number will factor into a discussion on the upper limits of cardinality (if I understand correctly) of primary keys, sets, and maps.

hollow/src/test/java/com/netflix/hollow/core/util/HashCodesTest.java
```diff
@@ -41,7 +45,91 @@ public void testStringHashCode() {
...
+        N = HASH_TABLE_MAX_SIZE;
+        Assert.assertEquals(HASH_TABLE_MAX_SIZE, HashCodes.hashTableSize(N));
+        try {
+            HashCodes.hashTableSize(N + 1);
+            Assert.fail("exception expected");
+        } catch (IllegalArgumentException ex) {
+            Assert.assertEquals("exceeds max hash table size; numElements=1073741825", ex.getMessage());
+        }
+
+        // second integer overflow boundary condition (1_503_238_553)
```
ghost: Strictly speaking the remaining assertions are redundant as the remaining values for `N` exceed `HASH_TABLE_MAX_SIZE`. I kept them for two reasons:

1. if we remove the `IllegalArgumentException` then these are integer overflow boundary conditions we want to test
2. I decided on throwing `IllegalArgumentException` after figuring out these boundary conditions and want that at least captured in git history for posterity even if we simplify this test later

ghost:
@rpalcolea 

> ```
> typeState = {HollowObjectTypeReadState@6359} 
> unfilteredSchema = {HollowObjectSchema@8357} “MyClass @PrimaryKey(key1, key2)"
> sampler = {HollowObjectSampler@8358} 
> shardNumberMask = 0
> shardOrdinalShift = 0
> shards = {HollowObjectTypeReadStateShard[1]@8359} 
> maxOrdinal = 217435254
> ```
> 
> ~217 Million ordinals.

Approx 217M lines up with the boundary conditions we uncovered. The first was the `N * 10` overflow that I fixed first and you tested. The next boundary condition that I hadn't fixed at the time of your test occurs at `~375M`. Anyhow, that should be also be fixed with the latest changes I pushed to the PR.

The hard limit of `~1073M` unique records in a primary key index should now *theoretically* be reachable. Practically speaking, performance of the underlying hash table will start degrading substantially prior to that. The *practical* limit should be `~750M` with good performance (if our hashing is good).

hollow/src/test/java/com/netflix/hollow/core/util/HashCodesTest.java
rpalcolea: great idea... the "known limitations" section will be a nice addition to the docs

hollow/src/main/java/com/netflix/hollow/core/memory/encoding/HashCodes.java
rpalcolea: nice! `IllegalArgumentException` will help a lot while trying to determine issues caused by a huge dataset

rpalcolea:
@toolbear,

It's good to know that we could potentially hit the ~750M. We think that eventually we will need to do some sharding on our side and start distributing datasets from multiple producers... but we are not there yet.

We were not expecting to hit the ~230M (as of 5 minutes ago).

Thanks for looking at this!

hollow/src/main/java/com/netflix/hollow/core/HollowConstants.java
```diff
@@ -35,4 +35,9 @@
      * An ordinal of NULL_ORDINAL signifies "null reference" or "no ordinal"
      */
     int ORDINAL_NONE = -1;
+
+    /**
+     * The maximum size of Hollow hash table; largest positive power of 2 that fits in an int
+     */
+    int HASH_TABLE_MAX_SIZE = 1 << 30;
```
dkoszewnik: This should be the maximum number of buckets, not the maximum size.  Some empty space should be left over, otherwise performance will degrade significantly (approaching O(n) insertion and lookup).  The default selected was 70% populated, which seems to work well -- I would recommend a maximum size of `(1L << 30) * 7 / 10`. 

As a bonus, selecting this value would allow you to omit the overflow check -- values which would otherwise overflow would instead now throw your `IllegalArgumentException`

hollow/src/main/java/com/netflix/hollow/core/memory/encoding/HashCodes.java
dkoszewnik: It's a hard requirement _unless_ you change some of the math.  See, for example:
https://github.com/Netflix/hollow/blob/master/hollow/src/main/java/com/netflix/hollow/core/read/engine/set/HollowSetTypeReadStateShard.java#L64

> `long bucket = startBucket + (hashCode & (endBucket - startBucket - 1))`

Here we're taking the mod of the hash code using (_the number of buckets_ - 1).  This bitmask only works because the number of buckets is a power of 2.

As a side note -- if you don't use a power of two, then the mod operation will be significantly more expensive.

PR #228, opened 2018-05-25
performance

joshgord:
A few changes that should make the job of the compiler easier by removing inline depth, avoiding casting, adding immutability to the proxy classes and marking mutable state as volatile.


joshgord:
Checking on failing tests....

PR #229, opened 2018-05-28
Allow sampling of stale references to be configurable

joshgord:
Motivation is to make the sampling on historical data configurable in line with the other sampling mechanisms


ghost (approved):

ghost:
TravisCI build is failing because the build script only clones 50 commits back and this change was submitted a while back...or perhaps because of me doing the conflict resolution using GitHub web UI. Either way, the build when the PR was submitted passed, so going to merge. Either `master` builds fine or I'll fix the build afterwards.

PR #234, opened 2018-06-06
Correctly escaping all special characters in String.

nkostiv:
Special characters like \t, \b, etc were not escaped during creating JSON from records and this case was not covered with tests. Fixed escaping and added junit.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/tools/stringifier/HollowRecordJsonStringifier.java
+++ b/hollow/src/main/java/com/netflix/hollow/tools/stringifier/HollowRecordJsonStringifier.java
@@ -36,7 +36,9 @@ import java.io.StringWriter;
 import java.io.Writer;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Set;
 
 /**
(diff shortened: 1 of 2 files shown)


hollow/src/main/java/com/netflix/hollow/tools/stringifier/HollowRecordJsonStringifier.java
```diff
@@ -384,9 +398,12 @@ private void appendFieldStringify(Writer writer, HollowDataAccess dataAccess, in
     }
 
     private String escapeString(String str) {
-        if(str.indexOf('\\') == -1 && str.indexOf('\"') == -1)
-            return str;
-        return str.replace("\\", "\\\\").replace("\"", "\\\"");
+        for (Character escapeSequence : escapedSymbols.keySet()) {
```
ghost, line 401: I'm a little concerned about the efficiency here. It wasn't great before this change, but now it's 7x worse.

The `#indexOf` should be O(n) given that the pattern to match is 1 char always. But we do it 7 times now.

The `str = str.replace(escapeSequence.toString(), escapedSymbols.get(escapeSequence));` will generate 7 objects to GC. I wonder if we can use `StringBuilder` to handle this more efficiently? If we're able to calculate or estimate the replacements beforehand we can construct one with an initial size that isn't likely to need to grow partway through building it.

For simple, relatively shallow objects this is negligible. But we're already encountering data models with some large object graphs for which this stringify already performs pretty badly.

hollow/src/main/java/com/netflix/hollow/tools/stringifier/HollowRecordJsonStringifier.java
```diff
@@ -49,6 +51,18 @@
     private final boolean collapseAllSingleFieldObjects;
     private final boolean prettyPrint;
 
+    private Map<Character, String> escapedSymbols = Collections.unmodifiableMap(
```
ghost, line 54: Two parallel arrays might be more efficient. The map is definitely more readable, but again we might be in an area where some slightly less readable optimizations are appropriate.

PR #235, opened 2018-06-07
Fix JDK8 support in subprojects

rpalcolea:
Looks like https://github.com/Netflix/hollow/blob/master/hollow/build.gradle#L1 overrides the `sourceCompatibility` from 1.8 to 1.7.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,8 +13,10 @@ subprojects {
   apply plugin: 'nebula.netflixoss'
   apply plugin: 'checkstyle'
 
-  sourceCompatibility = 1.8
-  targetCompatibility = 1.8
+  tasks.withType(JavaCompile) {
+    sourceCompatibility = 1.8
+    targetCompatibility = 1.8
+  }
 
   group = 'com.netflix.hollow'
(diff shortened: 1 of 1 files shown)


rpalcolea:
Proposal for java module here -> https://github.com/Netflix/hollow/pull/236

build.gradle
```diff
@@ -13,8 +13,10 @@ subprojects {
   apply plugin: 'nebula.netflixoss'
   apply plugin: 'checkstyle'
 
-  sourceCompatibility = 1.8
-  targetCompatibility = 1.8
+  tasks.withType(JavaCompile) {
```
ghost, line 16: This change was essentially what our gradle experts suggested. Here's the exact idiom they recommend:

```groovy
subprojects {
  tasks.withType(JavaCompile) {
    sourceCompatibility = JavaVersion.VERSION_1_8
    targetCompatibility = JavaVersion.VERSION_1_8
  }
}
```

rpalcolea: Modified the code to use `JavaVersion.VERSION_1_8`

PR #238, opened 2018-06-12
Protecting against NPE when getStateEngine() is called

jatin146:
Returning null instead of NPE when calling getStateEngine()

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
@@ -185,7 +185,10 @@ public class HollowClientUpdater {
     }
 
     public HollowReadStateEngine getStateEngine() {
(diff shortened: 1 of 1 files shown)


hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
```diff
@@ -185,7 +185,10 @@ public StackTraceRecorder getStaleReferenceUsageStackTraceRecorder() {
     }
 
     public HollowReadStateEngine getStateEngine() {
-        return hollowDataHolder.getStateEngine();
+        if(hollowDataHolder != null) {
```
akhaku, line 188: minor - oneline this?
```
return hollowDataHolder == null ? null : hollowDataHolder.getStateEngine();

ghost: Doh. I was too quick on the merge button.

PR #239, opened 2018-06-18
Upgrade to Gradle 4.9 and modify build scripts to use java-library

rpalcolea:
This upgrades to Gradle 4.8 and also introduces `java-library` plugin. 

From the gradle docs (https://docs.gradle.org/current/userguide/java_library_plugin.html):

> The key difference between the standard Java plugin and the Java Library plugin is that the latter introduces the concept of an API exposed to consumers. A library is a Java component meant to be consumed by other components. It’s a very common use case in multi-project builds, but also as soon as you have external dependencies.

> The plugin exposes two configurations that can be used to declare dependencies: `api` and `implementation`. The `api` configuration should be used to declare dependencies which are exported by the library API, whereas the `implementation` configuration should be used to declare dependencies which are internal to the component.


rpalcolea:
Hi @toolbear 

This upgrades to gradle 4.8.1 too

rpalcolea:
This is ready for review again @toolbear

PR #240, opened 2018-06-26
Preserve encoding for compact Strings

zielu:
Extension of compact String handling from PR #221 to handle variable encodings possible (UTF16 and LATIN1). 

Please note that solution differs from original code (#221 and before it) in control flow when object passed to mapper in copy method is an instance of String - there is no copy created (getStringFromField method) but original String is returned.
If there is a specific reason that forces us to create such a copy then we could do  `new String((String) obj)`


ghost:
> If there is a specific reason that forces us to create such a copy then we could do new String((String) obj)

There's no reason to create a copy. This looks good as-is.

PR #242, opened 2018-07-03
Override default for 'sourceCompatibility' from 'nebula.netflixoss', update latest stable version to '3.0.1'

blekit:
I've moved setting of values for `sourceCompatibility` and `targetCompatibility` to subprojects section to override the default setting from `nebula.netflixoss` plugin. Without it, after checking out project failed to build with following error:

```
A problem occurred configuring project ':hollow'.
> Could not locate a compatible JDK for target compatibility 1.7. Change the source/target compatibility, set a JDK_17 environment variable with the location, or install to one of the default search locations
```

I've also noticed that latest stable version hasn't been updated for quite some time so I'd like to change that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Binaries are available from Maven Central and jcenter.
 
 |GroupID/Org|ArtifactID/Name|Latest Stable Version|
 |-----------|---------------|---------------------|
-|com.netflix.hollow|hollow|2.16.4|
+|com.netflix.hollow|hollow|3.0.1|
 
 In a Maven `.pom` file:
 
@@ -29,14 +29,14 @@ In a Maven `.pom` file:
         <dependency>
                 <groupId>com.netflix.hollow</groupId>
                 <artifactId>hollow</artifactId>
-                <version>2.16.4</version>
+                <version>3.0.1</version>
         </dependency>
         ...
 
 In a Gradle `build.gradle` file:
 
         ...
-        compile 'com.netflix.hollow:hollow:2.16.4'
+        compile 'com.netflix.hollow:hollow:3.0.1'
         ...
         
 ## Get Support
--- a/build.gradle
+++ b/build.gradle
@@ -6,17 +6,12 @@ plugins {
   id 'nebula.netflixoss' version '5.0.0'
 }
 
-sourceCompatibility = 1.8
-targetCompatibility = 1.8
-
 subprojects {
   apply plugin: 'nebula.netflixoss'
   apply plugin: 'checkstyle'
 
-  tasks.withType(JavaCompile) {
-    sourceCompatibility = JavaVersion.VERSION_1_8
-    targetCompatibility = JavaVersion.VERSION_1_8
-  }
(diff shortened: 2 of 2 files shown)


rpalcolea:
Hi @blekit ,

How is this different from https://github.com/Netflix/hollow/blob/master/build.gradle#L16?

I've been using Hollow with Java 8 since that change without issues. Wonder if that's more related to your environment?

blekit:
Hi @rpalcolea,

I'm not an expert on Gradle, but it seems to me that currently for subprojects you only override the source and target compatibility settings in tasks that extend JavaCompile. At the same time you apply in those projects plugin `nebula.netflixoss` in which the default value of `sourceCompatibility` is 1.7 and for some reason it took precedence. I don't have the JDK 1.7 on my machine and when I tried to build the project for the first time I got error described in the first post. I thought that since you don't support Java 7 anymore, it shouldn't be required to build the project and that's why I created this pull request. Perhaps it is indeed related to my environment, but the problem was resolved by change I'm proposing. 

Cheers,
Michal

rpalcolea:
Oh I see what you are saying now @blekit 

I was able to reproduce this... Here is a build scan with the full error log https://scans.gradle.com/s/pc6uj4jwppcqg/failure?openFailures=WzBd&openStackTraces=WzEse31d#top=0

And yes seems like an issue with nebula-plugin-plugin which is the base plugin for all nebula plugins

AFAIK, the nebula-plugin-plugin sets the `targetCompatibility` and `sourceCompatibility` to 1.7 in https://github.com/nebula-plugins/nebula-plugin-plugin/blob/master/src/main/groovy/nebula/plugin/plugin/NebulaPluginPlugin.groovy#L62

And this plugin gets applied to the `gradle-java-cross-compile-plugin` which can not find the JDK 1.7 (https://github.com/nebula-plugins/gradle-java-cross-compile-plugin/blob/master/src/main/kotlin/nebula/plugin/compile/JavaCrossCompilePlugin.kt#L84) and fails.

I think the best thing to do for now is keep your change but also remove the 

```
tasks.withType(JavaCompile) {
     sourceCompatibility = JavaVersion.VERSION_1_8	     sourceCompatibility = JavaVersion.VERSION_1_8
     targetCompatibility = JavaVersion.VERSION_1_8	     targetCompatibility = JavaVersion.VERSION_1_8
}
``` 

block

I would say that this is expected since nebula requires 1.7... so setting the target for all tasks should be good.

rpalcolea:
heads up @toolbear @akhaku ^^

blekit:
@rpalcolea should I apply the change you suggest in my fork or are you going to do it instead?

rpalcolea:
Can you do it In your fork... I don’t have writing privileges in the repo 😀

blekit:
sure, on it :) 
... and done

rpalcolea:
sweet, we will have to wait for someone from Netflix to merge this! good find!

ghost:
Reproduced on our end, too. Merging.

Issue #243, opened 2018-07-05
AbstractHollowOrdinalIterable 's iterator() method is not idempotent


blekit:
Hi,

during development of our sollution backed by Hollow, we've noticed that `iterator()` method in `AbstractHollowOrdinalIterable` class is not idempotent and invoking it might corrupt wrapped `HollowOrdinalIterator`.

Each time `iterator` in `AbstractHollowOrdinalIterable` is called it creates a new instance of `Iterator` interface and during its initialization invokes `next()` method of underlying `HollowOrdinalIterator`. As a result, we've observed that some of the ordinals might be omitted during later processing if `iterator()` was called more than once (directly or indirectly - e.g. by calling `spliterator` method).

Cheers!

rpalcolea:
I think this issue could be closed now

rpalcolea:
@blekit ^^

blekit:
Yeah, it definitely can.

PR #244, opened 2018-07-06
Fix AbstractHollowOrdinalIterable - make iterator() method idempotent

blekit:
Proposed fix for issue #243 

When constructing `AbstractHollowOrdinalIterable` we store the first ordinal from wrapped `HollowOrdinalIterator` in instance field. This value is then used to initialize each created `Iterator` instance. Now the method should be idempotent.


blekit:
Sorry, I must've messed something with my fork and now it shows too many commits in this Pull Request... The only relevant is the last one.

ghost:
Thanks for the fix!

PR #245, opened 2018-07-06
Support for stringify iterator of HollowRecords

rpalcolea:
Hi @toolbear,

We have a use case where we search for `HollowObjects` using a `HashIndex`, for example `BookAPIHashIndex` where the model is:

```Book.java
@HollowPrimaryKey(fields={"bookId"})
public class Book {
    long bookId;
    String title;
    String publisher;
}
```

The `HashIndex` returns a `Iterable<Book>` and we wanted to Stringify this right away as a JSON to serve it via HTTP server.

The current implementation for `HollowStringifier.java` doesn't support a method that takes a `Writer` and `Iterable<HollowObject>` as arguments and stringify them properly.

This change includes a default method which does nothing for `HollowStringifier` in case Hollow users are implementing their own stringifiers and also default implementation for `HollowRecordJsonStringifier` and `HollowRecordStringifier`

Please let me know your thoughts or if you think there is a better way to accomplish this

I'm also proposing adding re-usable test types (POJOs) to `hollow-test` project so they could be shared across multiple tests and potentially simplify some tests. Please look at `TestTypeA` and let me know what you think of this approach

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/tools/stringifier/HollowStringifier.java
+++ b/hollow/src/main/java/com/netflix/hollow/tools/stringifier/HollowStringifier.java
@@ -49,7 +49,9 @@ public interface HollowStringifier<T extends HollowStringifier> {
      *
      * @throws IOException thrown if there is an error writing to the Writer
      */
-    public void stringify(Writer writer,Iterable<HollowRecord> records) throws IOException;
+    public default void stringify(Writer writer,Iterable<HollowRecord> records) throws IOException {
+        //Do nothing
+    }
 
     /**
      * Create a String representation of the record in the provided dataset, of the given type, with the specified ordinal.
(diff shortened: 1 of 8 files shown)


hollow/src/main/java/com/netflix/hollow/tools/stringifier/HollowStringifier.java
```diff
@@ -44,6 +44,15 @@
      */
     public void stringify(Writer writer, HollowRecord record) throws IOException;
 
+    /**
+     * Writes a String representation of the specified collection of {@link HollowRecord} to the provided Writer.
+     *
+     * @throws IOException thrown if there is an error writing to the Writer
+     */
+    public default void stringify(Writer writer,Iterable<HollowRecord> records) throws IOException {
+        //Do nothing
```
ghost: Maybe `throw new UnsupportedOperationException("not implemented")`? My gut tells me that's a default that would surface errors sooner.

ghost (changes requested):
Requesting changes, but I'm open to being convinced that a no-op default method is better.

hollow/src/main/java/com/netflix/hollow/tools/stringifier/HollowStringifier.java
rpalcolea: hmmm... actually, I think it makes sense to throw the exception. That gives more visibility and could prevent silent errors in the future. Changed in https://github.com/Netflix/hollow/pull/245/commits/14a501eb7e9ecc99f522064b7f76972958afcb8b

PR #250, opened 2018-08-06
Allow creating HollowConsumers for testing purposes

akhaku:
Add TestHollowConsumer, TestBlobRetriever, and
TestAnnouncementWatcher to facilitate creating
HollowConsumer objects for use in unit tests.
Generally speaking, a TestHollowConsumer is constructed
with a TestAnnouncementWatcher and TestBlobRetriever, and
can be manipulated load a snapshot as represented by a
HollowWriteStateEngine. Support for loading deltas will come
later.
As part of this effort, refactor
HollowReadStateEngineBuilder to extract a
HollowWriteStateEngineBuilder.

### DIFF
--- a/hollow-test/src/main/java/com/netflix/hollow/test/consumer/TestHollowConsumer.java
+++ b/hollow-test/src/main/java/com/netflix/hollow/test/consumer/TestHollowConsumer.java
@@ -1,0 +1,107 @@
+/*
+ *
+ *  Copyright 2018 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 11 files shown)


akhaku:
The javadoc on TestHollowConsumer has a usage example, and there are more in the unit tests, in TestHollowConsumerTest. I'll add info to the testing docs if/when this is merged and released.

hollow-test/src/main/java/com/netflix/hollow/test/consumer/TestHollowConsumer.java
```diff
@@ -0,0 +1,107 @@
...
+import java.io.IOException;
+import java.util.List;
+import java.util.concurrent.Executor;
+
+/**
+ * This class allows constructing HollowConsumer objects for use in unit and integration tests.
+ * This uses an in-memory blob store, which means that all state is kept in memory, and so this is
+ * not suited for use in normal application usage, just in tests.
+ *
+ * Example usage:
```
ghost, line 39: This still seems like a lot of boilerplate test fixture setup, but it is still a good step forward.

ghost (approved):
Looks really good.

Issue #253, opened 2018-08-21
HollowObjectMapper fails with unhelpful NPE if writing a Map with a null value


akhaku:
```
Caused by: java.lang.NullPointerException
    at com.netflix.hollow.core.write.objectmapper.HollowObjectTypeMapper.write(HollowObjectTypeMapper.java:170) ~[hollow-3.1.0.jar:3.1.0]
    at com.netflix.hollow.core.write.objectmapper.HollowMapTypeMapper.write(HollowMapTypeMapper.java:76) ~[hollow-3.1.0.jar:3.1.0]
    at com.netflix.hollow.core.write.objectmapper.HollowObjectTypeMapper$MappedField.copy(HollowObjectTypeMapper.java:470) ~[hollow-3.1.0.jar:3.1.0]
    at com.netflix.hollow.core.write.objectmapper.HollowObjectTypeMapper.write(HollowObjectTypeMapper.java:176) ~[hollow-3.1.0.jar:3.1.0]
    at com.netflix.hollow.core.write.objectmapper.HollowListTypeMapper.write(HollowListTypeMapper.java:79) ~[hollow-3.1.0.jar:3.1.0]
    at com.netflix.hollow.core.write.objectmapper.HollowObjectTypeMapper$MappedField.copy(HollowObjectTypeMapper.java:470) ~[hollow-3.1.0.jar:3.1.0]
    at com.netflix.hollow.core.write.objectmapper.HollowObjectTypeMapper.write(HollowObjectTypeMapper.java:176) ~[hollow-3.1.0.jar:3.1.0]
    at com.netflix.hollow.core.write.objectmapper.HollowObjectMapper.add(HollowObjectMapper.java:70) ~[hollow-3.1.0.jar:3.1.0]
    at com.netflix.hollow.api.producer.WriteStateImpl.add(WriteStateImpl.java:41) ~[hollow-3.1.0.jar:3.1.0]
```

We should add a test to reproduce this, and either:
- fail with a more useful error message, or
- make it work

PaulSandoz:
Hollow does not support `null` in collections (see Hollow [documentation](https://hollow.how/advanced-topics/#in-memory-data-layout))). Exceptions should be produced closer to the source elements and with better messages.

PR #254, opened 2018-08-28
Better NPE reported. Fix for issue #253.

PaulSandoz:
Throw explicit NPEs for null values in collections and maps (null values
are not supported)

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/HollowListTypeMapper.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/HollowListTypeMapper.java
@@ -26,10 +26,12 @@ import com.netflix.hollow.core.write.HollowWriteRecord;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.List;
-import java.util.Objects;
 import java.util.Set;
 
 public class HollowListTypeMapper extends HollowTypeMapper {
+
+    private static final String NULL_ELEMENT_MESSAGE =
+            "Null element contained in instance of a List with schema \"%s\". Lists cannot contain null elements";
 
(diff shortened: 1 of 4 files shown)


hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/HollowListTypeMapper.java
```diff
@@ -68,6 +69,7 @@ public int write(Object obj) {
         if(ignoreListOrdering) {
             IntList ordinalList = getIntList();
             for(Object o : l) {
+                Objects.requireNonNull(o, "Null element. Lists cannot contain null elements");
```
akhaku: Nice. Is it possible to add the list type name to the message? Similar comments for the map below - might help track down which list/map has null elements or keys/values

hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/HollowListTypeMapper.java
```diff
@@ -69,7 +72,7 @@ public int write(Object obj) {
         if(ignoreListOrdering) {
             IntList ordinalList = getIntList();
             for(Object o : l) {
-                Objects.requireNonNull(o, "Null element. Lists cannot contain null elements");
+                Objects.requireNonNull(o, String.format(NULL_ELEMENT_MESSAGE, schema));
```
akhaku: Looks good, although I wonder if it's worth moving the String.format outside of the loop - no need to create that string once per list element.

PaulSandoz: Doh! good point. Will fix.

ghost (approved):

PR #257, opened 2018-09-05
Replace Throwable.printStackTrace with SEVERE log

zielu:
In order to improve integration with logging frameworks, printStackTrace() calls are replaced with logging.
Reason for this change is that we loose information in logs about for example OutOfMemory exceptions because they are caught and printed to output.

There is one more change in HollowClientUpdater related to this: to update failure metric before calling RefreshListeners. Otherwise in case something goes wrong inside listener metric will not be updated.


ghost (approved):
Looks good. Thank you for the contribution.

PR #260, opened 2018-09-14
Fix NPE when primary key index tracked by history is removed.

PaulSandoz:
This can occur when a delta occurs after a snapshot that removed instances of a primary key but the key is still in the primary key index map.

### DIFF
--- a/hollow/src/test/java/com/netflix/hollow/tools/history/HollowHistoryTest.java
+++ b/hollow/src/test/java/com/netflix/hollow/tools/history/HollowHistoryTest.java
@@ -224,6 +224,83 @@ public class HollowHistoryTest extends AbstractStateEngineTest {
         }
     }
 
+    @Test
+    public void testAddRemoveTypeThenDelta() throws IOException {
+        addRecord(1, 2, 3);
+        addRecord(2, 3, 4);
+        roundTripSnapshot();
+
+        long version = 1;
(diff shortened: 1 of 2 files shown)


hollow/src/test/java/com/netflix/hollow/tools/history/HollowHistoryTest.java
```diff
@@ -224,6 +224,83 @@ public void testRemoveType() throws IOException {
...
+            addRecord(2, 3, 4);
+
+            roundTripSnapshot();
+            setupKeyIndex(readStateEngine, history);
+            history.doubleSnapshotOccurred(readStateEngine, 5);
+
+            assertRecord(retrieveAddedRecord(history, 2L, B_TYPE, 1), B_FN_PREFIX, 1, 2);
+            assertRecord(retrieveAddedRecord(history, 2L, B_TYPE, 2), B_FN_PREFIX, 2, 3);
+            assertRecord(retrieveRemovedRecord(history, 3, B_TYPE, 1), B_FN_PREFIX, 1, 2);
+            assertRecord(retrieveRemovedRecord(history, 3, B_TYPE, 2), B_FN_PREFIX, 2, 3);
```
akhaku, line 300: nice work!

akhaku (approved):

PR #262, opened 2018-09-16
Add an object modification validator

akhaku:
This validator takes in a filter function, which it calls with the "old" object and the "new" object, for objects that are modified in a cycle.
Also minor cleanup - fix typo in method name, and whitespace cleanup in HollowProducer.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/validation/ObjectModificationValidator.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/validation/ObjectModificationValidator.java
@@ -1,0 +1,140 @@
+/*
+ *
+ *  Copyright 2018 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.hollow.api.producer.validation;
+
+import com.netflix.hollow.api.custom.HollowAPI;
+import com.netflix.hollow.api.objects.HollowObject;
+import com.netflix.hollow.api.producer.HollowProducer.ReadState;
+import com.netflix.hollow.api.producer.HollowProducer.Validator;
+import com.netflix.hollow.core.HollowConstants;
+import com.netflix.hollow.core.index.HollowPrimaryKeyIndex;
+import com.netflix.hollow.core.index.key.PrimaryKey;
+import com.netflix.hollow.core.read.dataaccess.HollowDataAccess;
+import com.netflix.hollow.core.read.engine.HollowTypeReadState;
+import com.netflix.hollow.core.read.engine.object.HollowObjectTypeReadState;
+import com.netflix.hollow.core.schema.HollowSchema.SchemaType;
+import java.util.Arrays;
+import java.util.BitSet;
+import java.util.function.BiFunction;
+import java.util.function.BiPredicate;
+import java.util.function.Function;
+
+/**
+ * This validator runs a provided {@link BiPredicate} on every instance of an object that was
+ * modified in a cycle. This "filter" function is called with the old object as the first argument
+ * and the new object as the second argument. If the function returns false, this validator is
+ * marked as failed and the producer cycle fails.
+ * Note that the filter function is not called with any objects that were added or removed.
+ * Also note that this validator can only be used on types that have a primary key.
+ * Example construction - note that {@code Movie} and {@code MovieAPI} are generated classes:
+ * <code>
+ *     new ObjectModificationValidator&lt;MovieAPI, Movie&gt;(
+ *         "Movie",
+ *         (oldMovie, newMovie) -&gt; oldMovie.getName().equals(newMovie.getName()),
+ *         MovieAPI::new,
+ *         (api, ordinal) -&gt; api.getMovie(ordinal));
+ * </code>
+ */
+public class ObjectModificationValidator<A extends HollowAPI, T extends HollowObject>
+        implements Validator {
+    private final String typeName;
+    private final Function<HollowDataAccess, A> apiFunction;
+    private final BiFunction<A, Integer, T> hollowObjectFunction;
+    private final BiPredicate<T, T> filter;
+
+    private SingleValidationStatus.SingleValidationStatusBuilder statusBuilder;
+
+    @SuppressWarnings("WeakerAccess")
+    public ObjectModificationValidator(String typeName, BiPredicate<T, T> filter,
+            Function<HollowDataAccess, A> apiFunction,
+            BiFunction<A, Integer, T> hollowObjectFunction) {
+		this.typeName = typeName;
+		this.filter = filter;
+		this.apiFunction = apiFunction;
+		this.hollowObjectFunction = hollowObjectFunction;
+        this.statusBuilder = SingleValidationStatus.builder(ObjectModificationValidator.class.getSimpleName())
+                .addAdditionalInfo("typeName", typeName);
+    }
+
+    @Override
+	public void validate(ReadState readState) throws ValidationException {
+        try {
+            HollowTypeReadState typeState = readState.getStateEngine().getTypeState(typeName);
+            if (typeState == null) {
+                throw new ValidationException("Cannot execute ObjectModificationValidator on missing type "
+                        + typeName);
+            } else if (typeState.getSchema().getSchemaType() != SchemaType.OBJECT) {
+                throw new ValidationException("Cannot execute ObjectModificationValidator on type " + typeName
+                        + " because it is not a HollowObjectSchema - it is a "
+                        + typeState.getSchema().getSchemaType());
+            }
+            BitSet latestOrdinals = typeState.getPopulatedOrdinals();
+            BitSet previousOrdinals = typeState.getPreviousOrdinals();
+            if (previousOrdinals.isEmpty()) {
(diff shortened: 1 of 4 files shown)


hollow/src/main/java/com/netflix/hollow/api/producer/validation/ObjectModificationValidator.java
```diff
@@ -0,0 +1,122 @@
...
+import com.netflix.hollow.core.index.key.PrimaryKey;
+import com.netflix.hollow.core.read.engine.HollowTypeReadState;
+import com.netflix.hollow.core.read.engine.object.HollowObjectTypeReadState;
+import com.netflix.hollow.core.schema.HollowSchema.SchemaType;
+import java.util.Arrays;
+import java.util.BitSet;
+import java.util.function.BiFunction;
+
+/**
+ * This validator runs a provided {@link BiFunction} on every instance of an object that was
```
PaulSandoz: You can use a BiPredicate instead (https://docs.oracle.com/javase/8/docs/api/java/util/function/BiPredicate.html).

"
A filtering validator that tests each object that was modified in a cycle. The test {@link BiPredicate function} accepts the prior object and the current object. If the test fails (returns false) then validation fails immediately and any subsequent modified objects will not be tested.
"

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ObjectModificationValidator.java
```diff
@@ -0,0 +1,122 @@
...
+ * Also note that this validator can only be used on types that have a primary key.
+ */
+public class ObjectModificationValidator<T extends HollowObject> implements Validator {
+    private final String typeName;
+    private final BiFunction<T, T, Boolean> filter;
+
+    private SingleValidationStatus.SingleValidationStatusBuilder statusBuilder;
+
+    @SuppressWarnings("WeakerAccess")
+    public ObjectModificationValidator(String typeName, BiFunction<T, T, Boolean> filter) {
```
PaulSandoz: `BiPredicate<? super T, ? super T> filter`, the function is consuming instances of T so can accept T or a super type.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ObjectModificationValidator.java
```diff
@@ -0,0 +1,122 @@
...
+            } else if (typeState.getSchema().getSchemaType() != SchemaType.OBJECT) {
+                throw new ValidationException("Cannot execute ObjectModificationValidator on type " + typeName
+                        + " because it is not a HollowObjectSchema - it is a "
+                        + typeState.getSchema().getSchemaType());
+            }
+            BitSet latestOrdinals = typeState.getPopulatedOrdinals();
+            BitSet previousOrdinals = typeState.getPreviousOrdinals();
+            if (previousOrdinals.isEmpty()) {
+                statusBuilder.withMessage("Skipping " + ObjectModificationValidator.class.getSimpleName()
+                        + " because previous cycle has no " + typeName).withStatus(Status.SKIP);
```
PaulSandoz: Arguably the validator is not skipping here and below when the ordinals are empty. There is nothing to validate so no work to do, where as skipping sort of implies "i could do work but i am choosing not to". IMHO the notion of skipping should be reconsidered.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ObjectModificationValidator.java
```diff
@@ -0,0 +1,122 @@
...
+        SingleValidationStatus status = statusBuilder.build();
+        return getClass().getSimpleName() + "<typeName=" + typeName + ", status=" + status.getStatus()
+                + ", message=" + status.getMessage() + ", additionalInfo=" + status.getAdditionalInfo()
+                + ", exception=" + status.getThrowable() + ">";
+    }
+
+    private static BitSet calculateRemovedAndModifiedOrdinals(BitSet latestOrdinals, BitSet previousOrdinals) {
+        // make sure we don't modify previousOrdinals or latestOrdinals
+        BitSet removedAndModified = new BitSet();
+        removedAndModified.or(previousOrdinals);
```
PaulSandoz, line 136: `return previousOrdinals.clone().andNot(latestOrdinals);`

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ObjectModificationValidator.java
```diff
@@ -0,0 +1,122 @@
...
+            HollowPrimaryKeyIndex index = new HollowPrimaryKeyIndex(readState.getStateEngine(), key);
+            int fromOrdinal = removedAndModified.nextSetBit(0);
+            while (fromOrdinal != HollowConstants.ORDINAL_NONE) {
+                Object[] candidateKey = index.getRecordKey(fromOrdinal);
+                int matchedOrdinal = index.getMatchingOrdinal(candidateKey);
+                if (matchedOrdinal != HollowConstants.ORDINAL_NONE) {
+                    T fromObject = (T) new GenericHollowObject(objectTypeState, fromOrdinal);
+                    T toObject = (T) new GenericHollowObject(objectTypeState, matchedOrdinal);
+                    if (!filter.apply(fromObject, toObject)) {
+                        throw new ValidationException("Validation failed for key " + Arrays.toString(candidateKey));
```
PaulSandoz: I wonder if validation should fail fast or should test every modified object and accumulate the failures?

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ObjectModificationValidator.java
ghost: Paul's interpretation aligns with how I've been thinking of skipping. @lkancode is the original advocate on SKIP as a status, though, so should consult with her.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ObjectModificationValidator.java
```diff
@@ -0,0 +1,122 @@
...
+import com.netflix.hollow.core.index.HollowPrimaryKeyIndex;
+import com.netflix.hollow.core.index.key.PrimaryKey;
+import com.netflix.hollow.core.read.engine.HollowTypeReadState;
+import com.netflix.hollow.core.read.engine.object.HollowObjectTypeReadState;
+import com.netflix.hollow.core.schema.HollowSchema.SchemaType;
+import java.util.Arrays;
+import java.util.BitSet;
+import java.util.function.BiPredicate;
+
+/**
```
ghost: Another stock validator. Squee!

ghost (commented):
What Paul said about SKIP. Discussion needed, otherwise looks good.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ObjectModificationValidator.java
```diff
@@ -0,0 +1,122 @@
...
+            HollowObjectTypeReadState objectTypeState = (HollowObjectTypeReadState) typeState;
+            PrimaryKey key = objectTypeState.getSchema().getPrimaryKey();
+            // this is guaranteed to give us items from the most recent cycle, not the last one
+            HollowPrimaryKeyIndex index = new HollowPrimaryKeyIndex(readState.getStateEngine(), key);
+            int fromOrdinal = removedAndModified.nextSetBit(0);
+            while (fromOrdinal != HollowConstants.ORDINAL_NONE) {
+                Object[] candidateKey = index.getRecordKey(fromOrdinal);
+                int matchedOrdinal = index.getMatchingOrdinal(candidateKey);
+                if (matchedOrdinal != HollowConstants.ORDINAL_NONE) {
+                    T fromObject = (T) new GenericHollowObject(objectTypeState, fromOrdinal);
```
PaulSandoz: FTR there is an issue here with unsafe casting (and the compiler/IDE will be informing that this is the case). A GenericHollowObject is created but the predicate may be expecting a sub-type (e.g. Movie). So a CastCastException will occur before invocation (those cast checks are inserted by the Java compiler).

akhaku: Yeah, some more work needed here - the intent was to allow people to write their filters using generated types, so after discussion with Paul I need to pass in some more arguments to this validator.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ObjectModificationValidator.java
akhaku: Question for you guys - when _would_ you think a validator has been "skipped"?

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ObjectModificationValidator.java
```diff
@@ -0,0 +1,140 @@
...
+import java.util.function.Function;
+
+/**
+ * This validator runs a provided {@link BiPredicate} on every instance of an object that was
+ * modified in a cycle. This "filter" function is called with the old object as the first argument
+ * and the new object as the second argument. If the function returns false, this validator is
+ * marked as failed and the producer cycle fails.
+ * Note that the filter function is not called with any objects that were added or removed.
+ * Also note that this validator can only be used on types that have a primary key.
+ * Example construction - note that `Movie` and `MovieAPI are generated classes:
```
akhaku: missing a closing backtick here - adding it now

ghost: Strictly speaking, `{@code Movie}` and `{@code MovieAPI}` is what you want.

I wonder if there's such a thing as a javadoc postprocessor that adds Markdown support?

akhaku: Fair point, updated

hollow/src/main/java/com/netflix/hollow/api/producer/validation/ObjectModificationValidator.java
akhaku: Paul and I looked at this together, but it gets messy because 1. clone returns an Object that needs to be cast back to a BitSet, and 2. andNot is a void function, so that needs to be on a separate line from the return anyway.

PR #266, opened 2018-09-18
Expose a consumer's initial load as a future

ghost:
Allow calling code to safely interact with a consumer after the initial data load has completed. By exposing a `CompletableFuture` callers can decide to block or receive a callback. Example of blocking until the initial load completes:

```java
HollowConsumer consumer = consumerBuilder.
        .forNamespace("tick.tock.v1-timt.1")
        .build();
consumer.getInitialLoad().join();
new HollowExplorerUIServer(consumer, port).start();
```

This is an improvement on the current guidance to call `consumer.triggerRefresh()`:

* expressed intent of waiting on the initial load versus triggering a refresh which, at a minimum, checks if there's a new version available and potentially performs another refresh
* For a new namespace without any cycles yet, `triggerRefresh()` completes successfully by initializing the consumer with an empty state, but this isn't truly the initial load. Thus calling `triggerRefresh` isn't a guarantee to have the expected dataset loaded
* relies on the background refresh thread to perform its work instead of requesting a redundant refresh

If a failure occurs while applying the initial refresh, the future is completed exceptionally.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
@@ -29,6 +29,7 @@ import com.netflix.hollow.core.read.engine.HollowReadStateEngine;
 import com.netflix.hollow.core.read.filter.HollowFilterConfig;
 import com.netflix.hollow.core.util.HollowObjectHashCodeFinder;
 import java.util.List;
+import java.util.concurrent.CompletableFuture;
 import java.util.logging.Logger;
 
 /**
@@ -38,6 +39,7 @@ public class HollowClientUpdater {
     private static final Logger LOG = Logger.getLogger(HollowClientUpdater.class.getName());
 
     private final HollowUpdatePlanner planner;
+    private final CompletableFuture<Long> initialLoad;
     private HollowDataHolder hollowDataHolder;
     private boolean forceDoubleSnapshot = false;
     private final FailedTransitionTracker failedTransitionTracker;
@@ -74,6 +76,7 @@ public class HollowClientUpdater {
         this.staleReferenceDetector.startMonitoring();
         this.metrics = metrics;
         this.metricsCollector = metricsCollector;
+        this.initialLoad = new CompletableFuture<>();
     }
 
     /**
@@ -124,6 +127,8 @@ public class HollowClientUpdater {
             metrics.updateTypeStateMetrics(getStateEngine(), version);
             if(metricsCollector != null)
                 metricsCollector.collect(metrics);
+
+            initialLoad.complete(getCurrentVersionId()); // only set the first time
             return getCurrentVersionId() == version;
         } catch(Throwable th) {
             forceDoubleSnapshotNextUpdate();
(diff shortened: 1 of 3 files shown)


hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
```diff
@@ -130,6 +176,7 @@ public synchronized boolean updateTo(long version) throws Throwable {
                 metricsCollector.collect(metrics);
             for(HollowConsumer.RefreshListener refreshListener : refreshListeners)
                 refreshListener.refreshFailed(beforeVersion, getCurrentVersionId(), version, th);
+            initialLoad.completeExceptionally(th);
```
ghost: This might be bad behavior. Originally I left the future unresolved. For consumers whose producers write often, they will have a new update to try that might succeed. Completing exceptionally means that calling code won't ever see that the consumer eventually had a successful initial load.

If a producer writes infrequently, calling code that blocks/has-callback could be blocked/waiting for a very long time or the life of the application.

I'm leaning now on removing this in favor of recovery where producers write often.

hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
```diff
@@ -51,6 +53,46 @@
...
+     * @param metrics
+     * @param metricsCollector
+     *
+     * @deprecated use {@link #HollowClientUpdater(HollowConsumer.BlobRetriever, List, HollowAPIFactory,
+     *                                             HollowConsumer.DoubleSnapshotConfig, HollowObjectHashCodeFinder,
+     *                                             HollowConsumer.ObjectLongevityConfig,
+     *                                             HollowConsumer.ObjectLongevityDetector, HollowConsumerMetrics,
+     *                                             HollowMetricsCollector, CompletableFuture)} instead
+     */
+    @Deprecated
```
PaulSandoz: Since HollowClientUpdater is internal is there an explicit need to mark as deprecated, instead can the constructor just be removed? This leads to a wider point: is the new constructor needed? HollowClientUpdater could control the instantiation of the CF rather than that being "percolated" up to  HollowConsumer that creates it then passes ownership to HollowClientUpdater.

hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
PaulSandoz: I agree. It would be nice if we could implement this in an extensible manner with a special refresh listener, then it might be possible to configure behavior, but i suspect order of refresh listener execution may be required to ensure the special refresh listener is invoked after all other refresh listeners.

PaulSandoz (approved):

hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
ghost: Good point. I waffled on this as I implemented it. I eventually landed on passing CF via `HollowClientUpdater`'s constructor for unit testing purposes, intending to pass a mock CF. When I ended up writing the unit test, I started with a mock, then just used a real CF.

That said, much of Hollow isn't written with a mockist style. We should probably align on whether we want to move in that direction before making local changes in that style which will differ from the style of the majority of existing code.

TL;DR - I'll revert to the original constructor and create the CF internally.

PR #270, opened 2018-09-25
Support the removal of the last n historical states

PaulSandoz:
This is useful for the removal of older historical state in response to memory pressure.


ghost (approved):
👍

PR #271, opened 2018-09-25
HollowPrefixIndex bugfix - multiple elements

akhaku:
* Implement a FixedLengthMultipleOccurrenceElementArray to allow storing multiple elements at a single node. Use this in the HollowPrefixIndex to allow storing multiple elements per node.
* Minor cleanup - get rid of IJ warnings in HollowPrefixIndex

Would be a good idea to review these commit by commit - the first commit is just cleanup. Or if it's easier, I can merge the first commit directly into master for ease of review.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
@@ -61,6 +61,7 @@ public class HollowPrefixIndex implements HollowTypeStateListener {
      *                        The fields in the path could reference another Object, List, Set or a Map.
      *                        The fields should be separated by ".".
      */
+    @SuppressWarnings("WeakerAccess")
     public HollowPrefixIndex(HollowReadStateEngine readStateEngine, String type, String fieldPath) {
 
         if (readStateEngine == null) throw new IllegalArgumentException("Read state engine cannot be null");
@@ -117,11 +118,8 @@ public class HollowPrefixIndex implements HollowTypeStateListener {
         BitSet ordinals = readStateEngine.getTypeState(type).getPopulatedOrdinals();
         int ordinal = ordinals.nextSetBit(0);
         while (ordinal != -1) {
-            String[] keys = getKeys(ordinal);
-            if (keys != null) {
-                for (String key : keys) {
-                    tst.insert(key, ordinal);
-                }
+            for (String key : getKeys(ordinal)) {
+                tst.insert(key, ordinal);
             }
             ordinal = ordinals.nextSetBit(ordinal + 1);
         }
@@ -135,11 +133,8 @@ public class HollowPrefixIndex implements HollowTypeStateListener {
     /**
      * This method estimates the total number of nodes that will required to create the index.
      * Override this method if lower/higher estimate is needed compared to the default implementation.
-     *
-     * @param totalWords
-     * @param averageWordLen
-     * @return
      */
+    @SuppressWarnings("WeakerAccess")
     protected long estimateNumNodes(long totalWords, long averageWordLen) {
         return totalWords * averageWordLen;
     }
@@ -179,6 +174,7 @@ public class HollowPrefixIndex implements HollowTypeStateListener {
      * @param prefix findKeysWithPrefix prefix.
      * @return An instance of HollowOrdinalIterator to iterate over ordinals that match the given findKeysWithPrefix.
      */
+    @SuppressWarnings("WeakerAccess")
     public HollowOrdinalIterator findKeysWithPrefix(String prefix) {
         TST current = this.prefixIndex;
         HollowOrdinalIterator it;
@@ -191,7 +187,6 @@ public class HollowPrefixIndex implements HollowTypeStateListener {
     /**
      * Check if the given exists in the index.
      *
-     * @param key
      * @return boolean value indicating if the key exists in the index.
      */
     public boolean contains(String key) {
@@ -209,6 +204,7 @@ public class HollowPrefixIndex implements HollowTypeStateListener {
      * Remember to call detachFromDeltaUpdates to stop the delta changes.
      * NOTE: Each delta updates creates a new prefix index and swaps the new with current.
      */
+    @SuppressWarnings("WeakerAccess")
     public void listenForDeltaUpdates() {
         readStateEngine.getTypeState(type).addListener(this);
     }
@@ -216,6 +212,7 @@ public class HollowPrefixIndex implements HollowTypeStateListener {
     /**
      * Stop delta updates for this index.
      */
+    @SuppressWarnings("WeakerAccess")
     public void detachFromDeltaUpdates() {
         readStateEngine.getTypeState(type).removeListener(this);
     }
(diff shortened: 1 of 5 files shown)


ghost (approved):
Is good!

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
```diff
@@ -31,45 +32,68 @@
...
 
+    /**
+     * @see #HollowPrefixIndex(HollowReadStateEngine, String, String)
+     * This constructor defaults the estimatedMaxStringDuplicates to 4. If you expect a large
+     * number of duplicate strings across your type, you should provide your own estimate for
+     * estimatedMaxStringDuplicates.
+     */
+    @SuppressWarnings("WeakerAccess")
+    public HollowPrefixIndex(HollowReadStateEngine readStateEngine, String type, String fieldPath) {
+        this(readStateEngine, type, fieldPath, 4);
```
kineshsatiya, line 66: 4?

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
```diff
@@ -31,45 +32,68 @@
...
+     *                                     List, Set or a Map. The fields should be separated by ".".
+     * @param estimatedMaxStringDuplicates The estimated number of strings that are duplicated
+     *                                     across instances of your type. Note that this means an
+     *                                     exactly matching string, not a prefix match. This
+     *                                     parameter affects the efficiency of the index building.
      */
-    public HollowPrefixIndex(HollowReadStateEngine readStateEngine, String type, String fieldPath) {
+    @SuppressWarnings("WeakerAccess")
+    public HollowPrefixIndex(HollowReadStateEngine readStateEngine, String type, String fieldPath,
+            int estimatedMaxStringDuplicates) {
```
kineshsatiya, line 85: We can calculate this.

hollow/src/main/java/com/netflix/hollow/core/memory/encoding/FixedLengthMultipleOccurrenceElementArray.java
```diff
@@ -0,0 +1,161 @@
...
+ * there are 8 elements at one node index, it will contain enough space to store at least 8 elements
+ * at each index.
+ *
+ * Running out of space at any index triggers a relatively expensive resize operation, where we
+ * create storage of a multiple (currently 1.5x) of the previous storage and copy items over, but
+ * this can be alleviated by passing in a better guess for max elements per node.
+ * Note that this class is currently designed to be used for a relatively small number for
+ * bitsPerElement.
+ */
+public class FixedLengthMultipleOccurrenceElementArray {
```
kineshsatiya, line 41: Should this be public API?

hollow/src/main/java/com/netflix/hollow/core/memory/encoding/FixedLengthMultipleOccurrenceElementArray.java
```diff
@@ -0,0 +1,161 @@
...
+    private final int bitsPerElement;
+    private final long elementMask;
+    private final long numNodes;
+
+    private volatile FixedLengthElementArray storage;
+    private volatile int maxElementsPerNode;
+
+    public FixedLengthMultipleOccurrenceElementArray(ArraySegmentRecycler memoryRecycler,
+            long numNodes, int bitsPerElement, int maxElementsPerNodeEstimate) {
+        nodesWithOrdinalZero = new FixedLengthElementArray(memoryRecycler, numNodes);
```
kineshsatiya, line 56: You can probably get away with this if we store a size with the other one.

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
akhaku: Yeah, picked a default that is small since we expect that to be the common case, but not too small so as to cause resizing even when we find one or two duplicates.

hollow/src/main/java/com/netflix/hollow/core/memory/encoding/FixedLengthMultipleOccurrenceElementArray.java
akhaku: It's public in the same way as FixedLengthElementArray is - it's declared as a "public" class (since it's used outside of this package), but not meant to be used outside of internal code.

hollow/src/main/java/com/netflix/hollow/core/memory/encoding/FixedLengthMultipleOccurrenceElementArray.java
akhaku: Interesting... Something like storing the current size for each node in an int[]?

hollow/src/main/java/com/netflix/hollow/core/memory/encoding/FixedLengthMultipleOccurrenceElementArray.java
kineshsatiya: If it's not used anywhere other than the index, would it better to make this inner class in the index? That was the same rationale for making TST as an inner class.

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
kineshsatiya: It would be better to calculate/estimate this. I'm not sure if it's a good idea to start with a default value that may or may not work (causing resizes for every string that is duplicated more than 4 times).

hollow/src/main/java/com/netflix/hollow/core/memory/encoding/FixedLengthMultipleOccurrenceElementArray.java
akhaku: The goal was to make it usable outside of the HollowPrefixKeyIndex, even if it isn't used anywhere else right now. I'd prefer to avoid an inner class as it causes bloat within that class, but I'm not opposed to making it package-private and moving it to the same package as the index (until we find another use case for it) if you think that's a good idea.

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
akhaku: Need to give this some more thought

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
akhaku: Yeah I agree, it would be nice to come up with a good estimate

hollow/src/main/java/com/netflix/hollow/core/memory/encoding/FixedLengthMultipleOccurrenceElementArray.java
kineshsatiya: We could have the first element in `storage` for every node that indicates the size of the list.

akhaku: Hm, we might be able make that work for our purposes because the number of bits needed to represent the size is the same number of bits needed to represent the long elements themselves, because in our case they are ordinals, and the size is bounded by the highest ordinal.
The only issue then is duplicates - we need to de-dupe in our storage to ensure that the size of elements stored per nodeIndex cannot grow beyond its upper bound (1 << numBitsPerElement - 1) - right now we don't handle duplicates and that condition does not hold.

PR #277, opened 2018-10-01
Use volatile reads

kineshsatiya:
- bug-fix for infinite loop on reads, change hot/cold swapping pattern with reading volatile directly to avoid error conditions


kineshsatiya:
Closing this PR to create a new one. My changes overrode the changes from the master branch.

PR #278, opened 2018-10-01
Fix volatile reads

kineshsatiya:
- Avoid using the swapping pattern to handle updates in the indexes to avoid potential infinite loops. 

- Rely on volatile reads to check if the latest changes have been read from the index.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/index/HollowSparseIntegerSet.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/index/HollowSparseIntegerSet.java
@@ -34,7 +34,6 @@ public class HollowSparseIntegerSet implements HollowTypeStateListener {
     private final FieldPath fieldPath;
     private final IndexPredicate predicate;
 
-    protected SparseBitSet sparseBitSet;
     protected volatile SparseBitSet sparseBitSetVolatile;
 
     private Set<Integer> valuesToSet;
@@ -88,7 +87,7 @@ public class HollowSparseIntegerSet implements HollowTypeStateListener {
         build();
     }
 
-    protected synchronized void build() {
+    protected void build() {
 
         // initialize an instance of SparseBitSet
         initSet(Integer.MAX_VALUE);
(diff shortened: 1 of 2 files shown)


PaulSandoz (approved):

hollow/src/main/java/com/netflix/hollow/core/index/HollowSparseIntegerSet.java
```diff
@@ -106,24 +105,24 @@ protected synchronized void build() {
...
     }
 
     protected void set(int ordinal) {
         if (predicate.shouldIndex(ordinal)) {
             Object[] values = fieldPath.findValues(ordinal);
             if (values != null && values.length > 0) {
+                SparseBitSet bitSet = sparseBitSetVolatile;
                 for (Object value : values) {
-                    sparseBitSet.set((int) value);
+                    bitSet.set((int) value);
```
ghost, line 117: I thought this was a bug at first, but I see that `SparseBitSet.set` is using compare-and-set under the covers.

hollow/src/main/java/com/netflix/hollow/core/index/HollowSparseIntegerSet.java
```diff
@@ -214,12 +216,12 @@ public void removedOrdinal(int ordinal) {
...
-        SparseBitSet updated = sparseBitSet;
+        boolean didSomeWork = false;
+        SparseBitSet updated = sparseBitSetVolatile;
         // first check if the max value among the new values to be added is more than the max value of the existing sparse bit set.
-        if (valuesToSet.size() > 0 && maxValueToSet > sparseBitSet.findMaxValue()) {
-            updated = SparseBitSet.resize(sparseBitSet, maxValueToSet);
-            swap = true;
+        if (valuesToSet.size() > 0 && maxValueToSet > updated.findMaxValue()) {
+            updated = SparseBitSet.resize(updated, maxValueToSet);
+            didSomeWork = true;
```
ghost, line 224: Because `swap` was a bit of a misnomer, since we're potentially mutating the current bitset unless `sparseBitSetVolatile` was reassigned by some other thread while we're doing work?

ghost: Oh, never mind. I see that the only modifications we do are to a resized copy.

ghost (approved):

PR #279, opened 2018-10-01
Use volatile reads for optimistic reading of state within a loop.

PaulSandoz:
A plain read may get hoisted out the loop by the compiler when
optimizing the code, as a result the loop might not terminate or
(perhaps eventually) throw some runtime exception/error.

The existing pattern could be utilized with opaque access using Java 9,
see VarHandle.getOpaque.  However, even then unless there is a
compelling performance benefit, i recommend using the simple pattern for
better understanding and maintainability.


kineshsatiya (approved):
Looks good to me.

dkoszewnik (approved):

PR #281, opened 2018-10-03
add a UniqueKeyIndex interface to generated indices

ghost:
A [suggestion from chat](https://gitter.im/Netflix/hollow?at=5bb218d83844923661c9eea3)

> Any thoughts or plans on creating a common interface for the generated primary key indexes?
They all have the findMatchmethod so would be nice if that was declared in a common interface. That way you could store indexes in a map or the like without the need to create you own interface or fallback to object.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/codegen/indexes/HollowUniqueKeyIndexGenerator.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/codegen/indexes/HollowUniqueKeyIndexGenerator.java
@@ -21,6 +21,7 @@ import com.netflix.hollow.api.codegen.CodeGeneratorConfig;
 import com.netflix.hollow.api.codegen.HollowAPIGenerator;
 import com.netflix.hollow.api.consumer.HollowConsumer;
 import com.netflix.hollow.api.consumer.index.AbstractHollowUniqueKeyIndex;
+import com.netflix.hollow.api.consumer.index.HollowUniqueKeyIndex;
 import com.netflix.hollow.api.custom.HollowAPI;
 import com.netflix.hollow.core.HollowDataset;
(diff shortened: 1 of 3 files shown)


hollow/src/main/java/com/netflix/hollow/api/codegen/indexes/HollowUniqueKeyIndexGenerator.java
```diff
@@ -109,6 +116,7 @@ protected void genPublicAPIs(StringBuilder builder) {
     }
 
     protected void genFindMatchAPI(StringBuilder builder) {
+        builder.append("    @Override\n");
```
PaulSandoz: Is that conditional on implementing the the interface?

hollow/src/main/java/com/netflix/hollow/api/consumer/index/UniqueKeyIndex.java
```diff
@@ -0,0 +1,5 @@
+package com.netflix.hollow.api.consumer.index;
+
+public interface UniqueKeyIndex<API, T> {
```
PaulSandoz: The API type variable is unused.

Separately i believe AbstractHollowUniqueKeyIndex can be declared more accurately as:
```
public abstract class AbstractHollowUniqueKeyIndex<API extends HollowAPI, T>
```

hollow/src/main/java/com/netflix/hollow/api/codegen/indexes/HollowUniqueKeyIndexGenerator.java
ghost: It should have an explicit condition. It worked that way in effect because the `PrimaryKeyIndexGenerator` overrode `genFindMatchAPI`, but it's more correct and easier to understand with your suggestion.

PR #283, opened 2018-10-04
Add subheadings to field paths section

DanielThomas:
I missed that hash indexes didn't support automatic expansion, so I added subheadings so you don't miss the last paragraph.


ghost (approved):
Winning!

ghost:
Docs don't automatically republish, but I manually rebuilt and pushed the docsite:

https://hollow.how/indexing-querying/#primary-and-hash-keys

Thanks for the edits!

PR #284, opened 2018-10-04
Make `WriteState` inoperable after populate stage completes

ghost:
* [x] add `close()` to `WriteStateImpl` and call it after populate stage in `runCycle()
* [x] make write state method (aside from `close()`) throw after being closed
* [x] make `getStateEngine()`, et al throw when closed
* [x] Javadoc
  * [x] document `@throws` on `WriteState`'s `add`, `getObjectMapper()`, and `getStateEngine()`. Also describe the consequences of caching them in locals that are closed over in async tasks
  * [x] document the contract on `Populator` interface
  * [x] reference the contract on `Producer.runCycle` with a link to `Populator` for details

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
@@ -428,11 +428,10 @@ public class HollowProducer {
         try {
             // 1a. Prepare the write state
             writeEngine.prepareForNextCycle();
-            WriteState writeState = new WriteStateImpl(toVersion, objectMapper, readStates.current());
 
             // 2. Populate the state
             ProducerStatus.Builder populateStatus = listeners.firePopulateStart(toVersion);
-            try {
+            try(CloseableWriteState writeState = new CloseableWriteState(toVersion, objectMapper, readStates.current())) {
                 task.populate(writeState);
                 populateStatus.success();
             } catch (Throwable th) {
@@ -445,9 +444,9 @@ public class HollowProducer {
             // 3. Produce a new state if there's work to do
             if(writeEngine.hasChangedSinceLastCycle()) {
                 // 3a. Publish, run checks & validation, then announce new state consumers
-                publish(writeState, artifacts);
+                publish(toVersion, artifacts);
 
-                ReadStateHelper candidate = readStates.roundtrip(writeState);
+                ReadStateHelper candidate = readStates.roundtrip(toVersion);
                 cycleStatus.version(candidate.pending());
                 candidate = checkIntegrity(candidate, artifacts);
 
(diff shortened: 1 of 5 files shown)


hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -428,12 +428,13 @@ protected void runCycle(Populator task, ProducerStatus.Builder cycleStatus, long
...
             // 1a. Prepare the write state
             writeEngine.prepareForNextCycle();
-            WriteState writeState = new WriteStateImpl(toVersion, objectMapper, readStates.current());
+            WriteStateImpl writeState = new WriteStateImpl(toVersion, objectMapper, readStates.current());
 
             // 2. Populate the state
             ProducerStatus.Builder populateStatus = listeners.firePopulateStart(toVersion);
             try {
                 task.populate(writeState);
+                writeState.seal();
```
PaulSandoz: Should be performed in a finally block.

I would argue that the complete WriteState instance should become inoperative if the instance is operated on beyond the scope of population stage (that includes multiple threads and/or the current thread if the instance leaks out). That sends a very clear message on the behavior and is easier to specify. In this respect the instance is not sealed but made inoperable.

hollow/src/main/java/com/netflix/hollow/api/producer/WriteStateImpl.java
```diff
@@ -26,9 +28,12 @@
...
+    private static final String LATE_MODIFICATION_MESSAGE =
+            "attempt to modify state after populate stage complete; version=%d";
     private final long version;
-    private final HollowObjectMapper objectMapper;
-    private final HollowProducer.ReadState priorReadState;
+    private HollowObjectMapper objectMapper;
+    private HollowProducer.ReadState priorReadState;
+    private volatile boolean sealed = false;
 
     protected WriteStateImpl(long version, HollowObjectMapper objectMapper, HollowProducer.ReadState priorReadState) {
```
PaulSandoz: Give WriteStateImpl is a package protected final class the protected modified is redundant.

hollow/src/main/java/com/netflix/hollow/api/producer/WriteStateImpl.java
```diff
@@ -26,9 +28,12 @@
  * @author Tim Taylor {@literal<tim@toolbear.io>}
  */
 final class WriteStateImpl implements HollowProducer.WriteState {
+    private static final String LATE_MODIFICATION_MESSAGE =
+            "attempt to modify state after populate stage complete; version=%d";
     private final long version;
-    private final HollowObjectMapper objectMapper;
-    private final HollowProducer.ReadState priorReadState;
```
PaulSandoz: Why is final removed?

ghost: Oh, that's on accident. At one point I had it implementing `AutoCloseable` and I was nulling those out. I reverted that bit, but forgot to bring back `final`.

ghost: That said, per your suggestion above to make it inoperable after the populate stage, I might bring back nulling out those fields when `seal()` (or probably `disable()`) is called.

hollow/src/main/java/com/netflix/hollow/api/producer/WriteStateImpl.java
ghost: Removed it. Thanks.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
ghost: > Should be performed in a finally block.

Good catch. Fixed.

> I would argue that the complete WriteState instance should become inoperative...

So all public methods would start throwing? And we either expose package-private accessors *or* stop using `WriteState` internally to pass things around?

The current implementation of `HollowIncrementalProducer` relies on `getStateEngine()`, but I'm not opposed to spending a little bit of time to make the incremental producer work with whatever changes we want to make to `HollowProducer`.

ghost:
I must have started this branch from my unique-key-index branch, which is why commit `6a7d2c42b9c61772992b4bfa129e3754798e0b36` is showing up. Once we're done peer reviewing I'll rebase that away. I'm avoiding rebasing now since that clobbers reviewer comments it seems.

hollow/src/main/java/com/netflix/hollow/api/producer/SealableWriteState.java
```diff
@@ -25,29 +25,36 @@
...
+        if (sealed)
+            throw new SealedWriteStateException(this);
         return objectMapper;
     }
 
     @Override
-    public HollowWriteStateEngine getStateEngine() {
+    public HollowWriteStateEngine getStateEngine() throws SealedWriteStateException {
+        if (sealed)
+            throw new SealedWriteStateException(this);
```
ghost: `HollowIncrementalProducer` relies on `getStateEngine()`. I did a cursory look at it and it seems that all uses are before sealing, but I'd want to review it more thoroughly and try it out before releasing this.

PR #286, opened 2018-10-05
Add possibility to deprecate API

koperm:
As the created API changes in a lifetime, I would like to be able to `@Deprecate` accessor methods. This PR gives that opportunity. 

Example:
```
@DeprecatedApi("@deprecated Use {@link #getPrePaddingTimeRef} instead")
private final long prePaddingTime;
```

Model annotated field transformed into:

```
/**
* @deprecated Use {@link #getPrePaddingTimeRef} instead.
*/
@Deprecated
public long getPrePaddingTime() {
    return delegate().getPrePaddingTime(ordinal);
}
```

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/codegen/objects/HollowObjectJavaGenerator.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/codegen/objects/HollowObjectJavaGenerator.java
@@ -35,6 +35,7 @@ import com.netflix.hollow.api.objects.HollowObject;
 import com.netflix.hollow.core.HollowDataset;
 import com.netflix.hollow.core.schema.HollowObjectSchema;
 import com.netflix.hollow.core.schema.HollowObjectSchema.FieldType;
+import com.netflix.hollow.core.write.objectmapper.DeprecatedApi;
 import com.netflix.hollow.tools.stringifier.HollowRecordStringifier;
 import java.util.Set;
 
@@ -117,6 +118,9 @@ public class HollowObjectJavaGenerator extends HollowConsumerJavaFileGenerator {
 
     private void appendAccessors(StringBuilder classBuilder) {
         for(int i=0;i<schema.numFields();i++) {
+
+            classBuilder.append(generateFieldAccessorDeprecatedApiAnnotation(schema.getDeprecatedApiAnnotation(i)));
+
             switch(schema.getFieldType(i)) {
                 case BOOLEAN:
                     classBuilder.append(generateBooleanFieldAccessor(i));
@@ -146,6 +150,17 @@ public class HollowObjectJavaGenerator extends HollowConsumerJavaFileGenerator {
 
             classBuilder.append("\n\n");
         }
+    }
+
+    private static String generateFieldAccessorDeprecatedApiAnnotation(DeprecatedApi deprecatedApiAnnotation) {
+        StringBuilder builder = new StringBuilder();
+        if (deprecatedApiAnnotation != null) {
+            builder.append("    /**\n")
+                    .append("    * " + deprecatedApiAnnotation.value() + "\n")
+                    .append("    */\n")
+                    .append("    @Deprecated\n");
+        }
+        return builder.toString();
     }
 
     private String generateByteArrayFieldAccessor(int fieldNum) {
--- a/hollow/src/main/java/com/netflix/hollow/api/codegen/objects/HollowObjectJavaGenerator.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/codegen/objects/HollowObjectJavaGenerator.java
@@ -154,11 +154,17 @@ public class HollowObjectJavaGenerator extends HollowConsumerJavaFileGenerator {
 
     private static String generateFieldAccessorDeprecatedApiAnnotation(DeprecatedApi deprecatedApiAnnotation) {
         StringBuilder builder = new StringBuilder();
+
         if (deprecatedApiAnnotation != null) {
-            builder.append("    /**\n")
-                    .append("    * " + deprecatedApiAnnotation.value() + "\n")
-                    .append("    */\n")
-                    .append("    @Deprecated\n");
+            String annotationValue = deprecatedApiAnnotation.value();
+
+            if (!annotationValue.isEmpty()) {
+                builder.append("    /**\n")
+                        .append("     * @deprecated ").append(annotationValue).append("\n")
+                        .append("     */\n");
+            }
+
+            builder.append("    @Deprecated\n");
         }
         return builder.toString();
     }
--- a/hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/DeprecatedApi.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/DeprecatedApi.java
@@ -6,6 +6,46 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+
+/**
+ * This annotation is meant to be used on data model fields to point
+ * deprecation and discourage to use them.
+ *
+ * <p>An annotated field with {@code @DeprecatedApi} will be transformed during
+ * Consumer API Generation into snippets consisting of JavaDoc information,
+ * if any exists, and {@link java.lang.Deprecated} annotation on accessor method.
+ *
+ * <p>Example:
+ *
+ * <p>The following data model snippet:
+ *
+ * <pre>
+ *  {@literal @DeprecatedApi}("Use {{@literal @link} #getNewValue()} instead.")
+ *  private final long value;
+ *
+ *  private final long newValue;
+ *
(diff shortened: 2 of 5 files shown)


ghost:
@koperm I like the intent of this PR. We'd like some time to review internally before committing to this API

hollow/src/main/java/com/netflix/hollow/api/codegen/objects/HollowObjectJavaGenerator.java
```diff
@@ -148,6 +152,17 @@ private void appendAccessors(StringBuilder classBuilder) {
         }
     }
 
+    private static String generateFieldAccessorDeprecatedApiAnnotation(DeprecatedApi deprecatedApiAnnotation) {
+        StringBuilder builder = new StringBuilder();
+        if (deprecatedApiAnnotation != null) {
+            builder.append("    /**\n")
+                    .append("    * " + deprecatedApiAnnotation.value() + "\n")
```
ghost: Two things:
* you could automatically add the `@deprecated` JavaDoc so that the annotation value doesn't have to redundantly include it, e.g. `@DeprecatedApi("Use {@link #getPrePaddingTimeRef} instead")`
* avoid string concatenation when already working with a `StringBuilder`; line 159 should be split into 3 `.append(...)` calls.

hollow/src/main/java/com/netflix/hollow/api/codegen/objects/HollowObjectJavaGenerator.java
```diff
@@ -148,6 +152,17 @@ private void appendAccessors(StringBuilder classBuilder) {
         }
     }
 
+    private static String generateFieldAccessorDeprecatedApiAnnotation(DeprecatedApi deprecatedApiAnnotation) {
+        StringBuilder builder = new StringBuilder();
+        if (deprecatedApiAnnotation != null) {
+            builder.append("    /**\n")
```
ghost: If we remove the default value from the annotation, then you could check if `deprecatedApiAnnotation.value()` and only append the JavaDoc when it was non-null. But always add the `@Deprecated` annotation to the generated accessor.

hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/DeprecatedApi.java
```diff
@@ -0,0 +1,15 @@
...
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.METHOD;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({FIELD, METHOD})
```
ghost: Target shouldn't have `METHOD`. Only fields are considered by `HollowObjectMapper`

hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/DeprecatedApi.java
```diff
@@ -0,0 +1,15 @@
...
+public @interface DeprecatedApi {
+    String value() default "Empty Javadoc";
```
ghost: I would omit the default and guard against `null` return for `value()` when generating the field accessor.

ghost (changes requested):

hollow/src/main/java/com/netflix/hollow/api/codegen/objects/HollowObjectJavaGenerator.java
ghost: That way the following is also valid usage:

```
@DeprecatedApi
private final long prePaddingTime;
```

ghost: For that matter, we could also support the core Java `@Deprecated` annotation for those cases  where you don't want a JavaDoc explanation:

```
@Deprecated
private final long prePaddingTime;
```

koperm:
@toolbear thanks for your insight and reply. 
I updated the code according to the tips and look forward to hearing from you.

hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/DeprecatedApi.java
```diff
@@ -1,15 +1,13 @@
...
 
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
-
 @Retention(RetentionPolicy.RUNTIME)
-@Target({FIELD, METHOD})
+@Target(FIELD)
 public @interface DeprecatedApi {
```
PaulSandoz, line 51: We should document this annotation, otherwise developers will not know how/where to use it and what the side-effects are.  I wish we could somehow reuse the existing @Deprecated annotation with JavaDoc but we loose the latter when reflectively processing.  Should it be recommended this annotation be used in conjunction with @Deprecated?

I wonder this annotation should not be directly passed through to the `HollowObjectSchema.addField` methods, since that abstraction is generally free of Java's reflective types.  More generally perhaps there will be other kinds of Hollow schema field modifier in the future?

PaulSandoz (changes requested):
I recommend just adding one new method `HollowObjectSchema.addField`, given the use of deprecation is likely rare.

I wonder if we can generalize this to pass in a Set of some "meta-data" (Hollow field schema annotation), since deprecation is just one of many possible (not yet thought up) field modifiers or annotations.  That relates to a comment on the `@DeprecatedApi` annotation.

hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/DeprecatedApi.java
koperm: > I recommend just adding one new method HollowObjectSchema.addField, given the use of deprecation is likely rare.

I need two of the `addField` method declarations in `HollowObjectTypeMapper`. The third one is the result of "telescoping".

> I wonder if we can generalize this to pass in a Set of some "meta-data" (Hollow field schema annotation), since deprecation is just one of many possible (not yet thought up) field modifiers or annotations. That relates to a comment on the @DeprecatedApi annotation.

Annotation processing is one of the scopes that can be took in consideration. If you would like to extend existing solution, which actually pass only types, there should be same abstract placeholder for passed through `meta-data`. 
 
> We should document this annotation, otherwise developers will not know how/where to use it and what the side-effects are.

Will do.

>  I wish we could somehow reuse the existing @deprecated annotation with JavaDoc but we loose the latter when reflectively processing. Should it be recommended this annotation be used in conjunction with @deprecated?

The goal of this PR is to provide possibility of using @Deprecated with JavaDoc comment, thus this specific annotation should be strictly related to the Java's @Deprecated.

> I wonder this annotation should not be directly passed through to the HollowObjectSchema.addField methods, since that abstraction is generally free of Java's reflective types. More generally perhaps there will be other kinds of Hollow schema field modifier in the future?

Previously mentioned abstract 'place-holder' will be enough? What future modifiers it should contains?

PaulSandoz: > I need two of the addField method declarations in HollowObjectTypeMapper. The third one is the result of "telescoping".

koperm:
I added JavaDoc and replied to @PaulSandoz comments. Give some feedback, please.

koperm:
@PaulSandoz, @toolbear any chance to move on with this PR?

PaulSandoz:
@koperm sorry for the delay, been busy with other things. The JavaDoc looks good but overall the implementation is still too specific to deprecation.

Another use-case we recently thought about was declaring a field to be non-null.  At the validation stage a validator could check if a null-value was assigned. This adds another aspect to any such design in that we don't want to add a dependency in hollow to an external library but we don't necessarily want to define our own annotation that's different to what developers ordinarily use, which means considering an abstraction to map a third party annotation to the associated hollow schema field artifact. 

So IMHO this requires more thought and work to get this right. How about we work together under a feature branch in the hollow repo?

koperm:
@PaulSandoz @Deprected annotation has a flaw because doesn't allow for any comment/message. That's why @DeprecatedApi was introduced, but the in general approach any already existed annotation could be used, including the Java's @Deprecated. 

As you mentioned a third-party dependencies problem arises in the generic approach. In the case of third-party annotations, `import` statement needs to be taken into consideration. The second thing is the validation of created java files from the model: AFAIK the validation involves the compilation of the files. This is still a valid approach in case of third-party annotations? 

These and probably many more problems should be overcome to accomplish generic approach.

I admit that this simplified feature would be very useful to me but in the long run generic way could give much more. Feel free to do whatever you consider right, just give me some notification. I would be glad to help.

PaulSandoz:
@koperm re: @deprecatedapi we are in strong agreement :-)

Next week i will create a pull request merging in your deprecated feature with a more general mechanism for "modifiers" on fields and notify you for review and feedback.

PR #289, opened 2018-10-07
Handle internal thread spawning so it is cleaned up.

### LINKED ISSUE #288: Thread leak from StaleHollowReferenceDetector in tests
The StaleHollowReferenceDetector creates a daemon thread internal that can never be terminated (https://github.com/Netflix/hollow/blob/master/hollow/src/main/java/com/netflix/hollow/api/client/StaleHollowReferenceDetector.java#L128). This causes a thread leak during tests and in a big enough project, you can easily cause thread exhaustion.

The method reference to the Thread should be moved to be an instance variable and a the `finalize()` method on `StaleHollowReferenceDetector` should stop the thread.

johnrengelman:
Closes #288 

Original Behavior during tests:
image

Thread cleanup:
image

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/client/StaleHollowReferenceDetector.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/client/StaleHollowReferenceDetector.java
@@ -58,6 +58,8 @@ public class StaleHollowReferenceDetector {
 
     private final StackTraceRecorder stackTraceRecorder;
 
+    private Thread monitor;
+
     public StaleHollowReferenceDetector(HollowConsumer.ObjectLongevityConfig config, HollowConsumer.ObjectLongevityDetector detector) {
         this.handles = new ArrayList<HollowWeakReferenceHandle>();
         this.config = config;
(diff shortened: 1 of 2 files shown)


hollow/src/main/java/com/netflix/hollow/api/client/StaleHollowReferenceDetector.java
```diff
@@ -111,27 +113,51 @@ private synchronized void housekeeping() {
...
-        t.start();
+        if (monitor == null) {
+            monitor = new Thread(new Monitor(this));
+            monitor.setDaemon(true);
+            monitor.start();
+        }
+    }
+
+    @Override
+    protected void finalize() throws Throwable {
```
PaulSandoz: It's often preferable to avoid finalize methods if possible.  In this case the monitor thread could complete after the weak reference is cleared instead of relying on the interrupted status set by the finalize method. Then no additional field is required.

PaulSandoz (approved):
Thanks, this is good. We can finesse the implementation (avoiding a finalizer) later if necessary.

PR #298, opened 2018-10-12
Add Jackson module to allow serialization of HollowObjects

DanielThomas:
This Jackson module allows direct serialization of `HollowObject`s using Jackson datamapper.

- It supports both the default and ergonomic generated APIs
- You can optionally preserve the original order of the fields, because reflection doesn't return methods in their declared order. Implemented by walking the class files to avoid coupling the module to a single consumer/API to get at the schema
- Single-value non-reference types are automatically expanded

The only situation where the serialized json won't match the original object is when a `HollowObject` with more than one time, or a reference type, is used as a `Map` key. In that case, we fall back to the default Jackson behavior of calling `toString` on the object.

I couldn't see a logic place to put this in the documentation, and you might want this to be a separate artifact, rather than bundled with the json adapter, but felt like a good start.

### DIFF
--- a/hollow-jsonadapter/build.gradle
+++ b/hollow-jsonadapter/build.gradle
@@ -1,12 +1,14 @@
 apply plugin: 'java'
- 
+
 dependencies {
     compile project(':hollow')
 
-    compile 'com.fasterxml.jackson.core:jackson-core:2.4.3'
-    compile 'com.fasterxml.jackson.core:jackson-databind:2.4.3'
+    compile 'com.fasterxml.jackson.core:jackson-core:2.6.0'
+    compile 'com.fasterxml.jackson.core:jackson-databind:2.6.0'
     compile 'commons-io:commons-io:2.4'
     compile 'commons-lang:commons-lang:latest.release'
+    compile 'org.ow2.asm:asm:6.2.1'
 
-    testCompile 'junit:junit:4.11'
+    testCompile project(':hollow').sourceSets.test.output
+    testCompile project(path: ':hollow', configuration: 'testCompile')
 }
--- a/hollow-jsonadapter/src/main/java/com/netflix/hollow/jsonadapter/HollowModule.java
+++ b/hollow-jsonadapter/src/main/java/com/netflix/hollow/jsonadapter/HollowModule.java
@@ -1,0 +1,205 @@
+package com.netflix.hollow.jsonadapter;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.BeanDescription;
+import com.fasterxml.jackson.databind.JavaType;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializationConfig;
+import com.fasterxml.jackson.databind.SerializerProvider;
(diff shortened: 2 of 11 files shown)


ghost:
🔜 

I've asked for feedback from Drew and Paul. Thanks for your patience, Danny.

ghost (approved):

ghost (approved):
Will need to fix merge conflicts, but hold off until additional feedback.

hollow-jsonadapter/src/main/java/com/netflix/hollow/jsonadapter/HollowModule.java
```diff
@@ -0,0 +1,268 @@
...
+        public List<BeanPropertyWriter> orderProperties(SerializationConfig config, BeanDescription beanDesc, List<BeanPropertyWriter> beanProperties) {
+            if (preservePropertyOrder) {
+                List<String> orderedBeanPropertyNames = declaredOrderBeanPropertyNames(beanDesc);
+                beanProperties.sort(Comparator.comparingInt(o -> orderedBeanPropertyNames.indexOf(o.getName())));
+            }
+            return beanProperties;
+        }
+
+        /**
+         * Unfortunately, {@link Class#getDeclaredMethods()} doesn't return methods in their declared order. {@link HollowObject#getSchema()} would allow us to get at the field ordering,
```
PaulSandoz, line 197: What is the declared order? There is no guarantee that the order of methods declared in the byte code reflect the order of methods declared in the Java source.

dkoszewnik:
Hi @DanielThomas, in what scenario would you prefer this over the `HollowRecordJsonStringifier`?

DanielThomas:
@PaulSandoz though there's no guarantee, it seems the order in the class file is the order as declared.

@dkoszewnik the objects in our model have the same shape as our API. For many of our APIs we return Hollow objects directly or objects composed into calculated values at runtime and have them serialize directly in Spring MVC responses.

hollow-jsonadapter/build.gradle
```diff
@@ -1,12 +1,14 @@
...
 dependencies {
     compile project(':hollow')
 
-    compile 'com.fasterxml.jackson.core:jackson-core:2.4.3'
-    compile 'com.fasterxml.jackson.core:jackson-databind:2.4.3'
+    compile 'com.fasterxml.jackson.core:jackson-core:2.6.0'
+    compile 'com.fasterxml.jackson.core:jackson-databind:2.6.0'
     compile 'commons-io:commons-io:2.4'
     compile 'commons-lang:commons-lang:latest.release'
+    compile 'org.ow2.asm:asm:6.2.1'
```
PaulSandoz, line 10: I have some concern about depending on asm. It can be problematic when there conflicts with other libraries depending on other versions. This is required for accessing methods in declaration order, but it is not clear to me it achieves the desired behaviour (if declaration order is source declaration order).

PaulSandoz:
A general concern i have is the Jackson JSON serializer is operating by reflecting over the generated API classes. This is likely to be fragile. The generated API is a representation on the schema (as are the  POJOs on the producing side) and that representation could change.

It is the schema + data that should ideally be operated on. For example, see `HollowRecordJsonStringifier`.  Is there a reason that class cannot be used to produce JSON?

PR #310, opened 2018-10-21
Fix HollowWriteStateEnginePrimaryKeyHasher to correctly retrieve fields values

### LINKED ISSUE #309: HollowWriteStateEnginePrimaryKeyHasher is not hashing correctly
I noticed there is a problem with hashing for objects with primary keys.

I have the following classes:
```java
class A {
   Map<B, C> map;
}

@HollowPrimaryKey{...}
class B {...}

class C {...}
```

The problem appears when trying to write the snapshot, `HollowWriteStateEnginePrimaryKeyHasher` tries to hash the `B` objects by the fields specified in the `@HollowPrimaryKey` annotation, but it fails due to incorrect retrieving the fields values.

The navigateToField method https://github.com/Netflix/hollow/blob/4796e71e68e6494cc39bc34815d0dc1d8fd05401/hollow/src/main/java/com/netflix/hollow/core/write/HollowWriteStateEnginePrimaryKeyHasher.java#L86-L88
computes incorrectly the offset. Instead of using `i` to compute the offset of the fields before the desired field, it always uses `fieldIdx`. 

It seems the solution would be to replace: `switch(schema.getFieldType(fieldIdx))` -> `switch(schema.getFieldType(i))`

flore77:
Fixes #309.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/write/HollowWriteStateEnginePrimaryKeyHasher.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/write/HollowWriteStateEnginePrimaryKeyHasher.java
@@ -85,7 +85,7 @@ class HollowWriteStateEnginePrimaryKeyHasher {
     
     private long navigateToField(HollowObjectSchema schema, int fieldIdx, SegmentedByteArray data, long offset) {
         for(int i=0;i<fieldIdx;i++) {
-            switch(schema.getFieldType(fieldIdx)) {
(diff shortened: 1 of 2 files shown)


PaulSandoz:
@flore77 would you point your PR to the hollow4 branch? then i will merge it in.

PaulSandoz (approved):

flore77:
@PaulSandoz will do, thanks for reviewing this.

Issue #308, opened 2018-10-18
Document the use of @HollowTransient on fields


DanielThomas:
There are cases in other JVM languages, such as delegates in Kotlin, where it's not possible to mark a field as transient. Support for a `@HollowIgnore` field annotation in addition to `transient` would allow these fields to be ignored by the mapper.

It'd also allow for legacy models to avoid marking fields transient that might need to be serialized in the legacy case, but aren't compatible with Hollow and an alternative is required (for instance you might have `@JsonIgnore` on a Hollow-only and `@HollowIgnore` on a JSON-only field).

For example, this Kotlin code:

```
private val parsedVersion: Version by lazy { Version.valueOf(version) }
```

Results in:
```
java.lang.IllegalArgumentException: Unexpected interface Lazy passed as field.
	at com.netflix.hollow.core.write.objectmapper.HollowObjectTypeMapper.<init>(HollowObjectTypeMapper.java:80)
	at com.netflix.hollow.core.write.objectmapper.HollowObjectMapper.getTypeMapper(HollowObjectMapper.java:137)
	at com.netflix.hollow.core.write.objectmapper.HollowObjectTypeMapper$MappedField.<init>(HollowObjectTypeMapper.java:329)
	at com.netflix.hollow.core.write.objectmapper.HollowObjectTypeMapper$MappedField.<init>(HollowObjectTypeMapper.java:248)
	at com.netflix.hollow.core.write.objectmapper.HollowObjectTypeMapper.<init>(HollowObjectTypeMapper.java:92)
	at com.netflix.hollow.core.write.objectmapper.HollowObjectMapper.getTypeMapper(HollowObjectMapper.java:137)
	at com.netflix.hollow.core.write.objectmapper.HollowObjectTypeMapper$MappedField.<init>(HollowObjectTypeMapper.java:329)
	at com.netflix.hollow.core.write.objectmapper.HollowObjectTypeMapper$MappedField.<init>(HollowObjectTypeMapper.java:248)
	at com.netflix.hollow.core.write.objectmapper.HollowObjectTypeMapper.<init>(HollowObjectTypeMapper.java:92)
	at com.netflix.hollow.core.write.objectmapper.HollowObjectMapper.getTypeMapper(HollowObjectMapper.java:137)
	at com.netflix.hollow.core.write.objectmapper.HollowObjectMapper.getTypeMapper(HollowObjectMapper.java:109)
[...]

PR #316, opened 2018-10-25
Document transient fields (fix for #308)

PaulSandoz:


docs/data-modeling.md
```diff
@@ -162,6 +162,12 @@ public class MaturityRating {
 
 We could have referenced these fields separately.  If we had done so, each `Movie` record, of which there are probably many, would have had to contain two separate references for these fields.  Instead, by recognizing that these fields were associated and pulling them together, space is saved because each `Movie` record now only contains one reference for this data.
 
+### Transient Fields
+
+A transient field is ignored and will not be included in an `Object` Schema.  A transient field is a field declared
+with the `transient` Java keyword or annotated with the `@HollowTransient` annotation.  The latter may be used for
```
akhaku, line 168: At that point maybe it's worth only recommending using `@HollowTransient`?

PaulSandoz: I think it better for developers to leverage the commonly understood `transient` keyword, and leverage the annotation for uncommon cases (e.g. shared data model where transient is not honored or where it cannot be encoded in say a JVM-based language).

PR #322, opened 2018-11-02
AbstractHollowHashIndex should not construct new state on a delta update

PaulSandoz:
The refresh listener should on a snapshot update detach the existing index,
construct a new one, and register it.
On a delta update only the api instance should be updated.  This avoids a
potential memory leak for delta update chains.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/consumer/index/AbstractHollowHashIndex.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/consumer/index/AbstractHollowHashIndex.java
@@ -94,16 +94,17 @@ public abstract class AbstractHollowHashIndex<API> {
     }
 
     private class RefreshListener implements HollowConsumer.RefreshListener {
-        @Override public void deltaUpdateOccurred(HollowAPI api, HollowReadStateEngine stateEngine, long version) throws Exception {
-            reindex(stateEngine, api);
-        }
-
-        @Override public void snapshotUpdateOccurred(HollowAPI api, HollowReadStateEngine stateEngine, long version) throws Exception {
-            reindex(stateEngine, api);
-        }
-
-        private void reindex(HollowReadStateEngine stateEngine, HollowAPI refreshAPI) {
+        @Override
+        public void snapshotUpdateOccurred(HollowAPI refreshAPI, HollowReadStateEngine stateEngine, long version) {
+            idx.detachFromDeltaUpdates();
             idx = new HollowHashIndex(stateEngine, queryType, selectFieldPath, matchFieldPaths);
+            idx.listenForDeltaUpdates();
+
+            api = castAPI(refreshAPI);
+        }
+
+        @Override
+        public void deltaUpdateOccurred(HollowAPI refreshAPI, HollowReadStateEngine stateEngine, long version) {
             api = castAPI(refreshAPI);
         }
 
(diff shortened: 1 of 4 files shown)


hollow/src/main/java/com/netflix/hollow/api/consumer/index/AbstractHollowHashIndex.java
```diff
@@ -94,16 +94,17 @@ public void detachFromDataRefresh() {
     }
 
     private class RefreshListener implements HollowConsumer.RefreshListener {
-        @Override public void deltaUpdateOccurred(HollowAPI api, HollowReadStateEngine stateEngine, long version) throws Exception {
-            reindex(stateEngine, api);
-        }
+        @Override
+        public void snapshotUpdateOccurred(HollowAPI refreshAPI, HollowReadStateEngine stateEngine, long version) {
+            idx.detachFromDeltaUpdates();
+            idx = new HollowHashIndex(stateEngine, queryType, selectFieldPath, matchFieldPaths);
```
akhaku, line 100: Just for my own knowledge - I'm guessing this blocks the refresh thread (that has the write lock) while the hash index builds?

hollow/src/main/java/com/netflix/hollow/api/consumer/index/AbstractHollowHashIndex.java
```diff
@@ -94,16 +94,17 @@ public void detachFromDataRefresh() {
...
-        @Override public void snapshotUpdateOccurred(HollowAPI api, HollowReadStateEngine stateEngine, long version) throws Exception {
-            reindex(stateEngine, api);
+            api = castAPI(refreshAPI);
         }
 
-        private void reindex(HollowReadStateEngine stateEngine, HollowAPI refreshAPI) {
-            idx = new HollowHashIndex(stateEngine, queryType, selectFieldPath, matchFieldPaths);
+        @Override
+        public void deltaUpdateOccurred(HollowAPI refreshAPI, HollowReadStateEngine stateEngine, long version) {
             api = castAPI(refreshAPI);
```
akhaku, line 108: Curious, why don't we need to reindex on delta updates?

hollow/src/test/java/com/netflix/hollow/api/consumer/AbstractHollowHashIndexTests.java
```diff
@@ -0,0 +1,282 @@
...
+        int ordinal = iter.next();
+        while (ordinal != HollowOrdinalIterator.NO_MORE_ORDINALS) {
+            ordinalSet.add(ordinal);
+            ordinal = iter.next();
+        }
+
+        for (int ord : expectedOrdinals) {
+            Assert.assertTrue(ordinalSet.contains(ord));
+        }
+        Assert.assertTrue(ordinalSet.size() == expectedOrdinals.length);
```
akhaku: assertEquals, maybe?

hollow/src/main/java/com/netflix/hollow/api/consumer/index/AbstractHollowHashIndex.java
PaulSandoz: Yes, and so it will be for deltas too internally via the call to `HollowTypeStateListener.endUpdate` event.  For long delta chains the previous code would retain up a bunch of redundant HollowHashIndex instances all listening and doing the same work and possibly contributing to holding the lock longer than necessary.

hollow/src/main/java/com/netflix/hollow/api/consumer/index/AbstractHollowHashIndex.java
PaulSandoz: The existing `HollowHashIndex` will receive updates via the `HollowTypeStateListener` that is attached via `listenForDeltaUpdates`

hollow/src/test/java/com/netflix/hollow/api/consumer/AbstractHollowHashIndexTests.java
PaulSandoz: Ok.

akhaku (approved):

PR #323, opened 2018-11-02
Specify runCycle to return the latests announced version.

PaulSandoz:
Previously if population resulted in no changes a version
was returned that would be the announced version if there had
been changes.
The change in behaviour is consistent with if the producer is not
the primary producer
The version may be consistently checked against consumers and their
current version.  The version reported to event listeners has not
changed.


akhaku (approved):

kineshsatiya (approved):
Should producer have an API to fetch version, like getLastSuccessfulCycleVersion() ?

PR #335, opened 2018-11-11
Maintain legacy behaviour when not using hollow primitives

akhaku:
A bugfix here changed behaviour last release, which causes
problems in legacy usage internally. Revert to old behaviour
for this use case.


hollow/src/test/java/com/netflix/hollow/api/codegen/AbstractHollowAPIGeneratorTest.java
```diff
@@ -43,6 +49,24 @@ void runGenerator(String apiClassName, String packageName, Class<?> clazz,
         HollowCodeGenerationCompileUtil.compileSrcFiles(sourceFolder, clazzFolder);
     }
 
+    void assertNonEmptyFileExists(String relativePath) throws IOException {
+        if (relativePath.startsWith("/")) {
+            throw new IllegalArgumentException("Relative paths should not start with /");
+        }
+        try (BufferedReader br = new BufferedReader(
```
PaulSandoz: Not sure of if wanna check if the file exists and its length > 0 or there is at least one line:
```
File f = new File(sourceFolder + "/" + relativePath);
assertTrue(f.exists() && f.length() > 0L);
```
or
```
try (Stream<String> lines = Files.lines(Paths.get("sourceFolder", "relativePath")) {
  assertTrue(lines.findFirst().isPresent());
}
```

PaulSandoz (approved):

hollow/src/test/java/com/netflix/hollow/api/codegen/AbstractHollowAPIGeneratorTest.java
PaulSandoz: Further, if using a `Stream` or `BufferedReader` then a `FileNotFoundException` will be thrown if the file does not exist so it will not reach the assert.

akhaku: I was going for length > 0 - thanks, will update

PR #336, opened 2018-11-11
Some hollow4 cleanup to take advantage of breaking changes

akhaku:


hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducerListenerV2.java
```diff
@@ -1,47 +0,0 @@
-/*
```
PaulSandoz, line 1: Do you estimate that `HollowProducerListenerV2` is so rarely used we can just opportunistically remove it?

akhaku: It was added relatively recently (I added it) to allow adding onCycleSkipped to HollowProducerListener while we were still on Java 7 and couldn't have default methods on interfaces. I know that this will break our main use case within Netflix, but fixing that is a simple rename. I don't anticipate many people having found it and used it - for the ones that have, they can switch to using the main HollowProducerListener.

PaulSandoz: Ok, from what you say i think we should be fine regarding the impact of that change.

PaulSandoz (approved):

PR #359, opened 2018-12-13
Support the addition of DuplicateDataDetectionValidators for an initialized data model

PaulSandoz:


hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -124,7 +160,56 @@ public ValidationResult onValidate(ReadState readState) {
...
+     * @see HollowProducer#initializeDataModel(Class[])
+     */
+    public static void addValidatorsForSchemaWithPrimaryKey(HollowProducer producer) {
+        producer.getWriteEngine().getOrderedTypeStates().stream()
+                .filter(ts -> ts.getSchema().getSchemaType() == SchemaType.OBJECT)
+                .map(ts -> (HollowObjectSchema) ts.getSchema())
+                .filter(hos -> hos.getPrimaryKey() != null)
+                .map(HollowObjectSchema::getPrimaryKey)
+                .forEach(k -> producer.addListener(new DuplicateDataDetectionValidator(k.getType())));
+    }
```
akhaku, line 215: wow 💯

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -124,7 +160,56 @@ public ValidationResult onValidate(ReadState readState) {
...
+     * {@code DuplicateDataDetectionValidator} validator is not already registered for the same primary key type name).
+     *
+     * @param producer the producer
+     * @apiNote This method registers a {@code DuplicateDataDetectionValidator} validator with only the primary key type
+     * name and not, in addition, the primary key fields.  This is to ensure, for the common case, duplicate listeners
+     * are not registered by this method if listeners with the same type names were explicitly registered when
+     * building the producer.
+     * @see HollowProducer#initializeDataModel(Class[])
+     */
+    public static void addValidatorsForSchemaWithPrimaryKey(HollowProducer producer) {
```
akhaku, line 208: Any value in doing this with the HollowProducer.Builder instead of the HollowProducer?

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -25,20 +26,24 @@
...
  * <p>
- * Note that if an object only has primary key fields then the duplicates are deduped by Hollow and will not result
- * in failing this validator. Ex: If Movie type has int id and String name and if primary key is also int id and String
- * name this validator will not fail as the two objects are deduped.
+ * The primary key may be declared on a data model class using the {@link HollowPrimaryKey} annotation, may be
+ * declared more directly using {@link PrimaryKey} with {@link HollowObjectSchema}, or declared explicitly when
+ * instantiating this validator.
  *
- * @author lkanchanapalli {@literal<lavanya65@yahoo.com>}
+ * @apiNote If all a record's fields are associated with a primary key then any duplicates added during the population
```
akhaku: took me a second to realize what you meant here, maybe add something like `(i.e. the object has no fields that are not in the primary key)`

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -60,16 +65,47 @@
...
+
+        this.dataTypeName = HollowTypeMapper.getDefaultTypeName(dataType);
+        this.fieldPathNames = null;
+    }
+
+    /**
+     * Creates a validator to detect duplicate records of the type that corresponds to the given data type name.
+     * <p>
+     * The validator will fail, when {@link #onValidate validating}, if a schema with a primary key definition does not
+     * exists for the given data type name.
```
akhaku: typo: I think you mean `exist` (same typo below)

akhaku (approved):

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
PaulSandoz: The `HollowProducer` has the type state information to query the schema (either from initializing the data model or after the first cycle). When building this information is not available.  We could enhance the builder to accept a data model and express that producer should be initialized and validated for that model before returning the built producer? is that what you are getting at?

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
PaulSandoz: We could use `flatMap` in some cases to compress the `map.filter` but without `Stream.ofNullable` (added in JDK 9) it's a little painful to express.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
PaulSandoz: Yes! fixed.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
PaulSandoz:
```
 * @apiNote If all a record's fields are associated with a primary key (in other words, the record has no
 * fields that are not associated the primary key) then any duplicates added during the population
 * stage will be naturally de-duped resulting in just one record present.  Therefore a validator created for such a
 * record's type will never fail.
```

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
akhaku: Argh, good point, we can't do it with the Builder. I was hoping to remove (or at least stop adding) ways to modify the HollowProducer after building it, but there are probably some cases where it's difficult to do that. Refresh listeners (eg for indexes) are another good example of that. Technically there isn't much reason to have initializing the data model be required, so what you did makes sense.

PR #360, opened 2018-12-13
Use CompletableFuture instead of wait/notify to wait for task

akhaku:
With the existing await/notify implementation, there are cases where JsonArrayChunkerInputSegment#fill throwing an exception can result in notify never being called, which means we have a stuck thread.
Switch to using CompletableFuture to simplify the queueing logic.

### DIFF
--- a/hollow-jsonadapter/src/main/java/com/netflix/hollow/jsonadapter/chunker/JsonArrayChunker.java
+++ b/hollow-jsonadapter/src/main/java/com/netflix/hollow/jsonadapter/chunker/JsonArrayChunker.java
@@ -19,8 +19,10 @@ package com.netflix.hollow.jsonadapter.chunker;
 
 import java.io.IOException;
 import java.io.Reader;
-import java.util.LinkedList;
+import java.util.ArrayDeque;
 import java.util.Queue;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.Executor;
 
 public class JsonArrayChunker {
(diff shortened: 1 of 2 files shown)


PaulSandoz:
Ah i get it now where the parallelism comes from. `fillOneSegment` adds a new segment to the end of the queue and submits work on that segment to the executer. The `initialize` fills up the queue. `nextSegment` opportunistically adds something to the end of the queue, takes the *first* item from the queue and waits for it to complete. So there is potentially a reasonable degree of parallelism. We could improve this by having a queue of `CompletableFuture<JsonArrayChunkerInputSegment>` then there is no need for the `wait`/`notify` dance and errors can be propagated.

PaulSandoz:
Good progress but I think you can simplify by doing the following:

```
    private final Queue<CompletableFuture<JsonArrayChunkerInputSegment>> bufferSegments;
```

A segment can be added as follows:

```
    private void fillOneSegment() throws IOException {
        final JsonArrayChunkerInputSegment seg = new JsonArrayChunkerInputSegment(segmentLength);
        eofReached = seg.fill(reader);

        CompletableFuture<JsonArrayChunkerInputSegment> cfseg = CompletableFuture.supplyAsync(
                () -> {
                    seg.findSpecialCharacterOffsets();
                    return seg;
                },
                executor);
        bufferSegments.add(cfseg);
    }
```

And a segment can be waited on as follows:

```
    private boolean nextSegment() throws InterruptedException, IOException {
        if (bufferSegments.size() == 0) {
            return false;
        }

        if (!eofReached) {
            fillOneSegment();
        }

        currentSegmentStartOffset += segmentLength;
        currentSegment = bufferSegments.remove().join();
        return true;
    }
```

`JsonArrayChunkerInputSegment.waitForDefinedOffsets` can be removed.  Thus all the asynchronous responsibility is kept within `JsonArrayChunker`, `JsonArrayChunkerInputSegment` is non the wiser as to what is going on, and we transparently handle all failures from `findSpecialCharacterOffsets`.

Extra bonus points for extracting the underlying exception from the `CompletionException` thrown by `join`, or change to `get` and extract from `ExecutionException` rather than re-throw it.

PR #362, opened 2018-12-19
Unify field paths parsing.

PaulSandoz:
Fields paths are predominantly utilized in three places:
1.  For primary key indexes;
2. Hash key indexes; and
3. Prefix indexes.

Each have subtly different semantics.  The processing of field paths and error reporting in each case has been unified without changing the semantics (with one exception, for case 3 auto-traversing collections is now disallowed).  New functionality has been added to process a field path into a structured form that can be reasoned about in those three cases but also for future cases such as a higher level type safe API for index querying (to follow soon), or perhaps by an IDE highlighting errors in field path declarations in, say, a hollow schema language.

Existing tests are currently relied upon.  Documentation on `com.netflix.hollow.core.index.FieldPaths` needs to be expanded later on to specify the field path semantics, formally specifying what is presented in the `hollow.how` documentation.

`com.netflix.hollow.core.index.FieldPath` has been made package private.  It was only used by `HollowPrefixIndex` and has semantics that are not applicable to the other cases.  It provides additional useful functionality, value extraction, that can be separated out from structured field paths.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/index/FieldPath.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/index/FieldPath.java
@@ -47,7 +47,7 @@ import java.util.List;
  * <p>
  * This class has a convenience method to find values following the field path and given a start field position.
  */
-public class FieldPath {
+class FieldPath {
 
     private final String fieldPath;
     private final HollowDataAccess hollowDataAccess;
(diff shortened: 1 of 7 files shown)


kineshsatiya (approved):
Nice! Neat job on unifying parsing of the paths.

akhaku (approved):

PR #364, opened 2019-01-03
Exporting refresh metrics via listeners

Sunjeet:
Highlights:
- Consumer refresh metrics computed in Hollow, can be attached as a listener.
- Added performance cost to Hollow is trivial if a metrics listener isn't registered (basically "transitionsPlanned" event is called on all listeners but no implementations will be bound).
- Introduced "transitionsPlanned" event that gets triggered after refresh start and once an update plan has been created but before it is executed, to provide early visibility into what is intended.
- Add notion of "overallRefreshType" only for metrics purposes which can be snapshot(1+ snapshots, 0+deltas) , delta(1+deltas) or reversedelta(1+ reversedeltas).
- Distinguish between initial vs. subsequent load type
- Additional details of update plan are now available like how many transitions and sequence of transitions for eg. [snapshot, delta, delta]. Also computes how many transitions were successful at the end of refresh.
- Remove tracking of consecutiveSuccesses metric
- last successful refresh time metric is now optional, which protects against the initial load failure case

Testing:
Unit tests and deployed to a cinderhollowhistoryexplorer in Test env. I was able to test backwards compatibility with existing metrics.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,8 +2,16 @@ buildscript {
   repositories { jcenter() }
 }
 
+plugins {
+  id 'nebula.netflixoss' version '6.0.3'
+}
+
 subprojects {
+  apply plugin: 'nebula.netflixoss'
   apply plugin: 'checkstyle'
+
+  sourceCompatibility = 1.8
+  targetCompatibility = 1.8
 
   tasks.withType(JavaCompile) {
     sourceCompatibility = 1.8
@@ -18,7 +26,6 @@ subprojects {
   }
 
   group = 'com.netflix.hollow'
-
   checkstyle {
     configFile = file("$rootProject.projectDir/config/checkstyle/checkstyle.xml")
   }
--- a/build.gradle
+++ b/build.gradle
@@ -2,16 +2,8 @@ buildscript {
   repositories { jcenter() }
 }
 
-plugins {
-  id 'nebula.netflixoss' version '6.0.3'
-}
-
 subprojects {
-  apply plugin: 'nebula.netflixoss'
   apply plugin: 'checkstyle'
-
-  sourceCompatibility = 1.8
-  targetCompatibility = 1.8
 
   tasks.withType(JavaCompile) {
     sourceCompatibility = 1.8
@@ -26,6 +18,7 @@ subprojects {
   }
 
   group = 'com.netflix.hollow'
+
   checkstyle {
     configFile = file("$rootProject.projectDir/config/checkstyle/checkstyle.xml")
   }
--- a/hollow/build.gradle
+++ b/hollow/build.gradle
@@ -1,10 +1,12 @@
 apply plugin: 'java'
- 
+
 dependencies {
     testCompile project(":hollow-test")
     testCompile 'junit:junit:4.11'
     testCompile 'org.mockito:mockito-core:2.15.0'
     testCompile 'com.google.code.findbugs:findbugs:3.0.1'
+
+    // compileOnly 'org.projectlombok:lombok:1.16.12' // TODO: get team's buy-in for using Lombok
 }
 
 // quiet warnings about sun.misc.Unsafe
--- a/hollow/build.gradle
+++ b/hollow/build.gradle
@@ -5,8 +5,6 @@ dependencies {
     testCompile 'junit:junit:4.11'
     testCompile 'org.mockito:mockito-core:2.15.0'
     testCompile 'com.google.code.findbugs:findbugs:3.0.1'
-
-    // compileOnly 'org.projectlombok:lombok:1.16.12' // TODO: get team's buy-in for using Lombok
 }
 
 // quiet warnings about sun.misc.Unsafe
--- a/hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
@@ -413,6 +413,7 @@ public class HollowConsumer {
 
         private final long fromVersion;
         private final long toVersion;
+        private final BlobType blobType;
 
         /**
          * Instantiate a snapshot to a specified data state version.
@@ -432,6 +433,13 @@ public class HollowConsumer {
         public Blob(long fromVersion, long toVersion) {
             this.fromVersion = fromVersion;
             this.toVersion = toVersion;
+
+            if (this.isSnapshot())
+                this.blobType = BlobType.SNAPSHOT;
+            else if (this.isReverseDelta())
+                this.blobType = BlobType.REVERSE_DELTA;
+            else
+                this.blobType = BlobType.DELTA;
         }
 
         /**
@@ -444,6 +452,24 @@ public class HollowConsumer {
          * @throws IOException if the input stream to the blob cannot be obtained
          */
         public abstract InputStream getInputStream() throws IOException;
+
+        /**
+         * Blobs can be of types {@code SNAPSHOT}, {@code DELTA} or {@code REVERSE_DELTA}.
+         */
+        public enum BlobType {
+            SNAPSHOT("snapshot"),
+            DELTA("delta"),
+            REVERSE_DELTA("reversedelta");
+
+            private final String type;
+            BlobType(String type) {
+                this.type = type;
+            }
+
+            public String getType() {
+                return this.type;
+            }
+        }
 
         public boolean isSnapshot() {
             return fromVersion == HollowConstants.VERSION_NONE;
@@ -463,6 +489,10 @@ public class HollowConsumer {
 
         public long getToVersion() {
             return toVersion;
+        }
+
+        public BlobType getBlobType() {
+            return blobType;
         }
     }
 
@@ -743,11 +773,27 @@ public class HollowConsumer {
          */
         void deltaApplied(HollowAPI api, HollowReadStateEngine stateEngine, long version) throws Exception;
 
+        /**
+         * Called after refresh started and update plan has been initialized, but before the update plan starts executing.
+         * It is called only once per update plan (and thus only once per consumer refresh). Exposes details of the
+         * update plan.
+         *
+         * @param beforeVersion The version when refresh started
+         * @param requestedVersion The intended version at the end of the refresh
+         * @param isSnapshotPlan Indicates whether the refresh involves a snapshot transition
+         * @param transitionSequence List of transitions comprising the refresh
+         */
+        void transitionsPlanned(long beforeVersion, long requestedVersion, boolean isSnapshotPlan, List<HollowConsumer.Blob.BlobType> transitionSequence);
     }
 
     public static class AbstractRefreshListener implements TransitionAwareRefreshListener {
         @Override
         public void refreshStarted(long currentVersion, long requestedVersion) {
+            // no-op
+        }
+
+        @Override
+        public void transitionsPlanned(long beforeVersion, long requestedVersion, boolean isSnapshotPlan, List<HollowConsumer.Blob.BlobType> transitionSequence) {
             // no-op
         }
 
--- a/hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
@@ -777,13 +777,14 @@ public class HollowConsumer {
          * Called after refresh started and update plan has been initialized, but before the update plan starts executing.
          * It is called only once per update plan (and thus only once per consumer refresh). Exposes details of the
          * update plan.
+         * @implSpec The default implementation provided does nothing.
          *
          * @param beforeVersion The version when refresh started
-         * @param requestedVersion The intended version at the end of the refresh
+         * @param desiredVersion The version that the consumer refresh tries update to, even though it might not be attainable eg. HollowConstants.VERSION_LATEST
          * @param isSnapshotPlan Indicates whether the refresh involves a snapshot transition
          * @param transitionSequence List of transitions comprising the refresh
          */
-        void transitionsPlanned(long beforeVersion, long requestedVersion, boolean isSnapshotPlan, List<HollowConsumer.Blob.BlobType> transitionSequence);
+        default void transitionsPlanned(long beforeVersion, long desiredVersion, boolean isSnapshotPlan, List<HollowConsumer.Blob.BlobType> transitionSequence) {}
     }
 
     public static class AbstractRefreshListener implements TransitionAwareRefreshListener {
@@ -793,7 +794,7 @@ public class HollowConsumer {
         }
 
         @Override
-        public void transitionsPlanned(long beforeVersion, long requestedVersion, boolean isSnapshotPlan, List<HollowConsumer.Blob.BlobType> transitionSequence) {
+        public void transitionsPlanned(long beforeVersion, long desiredVersion, boolean isSnapshotPlan, List<HollowConsumer.Blob.BlobType> transitionSequence) {
             // no-op
         }
 
--- a/hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
@@ -1,10 +1,10 @@
 package com.netflix.hollow.api.consumer.metrics;
 
-import static com.netflix.hollow.core.util.Versions.maybeVersion;
+import static com.netflix.hollow.core.HollowConstants.VERSION_NONE;
 
+import com.netflix.hollow.api.consumer.HollowConsumer;
 import com.netflix.hollow.api.consumer.HollowConsumer.AbstractRefreshListener;
 import com.netflix.hollow.api.consumer.HollowConsumer.Blob.BlobType;
-import com.netflix.hollow.api.consumer.HollowConsumer;
 import com.netflix.hollow.api.custom.HollowAPI;
 import com.netflix.hollow.core.read.engine.HollowReadStateEngine;
 import java.util.List;
@@ -22,7 +22,7 @@ import java.util.concurrent.TimeUnit;
 public abstract class AbstractRefreshMetricsListener extends AbstractRefreshListener implements RefreshMetricsReporting {
 
     private OptionalLong lastRefreshTimeNanoOptional;
-    private long refreshStart;
+    private long refreshStartTimeNano;
     private long consecutiveFailures;
     private ConsumerRefreshMetrics.Builder refreshMetricsBuilder;
     private BlobType overallRefreshType;    // Indicates whether the overall refresh (that could comprise of multiple transitions)
@@ -40,11 +40,11 @@ public abstract class AbstractRefreshMetricsListener extends AbstractRefreshList
     @Override
     public final void refreshStarted(long currentVersion, long requestedVersion) {
 
-        refreshStart = System.nanoTime();
+        refreshStartTimeNano = System.nanoTime();
         updatePlanDetails = new UpdatePlanDetails();
 
         refreshMetricsBuilder = new ConsumerRefreshMetrics.Builder();
-        refreshMetricsBuilder.setIsInitialLoad(!maybeVersion(currentVersion).isPresent());
+        refreshMetricsBuilder.setIsInitialLoad(currentVersion == VERSION_NONE);
     }
 
     /**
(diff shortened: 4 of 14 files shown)


build.gradle
```diff
@@ -2,17 +2,9 @@ buildscript {
   repositories { jcenter() }
 }
 
-plugins {
```
ghost: Don't think you want this committed.

build.gradle
```diff
@@ -2,17 +2,9 @@ buildscript {
...
-  id 'nebula.netflixoss' version '6.0.3'
-}
-
 subprojects {
-  apply plugin: 'nebula.netflixoss'
```
ghost: Don't think you want this committed.

build.gradle
```diff
@@ -2,17 +2,9 @@ buildscript {
...
   apply plugin: 'checkstyle'
 
-  sourceCompatibility = 1.8
```
ghost: Don't think you want this committed.

hollow/build.gradle
```diff
@@ -1,10 +1,12 @@
 apply plugin: 'java'
- 
+
 dependencies {
     testCompile project(":hollow-test")
     testCompile 'junit:junit:4.11'
     testCompile 'org.mockito:mockito-core:2.15.0'
     testCompile 'com.google.code.findbugs:findbugs:3.0.1'
+
+    // compileOnly 'org.projectlombok:lombok:1.16.12' // TODO: get team's buy-in for using Lombok
```
ghost: To date we've avoided 3rd party dependencies for the core of Hollow to avoid dependency hell. We're a little less strict for tests. Maybe we can be a little less strict for `compileOnly`.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
```diff
@@ -0,0 +1,178 @@
...
+    private long refreshLoadCount;
+
+    // Encapsulate the refresh metrics so that more metrics/properties can be added in future wihtout breaking existing clients.
+    public static class ConsumerRefreshMetrics {
+        private long refreshDurationMillis;
+        private ConsumerRefreshStatus refreshStatus;
+        private ConsumerRefreshTransitionType refreshTransitionType;
+        private ConsumerRefreshLoadType refreshLoadType;
+
+        public ConsumerRefreshMetrics(long refreshDurationMillis, ConsumerRefreshStatus refreshStatus, ConsumerRefreshTransitionType refreshTransitionType, ConsumerRefreshLoadType refreshLoadType) {
```
ghost: I think we want `ConsumerRefreshMetrics` and `ConsumerStateUpdate` (in `consumer.proto`) to be more similar than different. That could mean that `ConsumerStateUpdate` has to change or vice versa.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
```diff
@@ -0,0 +1,178 @@
...
+            this.refreshLoadType = Objects.requireNonNull(refreshLoadType);
+        }
+
+        public long getRefreshDurationMillis() {
+            return refreshDurationMillis;
+        }
+        public ConsumerRefreshStatus getRefreshStatus() {
+            return refreshStatus;
+        }
+        public ConsumerRefreshTransitionType getRefreshTransitionType() {
```
ghost: A particular refresh could consist of:

* 1 snapshot
* N deltas
* N reverse deltas
* 1 snapshot and N deltas

For the refreshes that have multiple transitions (2+ deltas or snapshot & 1+ deltas) I wasn't sure how to distill that down to a single "snapshot occurred" or "delta occurred". So in `ConsumerStateUpdate` I punted and just modeled the N *transitions* directly and let the calling code decide what to do.

An individual transition is 1-to-1 with a blob type.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
```diff
@@ -0,0 +1,178 @@
+package com.netflix.hollow.api.consumer.metrics;
+
+import com.netflix.hollow.api.consumer.HollowConsumer;
+import com.netflix.hollow.api.custom.HollowAPI;
+import com.netflix.hollow.core.read.engine.HollowReadStateEngine;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Objects;
+
+public class AbstractRefreshMetricsListener extends HollowConsumer.AbstractRefreshListener  {
```
ghost: I have some thoughts on "Refresh" as the verb/noun we use and why (for better or worse) I started to introduce different terminology in the Cinder Dashboard.

I landed on "consumer state update" which I'm not in love with, but is consistent with some other terminology in Hollow ("states"). The problem I have with "refresh" is that it implies an already loaded state and so it can be confusing that the initial load of data is also a refresh.

That said, refresh terminology is already baked into our class names. So the decision should be if we're committed to eventually renaming things like `RefreshListener`; otherwise introducing a *consumer state update* as a synonym for *refresh* just makes things worse.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
```diff
@@ -0,0 +1,178 @@
...
+        }
+        public ConsumerRefreshTransitionType getRefreshTransitionType() {
+            return refreshTransitionType;
+        }
+        public ConsumerRefreshLoadType getRefreshLoadType() {
+            return refreshLoadType;
+        }
+    }
+
+    public enum ConsumerRefreshTransitionType {
```
ghost: Reverse delta is also a transition type. I assume this was just an omission. However, if the intention is to group deltas and reverse deltas under the "delta" type, then I think we should avoid that.

Sometimes I wish we'd coined them both as "forward delta" and "reverse delta" thoughout the API and documentation.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
```diff
@@ -0,0 +1,178 @@
...
+        Delta("delta");             // refresh applied a delta transition
+
+        final private String type;
+        ConsumerRefreshTransitionType(String type) {
+            this.type = type;
+        }
+    }
+
+    public enum ConsumerRefreshLoadType {
+        Init("init"),               // consumer's first-time initialized with a snapshot
```
ghost: Maybe the longer "initial"? In `HollowConsumer` I recently introduced "initial load" nomenclature (by introducing `getInitialLoad()`).

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
```diff
@@ -0,0 +1,178 @@
...
+
+public class AbstractRefreshMetricsListener extends HollowConsumer.AbstractRefreshListener  {
+
+    private long consecutiveSuccesses;
+    private long consecutiveFailures;
+
+    private Instant lastRefreshInstant;
+    private Instant refreshStart;
+
+    private ConsumerRefreshStatus refreshStatus;                    // success or failure
```
ghost: I wonder if enums are overkill for things we can be reasonably confident will remain binary (e.g. success/fail, initial/subsequent). I've definitely gone the enum route myself in the past. I think @PaulSandoz, specifically in the case of a status enum in the producer, disliked how that went from `SUCCESS / FAIL` to `SUCCESS / FAIL / SKIP`; his preference was to stick with binary PASS / FAIL, treat skipping as passing and expose skipping through a different channel.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
```diff
@@ -0,0 +1,178 @@
...
+
+        final private String status;
+        ConsumerRefreshStatus(String status) {
+            this.status = status;
+        }
+    }
+
+
+    public AbstractRefreshMetricsListener() {
+        lastRefreshInstant = Instant.now(); // TODO: Is this an acceptable value for the first transition?
```
ghost: > Is this an acceptable value for the first transition?

We probably need a sentinel value. Since this is used to report a "time since last refresh" metric, until the initial load we probably shouldn't report that metric at all. For our use, we'd probably want this to be `null` and only when non-null would we start sending the metric.

Looked at another way, this metric should only report zero when it successfully completed an update; viewing a visualization of this metric the troughs should correspond to initial load or subsequent refreshes.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
```diff
@@ -0,0 +1,178 @@
...
+        refreshLoadCount = 0l;
+        consecutiveSuccesses = 0l;
+        consecutiveFailures = 0l;
+        refreshStatus = ConsumerRefreshStatus.Success;  // TODO: Ok to initialize as success for first refresh?
```
ghost: In our internal dashboard, we support showing an initial load that fails to succeed. The load type isn't necessarily the load that happened...until the moment of success it's the load/refresh that is being attempted.

So we either want to omit this value from being reported until success or failure happens, or we need to model a 3rd indeterminate state. I think I'm more in favor of the former.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
```diff
@@ -0,0 +1,178 @@
...
+        refreshLoadCount ++;
+
+        if (refreshLoadCount > 1) {
+            refreshLoadType = ConsumerRefreshLoadType.Subsequent;
+        }
+    }
+
+    @Override
+    public final void snapshotUpdateOccurred(HollowAPI refreshAPI, HollowReadStateEngine stateEngine, long version) {
+        refreshTransitionType = ConsumerRefreshTransitionType.Snapshot;
```
ghost: For the case where a consumer, in a single refresh, first applies a snapshot blob then applies a delta blob, we'll end up reporting that as a delta refresh/update once line 158 has been called.

If we're using a sampling metrics reporter (like the Specator library provides) then depending on when we sample these metrics (and assuming we're including transition type as a tag on a metric) then we might see one sample where the tag is snapshot then the subsequent sample the tag would change to delta. For example, if we report a timer metric for the entire load duration and tag that timer with transition type, then partway through that timer the transition might switch from snapshot to delta.

ghost: This probably illustrates a fundamental gap in our current refresh listener API. There's no way for the listener to know at the onset what the update plan has in store for it, even though Hollow determines the update plan early on, long before attempting to download and apply each transition.

build.gradle
Sunjeet: That was the intention. It's one of the loose ends that I mentioned that need to be tied up. I was trying to strike a balance between getting some early feedback on the way I've structured the metrics code and not doing repetitive cleanup because there will be more iterations. I'll tinker around with gradle to see if I can move this into a gitignore'd build.local.gradle (or do you think it's worth looking into the core nebula version conflict issue at this point?).

hollow/build.gradle
Sunjeet: Fair enough, I don't feel too strongly about introducing lombok here. Dropping it for now.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
Sunjeet: Oh! I missed ConsumerStateUpdate altogether! I'll look into that, and convergence sounds great.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
Sunjeet: Hmm.. I didn't think deltaUpdateOccurred() would be called on a "snapshot + N deltas" update plan because of https://stash.corp.netflix.com/projects/GM/repos/hollow/browse/hollow/src/main/java/com/netflix/hollow/api/client/HollowDataHolder.java#170 

That's not true for deltaApplied() though, which is the one we're watching in KeystoneConsumerListener. I do agree about not knowing snapshot-only vs. snapshot+delta(s) vs. delta(s) at the onset of refreshStart (and therefore at refreshFailure) and that might be something to look into adding.

I think we need to talk this one out as I might be missing more here than I thought. Look fwd to our in person chat.

Sunjeet: I'd also like to understand a bit more about the need for sampling here.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
Sunjeet: That sounds like a good strategy. 

On one hand, it feels like having that finer grain visibility is useful, and on the other hand, since individual _transitions_ in a multi-_transition_ update plan don't provide fresh data to the consumer (for the lack of a better description) it's useful to also look at the refresh as a whole.

Currently, snapshotApplied() and deltaApplied() give good visibility on the individual _transitions_ , but the behavior of snapshotUpdateOccurred() and deltaUpdateOccurred() is-

For a single refresh involving snapshot only or snapshot + n deltas,
snapshotUpdateOccurred() is currently called once at the end.
For a single refresh involving n deltas,
deltaUpdateOccurred() is currently called once after every delta transition.

Re-working deltaUpdateOccurred() might provide a solution here, but not sure what other side-effects that might lead to. Lets chat about this further to validate that my understand is correct, and get your thoughts on it.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
Sunjeet: I think that might have something to do with the fact that I didn't come across ConsumerStateUpdate until you mentioned it in one of the PR comments :P Your comment makes perfect sense to me, and I'll look into this further, but worried about the blast radius of renaming refresh to state everywhere.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
Sunjeet: That was a miss. Adding.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
Sunjeet: Sure!

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
Sunjeet: sounds good

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
Sunjeet: ACK

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
Sunjeet: Need to chat about this one

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
```diff
@@ -445,6 +453,24 @@ public Blob(long fromVersion, long toVersion) {
          */
         public abstract InputStream getInputStream() throws IOException;
 
+        /**
+         * Blobs can be of types {@code SNAPSHOT}, {@code DELTA} or {@code REVERSE_DELTA}.
+         */
+        public enum BlobType {
+            SNAPSHOT("snapshot"),
```
PaulSandoz, line 460: Do you need a `String` field whose value is the lower case of the enum value?

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
Sunjeet: Looks a bit more like ConsumerStateUpdate now, but not entirely (missing version, announcedVersion, pinnedVersion, and pinned to limit scope)

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
Sunjeet: overallRefreshType now embodies 3 types-
"snapshot", for 1 snapshot or 1 snapshot + N deltas
"delta", for N deltas
"reversedelta" for N reverse deltas
As per your comment, I didn't want to make this a standard definition for an update plan, so it's evaluated in the Metrics listener. Thought it might be okay to bias with this definition for refresh metrics purposes.

I also now have a list of transition types comprising a refresh, for eg, [SNAPSHOT, DELTA, DELTA] that might be useful for logging for failure scenarios. More detail can be added later, thought we could start with this much.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
Sunjeet: Decided to keep the current terminology for now to limit the scope of this change. I like the idea of calling it state update, but renaming refresh to state update has a wider blast radius.  I think it would be better suited for a different PR.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
Sunjeet: Success/failure will be tagged on the duration metric at refresh end, and lastRefreshTimeNanoOptional is.. well.. optional :)

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
Sunjeet: ConsumerRefreshMetrics now has overallRefreshType set on the duration metric (it's snapshot if there is a snapshot transition in the refresh's update plan). The duration metric will continue to be reported as a Timer as it is now (as opposed to Long Task Timer or Gauge). Details of the refresh are exposed in ConsumerRefreshMetrics's UpdatePlanDetails for logging. This will be more explicit in the Cinder-side PR.

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
Sunjeet: Discussed offline, leaving as-is for consistency.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/ConsumerRefreshMetrics.java
```diff
@@ -0,0 +1,108 @@
...
+        return consecutiveFailures;
+    }
+    public OptionalLong getRefreshSuccessAgeMillisOptional() {
+        return refreshSuccessAgeMillisOptional;
+    }
+    public long getRefreshEndTimeNano() {
+        return refreshEndTimeNano;
+    }
+
+    private ConsumerRefreshMetrics(long durationMillis, boolean isRefreshSuccess, boolean isInitialLoad, BlobType overallRefreshType,
```
akhaku: One pattern I've seen used to great effect is to have the constructor take in the Builder and set fields based on the builder fields.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/UpdatePlanDetails.java
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * A class that contains details of the update plan that may be useful to report as metrics or logs. These details are computed
+ * during execution of the update plan {@code HollowClientUpdater} to {@code TransitionAwareRefreshListener}s.
+ */
+public class UpdatePlanDetails {
+
+    long beforeVersion;
+    long desiredVersion;
+    int numTransitions; // how many transitions in this update Plan
```
akhaku: is this needed - is it sufficient to use the length of the transitionSequence?

Sunjeet: that works, dropping the extra field

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/ConsumerRefreshMetrics.java
Sunjeet: Fits well here, thanks.

Issue #365, opened 2019-01-03
HollowIncrementalProducer sometimes does not update/remove records


kpieprzak:
Hi,
I use Hollow to keep data that changes pretty often. I use `HollowIncrementalProducer` and during one run cycle I update thousands of objects. Recently I found out that there are primary keys for which Hollow keeps multiple objects. Moreover after each run cycle the number of such duplicated objects increase.

I have made some digging and I found that there is an issue in `HollowIncrementalCyclePopulator`. It marks records to remove using `BitSet` but it does this using multiple threads. As `BitSet` is not thread safe it happens that some random records are not correctly set to remove. And because of that after run cycle some records may **not be removed** or may **be duplicated** (in case of update).

If you use only one thread everything works fine. Number of threads used to perform this action is determined by number of available processors so I can imagine that on some machines this issue may never happen. However on machines with multiple processors it happens always when you modify a lot of objects during one run cycle and it happens sometimes when you modify only few objects.

There are tests in `HollowIncrementalProducerTest` that are able to discover this bug but they pass most of the time because they are too small (they modify/remove less than 10 objects). For example I have run test `addAndDeleteCollectionsInParallel` 100 times and it failed because of duplicated objects only twice (in fact it failed other 10 times because of bug in test - it modifies one object twice with different values during one run cycle and expects that one of the value will always win but it can not be guaranteed as it tests parallel execution of `addOrModify`).

PR #366, opened 2019-01-03
Fix multithreading issue in incremental cycle populator

kpieprzak:
It is proposition of fix for issue #365

### DIFF
--- a/hollow/src/test/java/com/netflix/hollow/api/producer/HollowIncrementalProducerMultithreadingTest.java
+++ b/hollow/src/test/java/com/netflix/hollow/api/producer/HollowIncrementalProducerMultithreadingTest.java
@@ -1,0 +1,135 @@
+package com.netflix.hollow.api.producer;
+
+import com.netflix.hollow.api.consumer.HollowConsumer;
+import com.netflix.hollow.api.consumer.InMemoryBlobStore;
+import com.netflix.hollow.api.objects.generic.GenericHollowObject;
+import com.netflix.hollow.api.producer.fs.HollowInMemoryBlobStager;
+import com.netflix.hollow.core.index.HollowPrimaryKeyIndex;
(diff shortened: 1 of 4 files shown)


hollow/src/test/java/com/netflix/hollow/api/producer/HollowIncrementalProducerMultithreadingTest.java
```diff
@@ -0,0 +1,135 @@
...
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+public class HollowIncrementalProducerMultithreadingTest {
+
+    private static final int ELEMENTS = 20000;
+    private static final double THREADS = 50.0d;
```
PaulSandoz: IIUC running on a machine with say 32 hardware threads will potentially create a lot of threads for not much benefit as only 32 threads will be running.

Can you please reduce to say 4 (that is the degree of over-provisioning fork join does to compensate for lumpy loads) but run each test within a loop for a number of iterations, thereby increasing the likely hood that a race condition will be encountered.

PaulSandoz:
Thank you for doing this. I question the approach to parallelism within `HollowIncrementalCyclePopulator` but we can revise that later on (perhaps using parallel streams that will likely have better characteristics with regards to load distribution, and/or thinking about the axis of parallelism, such as the type state thereby not requiring a thread safe bit set).

dkoszewnik (approved):
Thank you very much @kpieprzak

hollow/src/test/java/com/netflix/hollow/api/producer/HollowIncrementalProducerMultithreadingTest.java
kpieprzak: Done.

PaulSandoz (approved):

PR #367, opened 2019-01-04
Made changes in announced.version atomic 

nkostiv:
Avoid situation when announced.version file is already created, but is empty. This leads to NumberFormatException in HollowFilesystemAnnouncementWatcher.readLatestVersion.

We faced with such behaviour multiple times in our services.


PaulSandoz (approved):
Thanks!

PR #371, opened 2019-01-09
Type safe primary key index.

PaulSandoz:
Replaces the generated API for supporting primary key indexing with a generic and type safe API for building a `PrimaryKeyIndex`.   Refactored code in `HashSelectIndex` to be reused by `PrimaryKeyIndex`.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/consumer/index/HashIndex.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/consumer/index/HashIndex.java
@@ -62,6 +62,7 @@ public final class HashIndex<T extends HollowRecord, Q> extends HashIndexSelect<
      * @param consumer the consumer containing instances of the given root type
      * @param rootType the root type to match and select from
      * @param <T> the root type
+     * @return a builder
      */
     public static <T extends HollowRecord> Builder<T> from(HollowConsumer consumer, Class<T> rootType) {
         Objects.requireNonNull(consumer);
@@ -71,10 +72,6 @@ public final class HashIndex<T extends HollowRecord, Q> extends HashIndexSelect<
 
     /**
      * The builder of a {@link HashIndex} or a {@link HashIndexSelect}.
-     * <p>
-     * The default configuration specifies, unless specified differently by an initiating builder method, that
-     * the hash index listen to {@code HollowConsumer} version updates as if by a call to {@link #listenToDataRefresh}
-     * with a value of {@code true}.
      *
      * @param <T> the root type
      */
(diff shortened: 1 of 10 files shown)


hollow/src/main/java/com/netflix/hollow/api/consumer/index/PrimaryKeyIndex.java
```diff
@@ -0,0 +1,319 @@
...
+import java.util.Objects;
+import java.util.stream.Stream;
+
+/**
+ * A type safe index for indexing with a primary, or unique, key.
+ *
+ * @param <T> the primary type
+ * @param <Q> the key type
+ */
+public class PrimaryKeyIndex<T extends HollowObject, Q> {
```
ghost: Should this be `PrimaryKeyIndex` or `UniqueKeyIndex`?

@davesu believes that `PrimaryKeyIndex` is a misleading name. It implies a strong connection to the `@HollowPrimaryKey` annotation, of which there can be one per type. Whereas a type could have more than one unique key index (and also not have a primary key defined at all).

Options I can think of:

* **as-is in this PR**. Go with this PR as written. Address the nuance of primary versus unique keys through documentation
* **Rename this class to `UniqueKeyIndex`**. Keep constructors or helpers when you have a `PrimaryKey`
* **Introduce `UniqueKeyIndex` superclass**. This `PrimaryKeyIndex` would subclass and exist just for the `PrimaryKey` constructors or helpers.

hollow/src/main/java/com/netflix/hollow/api/consumer/index/PrimaryKeyIndex.java
```diff
@@ -0,0 +1,319 @@
...
+         * Configures the primary key index to listen to {@code HollowConsumer} version updates.
+         * On an update the index recalculates so updated data will be reflected in the results of a query
+         * (performed after the update).
+         *
+         * @param listen {@code true} if the index listens to {@code HollowConsumer} version updates,
+         * otherwise {@code false} and the index does not listen to {@code HollowConsumer} updates.  The default
+         * value is {@code true}, unless specified differently by the initiating builder method returning this builder.
+         * @return this builder
+         */
+        public Builder<T> listenToDataRefresh(boolean listen) {
```
ghost: I wonder if there's a more conventional name to use here that we can switch to since we don't have to worry about backwards compatibility?

Or flip it around and make `PrimaryKeyIndex` implement `HollowConsumer.RefreshListener` and have callers call `consumer.addRefreshListener(myIndex)` explicitly?

hollow/src/main/java/com/netflix/hollow/api/consumer/index/PrimaryKeyIndex.java
```diff
@@ -0,0 +1,319 @@
...
+    final String primaryTypeName;
+    final String[] matchFieldPaths;
+    HollowPrimaryKeyIndex hpki;
+    RefreshListener refreshListener;
+
+    PrimaryKeyIndex(
+            HollowConsumer consumer,
+            Class<T> primaryType,
+            PrimaryKey primaryTypeKey,
+            boolean listenToDataRefresh,
```
ghost: Maybe it's okay to remove the constructor argument and require that users must call the method.

ghost (approved):
I don't feel that strongly about the detailed feedback in this review, so this PR could go as-is.

I've never liked the `listenToDataRefresh` and `detatchFromDataRefresh` names. That's the closest to a strong opinion in all these suggestions.

PR #375, opened 2019-01-14
Flat Records 

dkoszewnik:
FlatRecords provide the ability to serialize records for addition to a remote state engine.

This is experimental functionality for now -- the API and serialized representations are both subject to backwards incompatible changes without notice.


hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/flatrecords/FlatRecord.java
```diff
@@ -0,0 +1,109 @@
+package com.netflix.hollow.core.write.objectmapper.flatrecords;
```
ghost, line 1: Per the incubating/forking theme, this could be `com.netflix.hollow.core.write.objectmapper.flatrecords.incubating` or `com.netflix.hollow.incubating.core.write.objectmapper.flatrecords`. Or replacing `incubating` with `experimental`.

hollow/src/main/java/com/netflix/hollow/core/write/HollowWriteFieldUtils.java
```diff
@@ -0,0 +1,26 @@
+package com.netflix.hollow.core.write;
+
+import com.netflix.hollow.core.memory.ByteData;
+
+public class HollowWriteFieldUtils {
+
+    public static int readIntBits(ByteData data, long fieldPosition) {
```
PaulSandoz, line 7: These two methods could be placed as default methods on ByteData, which then allows for more optimal implementations by overriding implementations (if present)

hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/flatrecords/FlatRecord.java
```diff
@@ -0,0 +1,109 @@
...
+import com.netflix.hollow.core.memory.encoding.ZigZag;
+import com.netflix.hollow.core.schema.HollowObjectSchema;
+import com.netflix.hollow.core.schema.HollowObjectSchema.FieldType;
+import com.netflix.hollow.core.schema.HollowSchema;
+import com.netflix.hollow.core.schema.HollowSchema.SchemaType;
+import com.netflix.hollow.core.write.HollowWriteFieldUtils;
+import com.netflix.hollow.core.write.objectmapper.RecordPrimaryKey;
+
+/**
+ * Warning: Experimental.  This is a BETA API and is subject to breaking changes.
```
ghost, line 15: Maybe:

> Warning: Experimental.  This API is subject to breaking changes.

Or if we go with the incubating nomenclature:

> Warning: Incubating.  This API is subject to breaking changes.

I picked up the "Incubating" nomenclature from a Netflix OSS meeting topic on being more explicit about project maturity. Anyhow...

I don't have a strong opinion on the word we use except that I think we should move away from using "beta". Ironically as it seems, I think "beta" has come to mean "almost mature" whereas experimental or incubating does a better job of indicating that the maturity of this feature.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalCyclePopulator.java
```diff
@@ -138,15 +144,32 @@ private void removeRecordsFromNewState(HollowProducer.WriteState newState, Map<S
...
+        AtomicInteger nextMutation = new AtomicInteger(0);
+        
         SimultaneousExecutor executor = new SimultaneousExecutor(threadsPerCpu);
-        for(final Map.Entry<RecordPrimaryKey, Object> entry : mutations.entrySet()) {
-            executor.execute(new Runnable() {
-                public void run() {
-                    if(entry.getValue() != DELETE_RECORD)
-                        newState.add(entry.getValue());
+        executor.execute(() -> {
+            FlatRecordDumper flatRecordDumper = null;
```
ghost: Per the incubating/forking theme, this gets simpler if all we're supporting (initially) is a FlatRecord based incremental producer.

Again, maybe Paul already has an idea for how to generalize this that makes my suggestion moot.

hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/HollowMapTypeMapper.java
```diff
@@ -76,7 +78,25 @@ protected int write(Object obj) {
...
+        int assignedOrdinal = writeState.add(rec);
+        
+        if(obj instanceof MemoizedMap) {
+            ((MemoizedMap<?, ?>)obj).__assigned_ordinal = (long)assignedOrdinal | cycleSpecificAssignedOrdinalBits();
+        }
+        
+        return assignedOrdinal;
+    }
+    
+    @Override
```
ghost, line 92: Add the admonition?

```
    /**
     * Warning: Experimental.  This is a BETA API and is subject to breaking changes.
     */
```

hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalCyclePopulator.java
PaulSandoz: There is no parallelism since only one task is executed asynchronously and the code that follows waits for that task in the executor to complete. Was that the intention?
My suggestion would be to keep this simple, snapshot the mutations (`values()`) into a list, and operate sequentially within the same thread. We can easily change this later to a stream and operate in parallel getting good decomposition. Or alternatively split the list into N parts in proportion to the number of hardware cores you want to use and execute N times (but parallel streams can do this more easily and better).
The guide of "threadsPerCpu" to control the parallelism is likely to be wrong most of the time. We need to express something that takes into account the machine parallelism (number of hardware cores to use), the amount of work (number of mutations), and the unit cost per work (the cost for each mutation). This is not easy to get right, hence why there is currently no mechanism in parallel streams :-) however we can control the former (machine parallelism) by executing the parallel stream within a custom fork join pool (to avoid using the common fork join pool).

PaulSandoz (approved):
Nice! I wrote some comments but i think we can iterate in master given the experimental nature.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowIncrementalCyclePopulator.java
dkoszewnik: You're right, that was not the intention.  I have corrected it.

@rpalcolea originally added parallelism here when he pulled out the `HollowIncrementalCyclePopulator`, perhaps in response to a bottleneck he encountered in practice, so I'll retain that for now.

ghost (approved):
Instead of adapting the existing incremental producer, should we instead build up a new replacement API? That way we can deprecate the current API as-is without introducing any changes to it. Also, I wonder if there are features in the current API that we'd like to abandon. Or if it's just easier to iterate on the design without being encumbered by choices in the current implementation.

If we want to go the route of introducing a new `HollowIncrementalProducer` and keep some of the same names, we could create the FlatRecord based implementation under an `incubating` package name.

I don't know if this lines up with the changes Paul had in mind. If the new API is far more similar to the current API then maybe it's a bad idea to fork the classes. I want to be able to release without having to disrupt Target or anyone else using the current API **until** we have an incremental producer that's a mature replacement for the "beta" one we already released.

Alternatively, we could do as we warned we might and make any breaking changes we need, suggesting that people fork the older API if they need it or await a mature release of the new API.

hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/flatrecords/FlatRecord.java
dkoszewnik: :+1: Improved the admonitions in the master branch after merge.

rydenius:
Curios if flat records ever left the experimental feature state, and if so where to find documentation for it? I can not find any mention on https://hollow.how/.

PR #378, opened 2019-01-17
Size ByteArrayOrdinalMap.

PaulSandoz:
The first commit reformats `ByteArrayOrdinalMap`. The second commit adds a new constructor that accepts a `size` argument that is used to pre-size the hash array.  Type write state restoration will size instances of ByteArrayOrdinalMap according to the cardinality of the populated ordinals on the read state.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
@@ -17,24 +17,6 @@
  */
 package com.netflix.hollow.core.memory;
 
-/*
-*
-*  Copyright 2013 Netflix, Inc.
-*
-*     Licensed under the Apache License, Version 2.0 (the "License");
-*     you may not use this file except in compliance with the License.
-*     You may obtain a copy of the License at
-*
-*         http://www.apache.org/licenses/LICENSE-2.0
-*
-*     Unless required by applicable law or agreed to in writing, software
-*     distributed under the License is distributed on an "AS IS" BASIS,
-*     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-*     See the License for the specific language governing permissions and
-*     limitations under the License.
-*
-*/
-
 import com.netflix.hollow.core.memory.encoding.HashCodes;
 import com.netflix.hollow.core.memory.encoding.VarInt;
 import com.netflix.hollow.core.memory.pool.WastefulRecycler;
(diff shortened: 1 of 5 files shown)


DanielThomas (commented):
If I'm reading `HollowTypeWriteState` and it's callers properly - this will improve the performance of the initial restore, but the very first population will use default sized maps and encounter resizes (assuming cardinality is about the same), but subsequent cycles will have a size that reflects the restored state? Any reason you couldn't size `ordinalMap` in `restoreFrom`?

PaulSandoz:
Yes, i missed that part e.g. in `HollowTypeWriteState.reuseOrdinalFromRestoredState`:
```
        if(restoredSchema instanceof HollowObjectSchema) {
            ((HollowObjectWriteRecord)rec).writeDataTo(scratch, (HollowObjectSchema)restoredSchema);
            int preferredOrdinal = restoredMap.get(scratch);
            scratch.reset();
            rec.writeDataTo(scratch);
            ordinal = ordinalMap.getOrAssignOrdinal(scratch, preferredOrdinal);
        } else {
```
We can add a method `ByteArrayOrdinalMap .resize` method.

PaulSandoz:
I made some further tweaks to ensure pre-sizing on restoration and when transitioning from write state to a read state.

In addition i attempted to strengthen against possible race conditions when a get is performed in one thread while a put is performed in another thread which results resizing (increasing in size) of arrays.

PaulSandoz:
Added benchmark. See results:

https://github.com/Netflix/hollow/pull/378/files#diff-b8ea099e68ada99dfedce5a1342df319R73

Resizing can have an 1.5x to 2x impact.

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
```diff
@@ -43,471 +25,554 @@
...
+    }
+
+    /**
+     * Resize the ordinal map by increasing its capacity.
+     * <p>
+     * No action is take if the current capacity is sufficient for the given size.
+     *
+     * @param size the size to increase to, rounded up to the nearest power of two.
+     */
+    public void resize(int size) {
```
dkoszewnik, line 456: May want to add an admonition similar to the one above `put()` that this operation is not thread-safe (since it can result in missed updates from the calls to `getOrAssignOrdinal`).  This scenario would also leave the `FreeOrdinalTracker` out of sync.

dkoszewnik (approved):

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
PaulSandoz: I updated the method documentation.

IIUC since a resize does not affect ordinal assignment (just the redistribution in the map) it means the free ordinal tracker size is out of sync, which previously would occur as ordinals get assigned (since the tracker size increases as ordinals are returned to the pool).

Looks like there is some opportunity to better size the tracker in `recalculateFreeOrdinals` (length - cardinality), although we could avoid the allocation of a temporary bit set instead. I am less certain about compaction.

PR #380, opened 2019-01-18
Make HollowObjectCacheProvider thread safe

ghost:
* [x] synchronize on `cachedItems`
* [x] avoid NPE on `factory` after detaching but still listening to type state
* [x] clearer (and more optimal) loop to fill ordinal gaps with `null`

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/objects/provider/HollowObjectCacheProvider.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/objects/provider/HollowObjectCacheProvider.java
@@ -93,16 +93,16 @@ public class HollowObjectCacheProvider<T> extends HollowObjectProvider<T> implem
     public void detach() {
         synchronized (cachedItems) {
             cachedItems.clear();
+            factory = null;
+            typeAPI = null;
+            typeReadState = null;
         }
-        factory = null;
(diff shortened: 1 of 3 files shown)


PaulSandoz (approved):

hollow/src/main/java/com/netflix/hollow/api/objects/provider/HollowObjectCacheProvider.java
```diff
@@ -85,21 +85,32 @@ public HollowObjectCacheProvider(HollowTypeDataAccess typeDataAccess, HollowType
...
         typeAPI = null;
         typeReadState = null;
     }
 
     @Override
     public void addedOrdinal(int ordinal) {
-        while(ordinal >= cachedItems.size())
-            cachedItems.add(null);
-        cachedItems.set(ordinal, instantiateCachedObject(factory, typeReadState, typeAPI, ordinal));
+        if (factory == null) {
```
ghost: This needs to be within the `synchronized` block.

ghost:
The synchronization was subsequently removed (f20d7baab37e65a955f1c2fe67a13ee1c069e9b7) due to performance concerns. The backfilled unit tests and the bugfix for detach were kept.

PR #382, opened 2019-01-21
Added FlatRecordExtractor

dkoszewnik:
to extract `FlatRecord`s from an existing `HollowReadStateEngine`

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/flatrecords/FlatRecord.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/flatrecords/FlatRecord.java
@@ -19,12 +19,10 @@ public class FlatRecord {
     final ByteData data;
     final int dataStartByte;
     int dataEndByte;
-    final int dataLength;
     RecordPrimaryKey recordPrimaryKey;
 
-    public FlatRecord(ByteData recordData, int dataLength, HollowSchemaIdentifierMapper schemaIdMapper) {
+    public FlatRecord(ByteData recordData, HollowSchemaIdentifierMapper schemaIdMapper) {
         this.data = recordData;
-        this.dataLength = dataLength;
         this.recordPrimaryKey = null;
         this.schemaIdMapper = schemaIdMapper;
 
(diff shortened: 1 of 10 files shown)


hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/flatrecords/FlatRecord.java
```diff
@@ -19,10 +19,12 @@
     final ByteData data;
     final int dataStartByte;
     int dataEndByte;
+    final int dataLength;
     RecordPrimaryKey recordPrimaryKey;
 
-    public FlatRecord(ByteData recordData, HollowSchemaIdentifierMapper schemaIdMapper) {
+    public FlatRecord(ByteData recordData, int dataLength, HollowSchemaIdentifierMapper schemaIdMapper) {
```
PaulSandoz: Instead of a `dataLength` parameter and field we could add a method to `ByteData` to get the length? (the default implementation throws `UnsupportedOperationException` and all hollow implementations override it. Alternatively create sub-type e.g. `ByteDataAccessor` with further methods.

hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/flatrecords/FlatRecordExtractor.java
```diff
@@ -0,0 +1,204 @@
...
+        public boolean ordinalIsMapped(String type, int originalOrdinal) {
+            throw new UnsupportedOperationException();
+        }
+        
+        public void clear() {
+            mappedFlatOrdinals.clear();
+        }
+        
+        private static class TypedOrdinal {
+            private final String type;
```
PaulSandoz, line 183: If `type` can never be `null` the the equals/hashCode can be simplified.

PaulSandoz (approved):

hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/flatrecords/FlatRecord.java
dkoszewnik: Yes, that's better.  Updated.

hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/flatrecords/FlatRecordExtractor.java
dkoszewnik: Simplified.

PR #393, opened 2019-02-12
Typesafe builders

ghost:
Exploring some changes in the consumer & producer builder APIs:

* typesafe builders typed by the `HollowAPI` they produce or consume
* consistent way to disable a feature or reset it to platform default
* avoid eager building in `withFoo`; only build during `build`
* better interaction with Builder subclasses (Cinder use-case)

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
@@ -66,7 +66,9 @@ public class HollowClientUpdater {
                                HollowMetricsCollector<HollowConsumerMetrics> metricsCollector) {
         this.planner = new HollowUpdatePlanner(transitionCreator, doubleSnapshotConfig);
         this.failedTransitionTracker = new FailedTransitionTracker();
-        this.staleReferenceDetector = new StaleHollowReferenceDetector(objectLongevityConfig, objectLongevityDetector);
+        this.staleReferenceDetector = objectLongevityConfig != null && objectLongevityDetector != null
+                ? new StaleHollowReferenceDetector(objectLongevityConfig, objectLongevityDetector)
+                : null;
         // Create a copy of the listeners, removing any duplicates
         this.refreshListeners = new CopyOnWriteArrayList<>(
                 refreshListeners.stream().distinct().toArray(HollowConsumer.RefreshListener[]::new));
@@ -74,10 +76,12 @@ public class HollowClientUpdater {
         this.hashCodeFinder = hashCodeFinder;
         this.doubleSnapshotConfig = doubleSnapshotConfig;
         this.objectLongevityConfig = objectLongevityConfig;
-        this.staleReferenceDetector.startMonitoring();
         this.metrics = metrics;
         this.metricsCollector = metricsCollector;
         this.initialLoad = new CompletableFuture<>();
+
+        if (staleReferenceDetector != null)
+            staleReferenceDetector.startMonitoring();
     }
 
     /**
--- a/hollow/src/main/java/com/netflix/hollow/api/client/HollowDataHolder.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/client/HollowDataHolder.java
@@ -124,7 +124,7 @@ class HollowDataHolder {
     }
 
     private void initializeAPI() {
-        if(objLongevityConfig.enableLongLivedObjectSupport()) {
+        if(objLongevityConfig != null && objLongevityConfig.enableLongLivedObjectSupport()) {
             HollowProxyDataAccess dataAccess = new HollowProxyDataAccess();
             dataAccess.setDataAccess(stateEngine);
             currentAPI = apiFactory.createAPI(dataAccess);
@@ -132,7 +132,8 @@ class HollowDataHolder {
             currentAPI = apiFactory.createAPI(stateEngine);
         }
         
-        staleReferenceDetector.newAPIHandle(currentAPI);
+        if (staleReferenceDetector != null)
+            staleReferenceDetector.newAPIHandle(currentAPI);
     }
 
     private void applyDeltaOnlyPlan(HollowUpdatePlan updatePlan, HollowConsumer.RefreshListener[] refreshListeners) throws Throwable {
@@ -145,7 +146,7 @@ class HollowDataHolder {
         try(InputStream is = blob.getInputStream()) {
             applyStateEngineTransition(is, blob, refreshListeners);
 
-            if(objLongevityConfig.enableLongLivedObjectSupport()) {
+            if(objLongevityConfig != null && objLongevityConfig.enableLongLivedObjectSupport()) {
                 HollowDataAccess previousDataAccess = currentAPI.getDataAccess();
                 HollowHistoricalStateDataAccess priorState = new HollowHistoricalStateCreator(null).createBasedOnNewDelta(currentVersion, stateEngine);
                 HollowProxyDataAccess newDataAccess = new HollowProxyDataAccess();
@@ -163,7 +164,7 @@ class HollowDataHolder {
                 priorHistoricalDataAccess = null;
             }
 
-            if(!staleReferenceDetector.isKnownAPIHandle(currentAPI))
+            if(staleReferenceDetector != null && !staleReferenceDetector.isKnownAPIHandle(currentAPI))
                 staleReferenceDetector.newAPIHandle(currentAPI);
             
             for(HollowConsumer.RefreshListener refreshListener : refreshListeners) {
--- a/hollow/src/main/java/com/netflix/hollow/api/client/HollowDataHolder.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/client/HollowDataHolder.java
@@ -124,7 +124,7 @@ class HollowDataHolder {
     }
 
     private void initializeAPI() {
-        if(objLongevityConfig != null && objLongevityConfig.enableLongLivedObjectSupport()) {
+        if(isObjectLongevityEnabled()) {
             HollowProxyDataAccess dataAccess = new HollowProxyDataAccess();
             dataAccess.setDataAccess(stateEngine);
             currentAPI = apiFactory.createAPI(dataAccess);
@@ -146,7 +146,7 @@ class HollowDataHolder {
         try(InputStream is = blob.getInputStream()) {
             applyStateEngineTransition(is, blob, refreshListeners);
 
-            if(objLongevityConfig != null && objLongevityConfig.enableLongLivedObjectSupport()) {
+            if(isObjectLongevityEnabled()) {
                 HollowDataAccess previousDataAccess = currentAPI.getDataAccess();
                 HollowHistoricalStateDataAccess priorState = new HollowHistoricalStateCreator(null).createBasedOnNewDelta(currentVersion, stateEngine);
                 HollowProxyDataAccess newDataAccess = new HollowProxyDataAccess();
@@ -211,4 +211,7 @@ class HollowDataHolder {
         currentVersion = version;
     }
 
+    private boolean isObjectLongevityEnabled() {
+        return objLongevityConfig != null && objLongevityConfig.enableLongLivedObjectSupport();
+    }
 }
--- a/hollow/src/main/java/com/netflix/hollow/api/client/StaleHollowReferenceDetector.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/client/StaleHollowReferenceDetector.java
@@ -17,6 +17,8 @@
  */
 package com.netflix.hollow.api.client;
 
+import static java.util.Objects.requireNonNull;
+
 import com.netflix.hollow.api.consumer.HollowConsumer;
 import com.netflix.hollow.api.custom.HollowAPI;
 import com.netflix.hollow.api.sampling.EnabledSamplingDirector;
@@ -28,6 +30,7 @@ import java.lang.ref.WeakReference;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Objects;
 
 /**
  * Detect stale Hollow references and USAGE of stale hollow references.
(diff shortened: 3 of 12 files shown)


hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
```diff
@@ -66,18 +66,22 @@ public HollowClientUpdater(HollowConsumer.BlobRetriever transitionCreator,
                                HollowMetricsCollector<HollowConsumerMetrics> metricsCollector) {
         this.planner = new HollowUpdatePlanner(transitionCreator, doubleSnapshotConfig);
         this.failedTransitionTracker = new FailedTransitionTracker();
-        this.staleReferenceDetector = new StaleHollowReferenceDetector(objectLongevityConfig, objectLongevityDetector);
+        this.staleReferenceDetector = objectLongevityConfig != null && objectLongevityDetector != null
```
ghost, line 69: I usually prefer to reduce the number of null checks in code. We already have `ObjectLongevityConfig.DEFAULT_CONFIG`. I could introduce a `NULL_CONFIG` or `NOOP_CONFIG`.

However, one thing I like about `null` here is it signals clear intent to permanently disable the feature, so we can leave out the detector (and its monitor thread) entirely.

hollow/src/main/java/com/netflix/hollow/api/client/StaleHollowReferenceDetector.java
```diff
@@ -114,7 +114,7 @@ private synchronized void housekeeping() {
 
     public void startMonitoring() {
         if (monitor == null) {
-            monitor = new Thread(new Monitor(this));
+            monitor = new Thread(new Monitor(this), "hollow-stale-reference-detector");
```
ghost: Just happened to notice this as I was here. I'll be extracting this change into its own PR.

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
```diff
@@ -163,6 +163,25 @@ protected HollowConsumer(BlobRetriever blobRetriever,
...
 
+    //// INCUBATING ////
+
+    protected HollowConsumer(AnnouncementWatcher announcementWatcher,
+                             HollowClientUpdater updater,
+                             Executor refreshExecutor,
+                             HollowConsumerMetrics metrics,
+                             ReadWriteLock refreshLock) {
+        /*
+         * notably different, this constructor doesn't auto-subscribe to announcement watcher
```
ghost, line 174: We've received some feedback that the subscribe to automatic background updates on consumer construction violated the principle of least astonishment.

We also have a Cinder use-case where we want more control over when and how the initial load happens even when using an `AnnouncementWatcher`.

hollow/src/main/java/com/netflix/hollow/api/consumer/incubating/HollowClientUpdaterV2.java
```diff
@@ -0,0 +1,29 @@
+package com.netflix.hollow.api.consumer.incubating;
+
+import static java.util.stream.Collectors.toList;
+
+import com.netflix.hollow.api.client.HollowClientUpdater;
+import com.netflix.hollow.api.custom.HollowAPI;
+import java.util.stream.Collectors;
+
+public class HollowClientUpdaterV2<A extends HollowAPI> extends HollowClientUpdater {
```
ghost, line 9: `FooV2` classes under an `incubating` package so that I could isolate these changes as much as possible while I noodled, not that I necessarily want to release them named as such.

hollow/src/main/java/com/netflix/hollow/api/consumer/incubating/HollowConsumerV2.java
```diff
@@ -0,0 +1,515 @@
...
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executor;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.stream.Collectors;
+
+public class HollowConsumerV2<A extends HollowAPI> extends HollowConsumer {
```
ghost, line 33: Having parameterized `HollowConsumer` and `HollowConsumer.Builder` was one of the original motivations.

`HollowConsumer` is unparameterized currently, so this would be a backwards compatible change if it were done in `HollowConsumer`.

hollow/src/main/java/com/netflix/hollow/api/consumer/incubating/HollowConsumerV2.java
```diff
@@ -0,0 +1,515 @@
...
+                builder.announcementWatcher.get(),
+                builder.updater,
+                builder.refreshExecutor.get(),
+                builder.metrics,
+                builder.refreshLock
+        );
+        this.updaterV2 = builder.updater;
+    }
+
+    public CompletableFuture<HollowConsumerV2<A>> initialLoad() {
```
ghost, line 51: My do-over on how I wish I'd written `getInitialLoad()`.

hollow/src/main/java/com/netflix/hollow/api/consumer/incubating/HollowConsumerV2.java
```diff
@@ -0,0 +1,515 @@
...
+     *
+     * @deprecated use type parameter on consumer, e.g. {@code HollowConsumerV2<T>} then call {@link #getAPI()}
+     */
+    @Override
+    @Deprecated
+    public <T extends HollowAPI> T getAPI(Class<T> apiClass) {
+        return apiClass.cast(updater.getAPI());
+    }
+
+    public static class Builder<A extends HollowAPI> {
```
ghost, line 75: Unlike `HollowConsumer`, the existing `HollowConsumer.Builder` is already parameterized. This would be a breaking change if applied to `HollowConsumer.Builder`

hollow/src/main/java/com/netflix/hollow/api/consumer/incubating/HollowConsumerV2.java
```diff
@@ -0,0 +1,515 @@
...
+        protected <T> Feature<T> feature() {
+            return new Feature<>();
+        }
+
+        public synchronized Builder<A> withBlobRetriever(BlobRetriever blobRetriever) {
+            requireNonNull(blobRetriever, "blobRetriever cannot be null; use blobRetrieverFeature().disable()");
+            return this.blobRetriever.set(blobRetriever);
+        }
+
+        public Feature<BlobRetriever> blobRetrieverFeature() {
```
ghost, line 120: I want every (or most) features to have a `withFoo(...)` and `noFoo()` pair of methods. I also want to avoid using `withFoo(null)` as the mechanism for disabling a feature; that way a caller with a coding error doesn't silently disable a feature in their builder.

I realized along the way that I really want 3 states: unset/default, user customized, and disabled. I got partway through making every feature have a triple of `withFoo(...)`, `resetFoo()`, and `noFoo()` before I decided to try `fooFeature().reset()` or `fooFeature().disable()`. Maybe for the final public API the `with|reset|disable` is simpler and better; for now it's one less boilerplate method I have to write as I'm experimenting.

hollow/src/main/java/com/netflix/hollow/api/consumer/incubating/HollowConsumerV2.java
```diff
@@ -0,0 +1,515 @@
...
+         * The instance returned from {@link HollowConsumer#getAPI()} will be of the provided type and can be cast
+         * to access generated methods.
+         *
+         * @param apiClass the code generated API class
+         * @return this builder
+         * @throws IllegalArgumentException if provided API class is {@code HollowAPI} instead of a subclass
+         */
+        public synchronized <A2 extends A> Builder<A2> withAPI(Class<A2> apiClass) {
+            /*
+             * FIXME(timt): not great...
```
ghost, line 181: Users are unlikely to come across this wart when using the builder in a fluent style, but I don't like it nonetheless.

Maybe we make it an error to call `withAPI(...)` more than once with a different `HollowAPI` subclass?

I like the solutions I've come up with about as much as leaving this as-is. Which is to say, I don't like any option that much.

hollow/src/main/java/com/netflix/hollow/api/consumer/incubating/HollowConsumerV2.java
```diff
@@ -0,0 +1,515 @@
...
+         *
+         * Use {@link #withAPI(Class)} to build a consumer with your custom client API and
+         * using the default high-density cache for all types.
+         * @param cachedType a type to enable object caching on
+         * @param additionalCachedTypes additional types to enable object caching on
+         * @return this builder
+         *
+         * @see <a href="https://hollow.how/advanced-topics/#caching">https://hollow.how/advanced-topics/#caching</a>
+         */
+        public synchronized <A2 extends A> Builder<A2> withAPI(Class<A2> apiClass,
```
ghost, line 240: Trivia: this entire PR is a yak shave that started when @DanielThomas asked for the ability to specify cached types in Cinder.

hollow/src/main/java/com/netflix/hollow/api/consumer/incubating/HollowConsumerV2.java
```diff
@@ -0,0 +1,515 @@
...
+        public synchronized Builder<A> withMetricsCollector(HollowMetricsCollector<HollowConsumerMetrics> metricsCollector) {
+            requireNonNull(metricsCollector, "metricsCollector cannot be null; use metricsCollectorFeature().disable()");
+            return this.metricsCollector.set(metricsCollector);
+        }
+
+        public Feature<HollowMetricsCollector<HollowConsumerMetrics>> metricsCollectorFeature() {
+            return metricsCollector;
+        }
+
+        public synchronized HollowConsumerV2<A> buildSync() {
```
ghost, line 320: I'm not married to the naming of `buildSync` and `buildAsync` or even having the builder be responsible for kicking off the initial load (directly or because the `HollowConsumer` does it). We have an independent requirement for a `consumer.dispose()` or `consumer.shutdown()` functionality, so maybe this becomes `consumer = builder.build().start()`.

Either way, we need to support an API where the initial load takes place in the caller's thread and not asynchronously with the caller blocking until ready (we want that use-case, too, hence `buildAsync()` or equivalent API sticks around).

hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
Sunjeet: 👍

PaulSandoz: `null` checks are cheap (HotSpot optimizes these away very efficiently), so if performance was a concern it need not be.

What if only one of the values is `null`. Then a `StaleHollowReferenceDetector` instance will be created and likely result in an NPE. You might wanna trigger this off config and if non-`null` requiring `objectLongevityDetector` to be non-`null` i.e. there should be `null` checks in `StaleHollowReferenceDetector` constructor for both values.

hollow/src/main/java/com/netflix/hollow/api/consumer/incubating/HollowConsumerV2.java
```diff
@@ -0,0 +1,515 @@
...
+     * @param apiClass the class of the API
+     * @param <T> the type of the API
+     * @return the API which wraps the underlying dataset
+     *
+     * @deprecated use type parameter on consumer, e.g. {@code HollowConsumerV2<T>} then call {@link #getAPI()}
+     */
+    @Override
+    @Deprecated
+    public <T extends HollowAPI> T getAPI(Class<T> apiClass) {
+        return apiClass.cast(updater.getAPI());
```
Sunjeet, line 72: should this be updaterV2?

hollow/src/main/java/com/netflix/hollow/api/client/HollowDataHolder.java
```diff
@@ -124,15 +124,16 @@ private void applySnapshotTransition(HollowConsumer.Blob snapshotBlob, HollowCon
     }
 
     private void initializeAPI() {
-        if(objLongevityConfig.enableLongLivedObjectSupport()) {
+        if(objLongevityConfig != null && objLongevityConfig.enableLongLivedObjectSupport()) {
```
PaulSandoz: Consider adding a private method `isObjectLongevityEnabled` rather than duplicating the logic.

I presume `staleReferenceDetector` is intertwined with `objLongevityConfig`. Is it also connected to the value returned by `objLongevityConfig.enableLongLivedObjectSupport`?

hollow/src/main/java/com/netflix/hollow/api/consumer/incubating/HollowConsumerV2.java
DanielThomas: I'm not a troublemaker I swear.

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
dkoszewnik: > We've received some feedback that the subscribe to automatic background updates on consumer construction violated the principle of least astonishment.

What would be the point of adding an `AnnouncementWatcher` if not to subscribe to automatic background updates?

hollow/src/main/java/com/netflix/hollow/api/consumer/incubating/HollowConsumerV2.java
ghost: Yes.

hollow/src/main/java/com/netflix/hollow/api/client/StaleHollowReferenceDetector.java
ghost: #402

ghost:
I may return to this in the future. Closing for now.

PR #394, opened 2019-02-14
Hash code updates.

PaulSandoz:
This pull request contains a number of updates related to the hash code of sets and maps primarily to enable the removal of `HollowObjectHashCodeFinder` and related hash code functionality at least on the producer side and limited use on the client side (specifically to ensure efficient lookup of elements or keys when using `HollowSet` or `HollowMap` until a solution is implemented that can reliably extract a hash key from an object).

- Support `@HollowHashKey` and hash key with empty fields for hollow set and map schema.  This feature can override the application of a default hash key (see `HollowObjectMapper.useDefaultHashKeys` which is true by default) when it is required that ordinals be used as the hash code.

- Allow a consumer to filter out hash keys of set or map schema.  The consumer needs to register an instance of `HollowObjectHashCodeFinder` whose method `getTypesWithDefinedHashCodes` returns a `Set` that contains the element or key type of the set or map schema.  This is a breaking change for an edge case: prior to this change a set/map with a hash key would override that of `HollowObjectHashCodeFinder`.  However, prior to this change it was not expected that a consumer implement `getTypesWithDefinedHashCodes` with anything but an empty set. 

- Ensure a map or set whose element or key type is filtered (or not present) can still be processed.  Any hash key is ignored since the field paths cannot be resolved.

- When performing a checksum calculation for a set or map relax the schema equality check for the hash keys if the schema evolved from no hash key to with a hash key.  This enables a transition from no hash key (the hash is defined externally using `HollowObjectHashCodeFinder`) to a hash key (the hash is defined by the hash key on the schema).

- Ensure a producer can transition (with restoration) from a state where `HollowObjectHashCodeFinder` was used (and where default hash keys were not enabled) to generate hash codes to where `HollowObjectHashCodeFinder` is not used (and where where default hash keys are enabled).  If hash codes are the same in both cases then the checksum calculation performed in the integrity check stage should not fail.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/read/engine/HollowBlobReader.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/read/engine/HollowBlobReader.java
@@ -195,7 +195,7 @@ public class HollowBlobReader {
                 HollowSetSchema setSchema = (HollowSetSchema) schema;
                 // Filter out hash key if overridden by a hash code finder
                 if (hasDefinedHashCode(setSchema.getElementType())) {
-                    setSchema = setSchema.withoutKeys();
+                    setSchema = setSchema.withoutHashKey();
                 }
                 populateTypeStateSnapshot(is, new HollowSetTypeReadState(stateEngine, setSchema, numShards));
             }
@@ -206,7 +206,7 @@ public class HollowBlobReader {
                 HollowMapSchema mapSchema = (HollowMapSchema) schema;
                 // Filter out hash key if overridden by a hash code finder
                 if (hasDefinedHashCode(mapSchema.getKeyType())) {
-                    mapSchema = mapSchema.withoutKeys();
+                    mapSchema = mapSchema.withoutHashKey();
                 }
                 populateTypeStateSnapshot(is, new HollowMapTypeReadState(stateEngine, mapSchema, numShards));
             }
--- a/hollow/src/main/java/com/netflix/hollow/core/read/engine/HollowBlobReader.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/read/engine/HollowBlobReader.java
@@ -165,6 +165,10 @@ public class HollowBlobReader {
         }
     }
 
+    private boolean hasDefinedHashCode(String type) {
+        return stateEngine.getHashCodeFinder().getTypesWithDefinedHashCodes().contains(type);
+    }
+
     private String readTypeStateSnapshot(DataInputStream is, HollowBlobHeader header, HollowFilterConfig filter) throws IOException {
         HollowSchema schema = HollowSchema.readFrom(is);
 
@@ -188,13 +192,23 @@ public class HollowBlobReader {
             if(!filter.doesIncludeType(schema.getName())) {
                 HollowSetTypeReadState.discardSnapshot(is, numShards);
             } else {
-                populateTypeStateSnapshot(is, new HollowSetTypeReadState(stateEngine, (HollowSetSchema)schema, numShards));
+                HollowSetSchema setSchema = (HollowSetSchema) schema;
+                // Filter out hash key if overridden by a hash code finder
+                if (hasDefinedHashCode(setSchema.getElementType())) {
+                    setSchema = setSchema.withoutKeys();
+                }
+                populateTypeStateSnapshot(is, new HollowSetTypeReadState(stateEngine, setSchema, numShards));
             }
         } else if(schema instanceof HollowMapSchema) {
             if(!filter.doesIncludeType(schema.getName())) {
                 HollowMapTypeReadState.discardSnapshot(is, numShards);
             } else {
-                populateTypeStateSnapshot(is, new HollowMapTypeReadState(stateEngine, (HollowMapSchema)schema, numShards));
+                HollowMapSchema mapSchema = (HollowMapSchema) schema;
+                // Filter out hash key if overridden by a hash code finder
+                if (hasDefinedHashCode(mapSchema.getKeyType())) {
+                    mapSchema = mapSchema.withoutKeys();
+                }
+                populateTypeStateSnapshot(is, new HollowMapTypeReadState(stateEngine, mapSchema, numShards));
             }
         }
         
--- a/hollow/src/main/java/com/netflix/hollow/core/read/engine/map/HollowMapTypeReadState.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/read/engine/map/HollowMapTypeReadState.java
@@ -260,9 +260,8 @@ public class HollowMapTypeReadState extends HollowTypeReadState implements Hollo
     @Override
     protected void applyToChecksum(HollowChecksum checksum, HollowSchema withSchema) {
         if(!getSchema().equals(withSchema)) {
-            // Apply checksum if one or other schema does not declare a hash key
-            if (getSchema().getHashKey() != null &&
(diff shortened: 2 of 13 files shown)


build.gradle
```diff
@@ -3,11 +3,12 @@ buildscript {
 }
 
 plugins {
-  id 'nebula.netflixoss' version '6.0.3'
+//  id 'nebula.netflixoss' version '6.0.3'
 }
 
 subprojects {
-  apply plugin: 'nebula.netflixoss'
+//  apply plugin: 'nebula.netflixoss'
```
dkoszewnik: Intentional?

hollow/src/main/java/com/netflix/hollow/core/read/engine/HollowBlobReader.java
```diff
@@ -189,13 +193,23 @@ private String readTypeStateSnapshot(DataInputStream is, HollowBlobHeader header
             if(!filter.doesIncludeType(schema.getName())) {
                 HollowSetTypeReadState.discardSnapshot(is, numShards);
             } else {
-                populateTypeStateSnapshot(is, new HollowSetTypeReadState(stateEngine, (HollowSetSchema)schema, numShards));
+                HollowSetSchema setSchema = (HollowSetSchema) schema;
+                // Filter out hash key if overridden by a hash code finder
+                if (hasDefinedHashCode(setSchema.getElementType())) {
```
dkoszewnik, line 197: We should probably note that it's possible to have a Set that uses explicit hashing via the `HollowObjectHashCodeFinder`, plus a Set with an element of the same type that uses hash keys.

I don't think it's likely that anyone does this, but this change would break that case.

hollow/src/main/java/com/netflix/hollow/core/read/engine/map/HollowMapTypeReadState.java
```diff
@@ -260,9 +260,16 @@ void setCurrentData(HollowMapTypeDataElements data) {
 
     @Override
     protected void applyToChecksum(HollowChecksum checksum, HollowSchema withSchema) {
-        if(!getSchema().equals(withSchema))
-            throw new IllegalArgumentException("HollowMapTypeReadState cannot calculate checksum with unequal schemas: " + getSchema().getName());
-        
+        if(!getSchema().equals(withSchema)) {
+            // Apply checksum if one or other schema does not declare a hash key
+            if (getSchema().getHashKey() != null &&
+                    (!(withSchema instanceof HollowMapSchema) || ((HollowMapSchema) withSchema).getHashKey() != null)) {
```
dkoszewnik: I think you want to compare both schemas `.withoutKeys()` here instead of making sure that the other _has_ a hash key?  This condition is not sufficient to determine equality.

hollow/src/main/java/com/netflix/hollow/core/read/engine/set/HollowSetTypeReadState.java
```diff
@@ -238,9 +238,16 @@ void setCurrentData(HollowSetTypeDataElements data) {
 
     @Override
     protected void applyToChecksum(HollowChecksum checksum, HollowSchema withSchema) {
-        if(!getSchema().equals(withSchema))
-            throw new IllegalArgumentException("HollowSetTypeReadState cannot calculate checksum with unequal schemas: " + getSchema().getName());
-        
+        if (!getSchema().equals(withSchema)) {
+            // Apply checksum if one or other schema does not declare a hash key
+            if (getSchema().getHashKey() != null &&
+                    (!(withSchema instanceof HollowSetSchema) || ((HollowSetSchema) withSchema).getHashKey() != null)) {
```
dkoszewnik: I think you want to compare both schemas `.withoutKeys()` here as well.

hollow/src/main/java/com/netflix/hollow/core/schema/HollowSchema.java
```diff
@@ -47,6 +48,24 @@
...
+     * A special constant for determining if a hollow set or map schema supports a hash key
+     * for ordinal values.
+     */
+    static final String ORDINAL_HASH_KEY_FIELD_NAME = "0rdinal";
+
+    /**
+     * A special constant for determining if a hollow set or map schema supports a hash key
+     * for ordinal values.
+     */
+    static final String[] ORDINAL_HASH_KEY_FIELD_NAMES = { ORDINAL_HASH_KEY_FIELD_NAME };
```
dkoszewnik: This concept is currently represented as an empty array.  I am not seeing what this change in serialization format of this concept is providing over the current representation?  Was the current representation incomplete in some way?  If so, can we fix it rather than change it?

dkoszewnik (changes requested):

build.gradle
PaulSandoz: Doh, no, side-effects of a composite builds i forgot to remove

hollow/src/main/java/com/netflix/hollow/core/read/engine/HollowBlobReader.java
PaulSandoz: Good point.  This is a transitory solution to retain some compatibility until consumers can remove the dependency on `HollowObjectHashCodeFinder`.  We could change the approach.  For example, the producer filters to remove the hash keys before publishing (since the producer and consumers both require knowledge of the schema types).

hollow/src/main/java/com/netflix/hollow/core/schema/HollowSchema.java
PaulSandoz: The problem is the varargs constructor:
```
    public HollowSetSchema(String schemaName, String elementType, String... hashKeyFieldPaths)
```
Although a `null` or empty array outside of this constructor can mean different things it does not mean the same thing when passed as an argument since the constructor cannot differentiate between the passing of an argument or not e.g. `new HollowSetSchema("A", "B")` results in the implicit passing of an empty `hashKeyFieldPaths` array.

The string "0rdinal" never appears in serialized form or accessible via a public API.  It's just an internal trick so `HollowSetSchema` can differentiate between null/empty and a hash key for ordinals.

hollow/src/main/java/com/netflix/hollow/core/read/engine/map/HollowMapTypeReadState.java
PaulSandoz: I was somewhat obtuse here. It was intentional :-) i wanted to only support the transition from without hash key to with a hash key. Going the other way around would seem to imply the use of `HollowObjectHashCodeFinder`?

PaulSandoz: I get it now. The check is too weak.

PaulSandoz:
I pushed another commit that:
- fixes the bug for schema mismatch when applying a checking; and
- simplifies the field path array argument processing for set and map schema when constructed to explicitly use the ordinal as the hash code.

hollow/src/main/java/com/netflix/hollow/core/schema/HollowSchema.java
PaulSandoz: FTR i removed this feature.

hollow/src/main/java/com/netflix/hollow/core/read/engine/HollowBlobReader.java
PaulSandoz: I updated the main section with an addition comment. This is likely rare since prior to this change a consumer is unlikely to implement `getTypesWithDefinedHashCodes` as it was only actionable for the producer.

PR #395, opened 2019-02-15
Producer cycle and announcement metrics using listeners

Sunjeet:
Following the Consumer refresh metrics change, this PR brings parity to the producer side in how metrics are being computed in Hollow. Clients can plug in their custom metrics reporting behavior.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
@@ -10,6 +10,8 @@ import com.netflix.hollow.core.read.engine.HollowReadStateEngine;
 import java.util.List;
 import java.util.OptionalLong;
 import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * A class for computing Hollow Consumer refresh metrics, requires plugging in metrics reporting implementation.
@@ -20,6 +22,8 @@ import java.util.concurrent.TimeUnit;
  * to implement custom metrics reporting behavior.
  */
 public abstract class AbstractRefreshMetricsListener extends AbstractRefreshListener implements RefreshMetricsReporting {
+
+    private static final Logger log = Logger.getLogger(AbstractRefreshMetricsListener.class.getName());
 
     private OptionalLong lastRefreshTimeNanoOptional;
     private long refreshStartTimeNano;
@@ -74,6 +78,23 @@ public abstract class AbstractRefreshMetricsListener extends AbstractRefreshList
         updatePlanDetails.numSuccessfulTransitions ++;
     }
 
+    /**
+     * Metrics reporting implementation is provided by the extending subclass. If exceptions are not gracefully handled
+     * in the extending subclass then an exception there can fail the consumer refresh, even though metrics reporting
+     * might not be mission critical. This method protects against that scenario by catching all exceptions, logging
+     * that there was an exception, and continuing with the consumer refresh.
+     * @param refreshMetrics Consumer refresh metrics being reported
+     */
+
+    private final void noFailRefreshEndMetricsReporting(ConsumerRefreshMetrics refreshMetrics) {
+        try {
+            refreshEndMetricsReporting(refreshMetrics);
+        } catch (Exception e) {
+            // Metric reporting is not considered critical to consumer refresh. Log exceptions and continue.
+            log.log(Level.WARNING, "Encountered an exception in reporting consumer refresh metrics, ignoring exception and continuing with consumer refresh", e);
+        }
+    }
+
     @Override
     public final void refreshSuccessful(long beforeVersion, long afterVersion, long requestedVersion) {
         long refreshEndTimeNano = System.nanoTime();
@@ -87,9 +108,8 @@ public abstract class AbstractRefreshMetricsListener extends AbstractRefreshList
                 .setConsecutiveFailures(consecutiveFailures)
                 .setRefreshSuccessAgeMillisOptional(0l)
                 .setRefreshEndTimeNano(refreshEndTimeNano);
-        ConsumerRefreshMetrics refreshMetrics = refreshMetricsBuilder.build();
 
-        refreshEndMetricsReporting(refreshMetrics);
+        noFailRefreshEndMetricsReporting(refreshMetricsBuilder.build());
     }
 
     @Override
@@ -105,8 +125,8 @@ public abstract class AbstractRefreshMetricsListener extends AbstractRefreshList
         if (lastRefreshTimeNanoOptional.isPresent()) {
             refreshMetricsBuilder.setRefreshSuccessAgeMillisOptional(TimeUnit.NANOSECONDS.toMillis(refreshEndTimeNano - lastRefreshTimeNanoOptional.getAsLong()));
         }
-        ConsumerRefreshMetrics refreshMetrics = refreshMetricsBuilder.build();
-        refreshEndMetricsReporting(refreshMetrics);
+
+        noFailRefreshEndMetricsReporting(refreshMetricsBuilder.build());
     }
 
     @Override
--- a/hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
@@ -85,13 +85,12 @@ public abstract class AbstractRefreshMetricsListener extends AbstractRefreshList
      * that there was an exception, and continuing with the consumer refresh.
      * @param refreshMetrics Consumer refresh metrics being reported
      */
-
     private final void noFailRefreshEndMetricsReporting(ConsumerRefreshMetrics refreshMetrics) {
         try {
             refreshEndMetricsReporting(refreshMetrics);
         } catch (Exception e) {
             // Metric reporting is not considered critical to consumer refresh. Log exceptions and continue.
-            log.log(Level.WARNING, "Encountered an exception in reporting consumer refresh metrics, ignoring exception and continuing with consumer refresh", e);
+            log.log(Level.SEVERE, "Encountered an exception in reporting consumer refresh metrics, ignoring exception and continuing with consumer refresh", e);
         }
     }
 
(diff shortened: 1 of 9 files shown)


hollow/src/main/java/com/netflix/hollow/api/producer/metrics/AbstractProducerMetricsListener.java
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * On cycle completion this method reports cycle metrics.
+     * @param status Whether the cycle succeeded or failed
+     * @param readState Hollow data state published by cycle, not used here because data size is known from when announcement completed
+     * @param version Version of data that was published in this cycle
+     * @param elapsed Cycle start to end duration
+     */
+    @Override
+    public void onCycleComplete(com.netflix.hollow.api.producer.Status status, HollowProducer.ReadState readState, long version, Duration elapsed) {
+        Optional<Boolean> isCycleSuccess;
```
PaulSandoz: An empty value is never created. It's present with either true or false.

hollow/src/main/java/com/netflix/hollow/api/producer/metrics/AnnouncementMetrics.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+        public Builder setAnnouncementDurationMillis(long announcementDurationMillis) {
+            this.announcementDurationMillis = announcementDurationMillis;
+            return this;
+        }
+        public Builder setIsAnnouncementSuccess(boolean isAnnouncementSuccess) {
+            this.isAnnouncementSuccess = isAnnouncementSuccess;
+            return this;
+        }
+        public Builder setLastAnnouncementSuccessTimeNanoOptional(OptionalLong lastAnnouncementSuccessTimeNanoOptional) {
```
PaulSandoz: The lack of calling or calling the builder method implies optionally indicating that it does not need to be expressed as a method argument.  A builder does not require that it be built in full fluently with variables accumulated before that being passed as arguments.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
```diff
@@ -74,6 +78,23 @@ public final void blobLoaded(HollowConsumer.Blob transition) {
...
     }
 
+    /**
+     * Metrics reporting implementation is provided by the extending subclass. If exceptions are not gracefully handled
+     * in the extending subclass then an exception there can fail the consumer refresh, even though metrics reporting
+     * might not be mission critical. This method protects against that scenario by catching all exceptions, logging
+     * that there was an exception, and continuing with the consumer refresh.
+     * @param refreshMetrics Consumer refresh metrics being reported
+     */
+
```
ghost: remove blank line

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
```diff
@@ -74,6 +78,23 @@ public final void blobLoaded(HollowConsumer.Blob transition) {
...
+    private final void noFailRefreshEndMetricsReporting(ConsumerRefreshMetrics refreshMetrics) {
+        try {
+            refreshEndMetricsReporting(refreshMetrics);
+        } catch (Exception e) {
+            // Metric reporting is not considered critical to consumer refresh. Log exceptions and continue.
+            log.log(Level.WARNING, "Encountered an exception in reporting consumer refresh metrics, ignoring exception and continuing with consumer refresh", e);
```
ghost: should probably log at `ERROR` level. Even though we recover, it may represent a genuine problem.

hollow/src/main/java/com/netflix/hollow/api/producer/metrics/AnnouncementMetrics.java
ghost: I think I understand @PaulSandoz's feedback and therefore I think I agree.

Meaning, the argument should just be of type `long` since if we omit the call to `setLastAnnouncementSuccessTimeNanoOptional` then `this.lastAnnouncementSuccessTimeNanoOptional` is already an empty optional.

Sunjeet: With making the arguments to the builder optional, I'm able to do this in onCycleSkip()-

> cycleMetricsBuilder
>                 .setConsecutiveFailures(consecutiveFailures)
>                 .setIsCycleSuccessOptional(Optional.empty())
>                 .setCycleDurationMillisOptional(OptionalLong.empty())
>                 .setLastCycleSuccessTimeNanoOptional(lastCycleSuccessTimeNanoOptional);
> cycleMetricsReporting(cycleMetricsBuilder.build());

as opposed to this-
> cycleMetricsBuilder
>                 .setConsecutiveFailures(consecutiveFailures)
>                 .setLastCycleSuccessTimeNano(lastCycleSuccessTimeNano);
> cycleMetricsReporting(cycleMetricsBuilder.build());

Both are accurate, but the former makes it more obvious to the reader that we are not reporting the success status or cycle duration for skipped cycles. IMO there is a readability benefit in cases such as this one, where the builder object is not entirely built in one place, so it's easier to follow what we're setting for each parameter.

Could that justify the usage of Optional in builder arguments?

hollow/src/main/java/com/netflix/hollow/api/producer/metrics/AbstractProducerMetricsListener.java
Sunjeet: Optional was convenient when the builder argument was Optional<Boolean>, but if that goes then this goes.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
Sunjeet: 👍

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
Sunjeet: 👍

hollow/src/main/java/com/netflix/hollow/api/producer/metrics/AnnouncementMetrics.java
ghost: > reporting the success status or cycle duration for skipped cycles

Got it. Your rationale makes sense.

Probably worth discussing the `SKIP` status with @PaulSandoz. He wants status to become `SUCCESS` or `FAIL` only, for skipped cycles to be successful cycles, and to have a different means for indicating skipped cycles. I don't know if this PR is the right time for that change. Assuming "no, not the this PR" then I'm fine with the explicit `.empty()` in `onCycleSkip`

PaulSandoz: There are effectively two layers of optionality here. Arguably the declaration with `Optional.isEmpty()` is no more expressive than a code comment (unless you support the reseting of content since you can delete it with no adverse functional consequences). I can make the counterpoint that when `setIsCycleSuccessOptional` is called with a value i have no idea whether it is empty or present without looking at the surrounding code to see how the value is constructed. I think it would be simpler if the builder is specified such that on construction (or after it has been built) it is empty and it can only be accumulated with present items thereby removing the ambiguity. Use of `Optional` is not required in such a model.

FWIW one primary motivation to use a builder is Java's lack of optional/default constructor/method parameters.  If there are only a few ways to construct declare constructors instead of introducing another API artifact.  If it is envisaged that more forms of construction may added later on a builder can be useful to avoid an explosion of telescoping constructors.

hollow/src/main/java/com/netflix/hollow/api/producer/metrics/AbstractProducerMetricsListener.java
```diff
@@ -0,0 +1,125 @@
...
+
+        HollowReadStateEngine stateEngine = readState.getStateEngine();
+        dataSizeBytes = stateEngine.calcApproxDataSize();
+
+        announcementMetricsBuilder
+                .setDataSizeBytes(dataSizeBytes)
+                .setIsAnnouncementSuccess(isAnnouncementSuccess)
+                .setAnnouncementDurationMillis(elapsed.toMillis());
+
+        if (lastAnnouncementSuccessTimeNanoOptional.isPresent())
```
PaulSandoz: One way to avoid the `isPresent` and `get` dance is to use `ifPresent`. I suspect you can pass in a method reference in this case. Up to you.

hollow/src/main/java/com/netflix/hollow/api/producer/metrics/AnnouncementMetrics.java
```diff
@@ -0,0 +1,64 @@
...
+    public long getDataSizeBytes() {
+        return dataSizeBytes;
+    }
+    public long getAnnouncementDurationMillis() {
+        return announcementDurationMillis;
+    }
+    public boolean getIsAnnouncementSuccess() {
+        return isAnnouncementSuccess;
+    }
+    public OptionalLong getLastAnnouncementSuccessTimeNanoOptional() {
```
PaulSandoz: Minor point. Naming-wise i don't think it necessary to express the kind of return type in the method name. IMHO the return type already expresses that. In the case of the setter of the builder it looks a little odd :-)

PaulSandoz (approved):

hollow/src/main/java/com/netflix/hollow/api/producer/metrics/AbstractProducerMetricsListener.java
Sunjeet: Neat, done

hollow/src/main/java/com/netflix/hollow/api/producer/metrics/AnnouncementMetrics.java
Sunjeet: Done

hollow/src/main/java/com/netflix/hollow/api/producer/metrics/AnnouncementMetrics.java
Sunjeet: This suggestion was taken

PR #397, opened 2019-02-18
Checksum performance improvement for collections.

PaulSandoz:
When the number of shards increases the time to calculate the checksum of a list, set or map increases. This will likely have an impact for models with a large number of collections, especially when each collection is comprised of a small number of elements.

There is an outer loop for the number of shards , a middle inner loop for the collection matching the shard number, and a further final inner loop for the collection's elements. That middle inner loop iterates over the whole populated bit set for the collection in single increments.  Instead it is possible to increment by the number of shards (with care to ensure correct rounding up to the nearest ordinal for the matching shard number). This will not change the checksum algorithm.

Results show that: (See the JMH benchmark https://github.com/Netflix/hollow/compare/master...PaulSandoz:checksum-perf?expand=1#diff-210ee32dee6840abed9fa98a6a8c152dR164) 

1. For the old code the time to calculate a checksum increases as the shards increases;
2. For the new code there is no regression when there is only one shard; and
3. For the new code the time to calculate a checksum is mostly independent of the number of shards (when the number of collections and size is held constant the calculation time is approximately constant).

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/read/engine/list/HollowListTypeReadStateShard.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/read/engine/list/HollowListTypeReadStateShard.java
@@ -99,7 +99,7 @@ class HollowListTypeReadStateShard {
     }
 
     protected void applyToChecksum(HollowChecksum checksum, BitSet populatedOrdinals, int shardNumber, int numShards) {
-        int ordinal = populatedOrdinals.nextSetBit(0);
+        int ordinal = populatedOrdinals.nextSetBit(shardNumber);
         while(ordinal != ORDINAL_NONE) {
             if((ordinal & (numShards - 1)) == shardNumber) {
                 int shardOrdinal = ordinal / numShards;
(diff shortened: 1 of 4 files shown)


hollow/src/main/java/com/netflix/hollow/core/read/engine/list/HollowListTypeReadStateShard.java
```diff
@@ -108,9 +108,14 @@ protected void applyToChecksum(HollowChecksum checksum, BitSet populatedOrdinals
...
-            }
 
-            ordinal = populatedOrdinals.nextSetBit(ordinal + 1);
+                ordinal = ordinal + numShards;
+            } else {
+                // Round up ordinal
+                int r = (ordinal & -numShards) + shardNumber;
+                ordinal = (r <= ordinal) ? r + numShards : r;
+            }
+            ordinal = populatedOrdinals.nextSetBit(ordinal);
```
dkoszewnik, line 118: Shouldn't this be `ordinal = populatedOrdinals.nextSetBit(ordinal+1);`?

dkoszewnik: Oh, I see that you are incrementing by `numShards` in the matched shard case above.  So most of the time, you're doing the rounding in the next iteration of the loop after this call.  
Might it be clearer to round/align the ordinal with the shard here, rather than in `(104,112-117)`?

dkoszewnik (approved):

hollow/src/main/java/com/netflix/hollow/core/read/engine/list/HollowListTypeReadStateShard.java
PaulSandoz: I am not following you. The calculation at line 112, `ordinal = ordinal + numShards`, is just an optimized version of that on lines 114-116. Both round up to align to the next ordinal to check if set.

A simplification could be:
```
if((ordinal & (numShards - 1)) == shardNumber) {
    ...
}
// Round up ordinal to align with shard number
int alignedOrdinal = (ordinal & -numShards) + shardNumber;
alignedOrdinal = (r <= ordinal) ? r + numShards : r;
ordinal = populatedOrdinals.nextSetBit(alignedOrdinal)
```
I doubt that would impact the performance.

PR #401, opened 2019-02-23
HollowJsonAdapter can convert json into FlatRecords

dkoszewnik:
* Added `SimpleHollowDataset`, since I seem to keep recreating this class in various projects.
* Bugfix: FlatRecordWriter was not logging a new record's ordinal when it's hash collided with an existing record.
* `HollowJsonAdapter`: added ability to parse json directly into `FlatRecord`s.

### DIFF
--- a/hollow-jsonadapter/src/main/java/com/netflix/hollow/jsonadapter/HollowJsonAdapter.java
+++ b/hollow-jsonadapter/src/main/java/com/netflix/hollow/jsonadapter/HollowJsonAdapter.java
@@ -30,6 +30,7 @@ import com.netflix.hollow.core.write.HollowObjectWriteRecord;
 import com.netflix.hollow.core.write.HollowSetWriteRecord;
 import com.netflix.hollow.core.write.HollowWriteRecord;
 import com.netflix.hollow.core.write.HollowWriteStateEngine;
+import com.netflix.hollow.core.write.objectmapper.flatrecords.FlatRecordWriter;
 import com.netflix.hollow.jsonadapter.field.FieldProcessor;
 import java.io.File;
 import java.io.IOException;
@@ -52,10 +53,11 @@ public class HollowJsonAdapter extends AbstractHollowJsonAdaptorTask {
 
     private final Map<String, ObjectFieldMapping> canonicalObjectFieldMappings;
 
-    ////TODO: Special 'passthrough' processing.
     private final Set<String> passthroughDecoratedTypes;
     private final ThreadLocal<PassthroughWriteRecords> passthroughRecords;
 
+    /// TODO: Would be nice to be able to take a HollowDataset here, if only producing FlatRecords,
+    ///       instead of requiring a HollowWriteStateEngine
     public HollowJsonAdapter(HollowWriteStateEngine stateEngine, String typeName) {
         super(typeName, "populate");
         this.stateEngine = stateEngine;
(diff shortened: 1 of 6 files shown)


hollow-jsonadapter/src/main/java/com/netflix/hollow/jsonadapter/HollowJsonAdapter.java
```diff
@@ -510,6 +519,14 @@ private void skipObjectField(JsonParser parser, JsonToken token) throws IOExcept
         }
     }
 
+    private int addRecord(String type, HollowWriteRecord rec, FlatRecordWriter flatRecordWriter) {
+        if(flatRecordWriter != null) {
```
ghost, line 523: There's a long call-chain of methods that are just passing `flatRecordWriter` along to the next method. I think that's a bit of a design smell. But I think that can be addressed in a separate refactoring.

hollow-jsonadapter/src/main/java/com/netflix/hollow/jsonadapter/HollowJsonToFlatRecordTask.java
```diff
@@ -0,0 +1,69 @@
...
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.netflix.hollow.core.write.objectmapper.flatrecords.FlatRecord;
+import com.netflix.hollow.core.write.objectmapper.flatrecords.FlatRecordWriter;
+import com.netflix.hollow.core.write.objectmapper.flatrecords.HollowSchemaIdentifierMapper;
+import java.io.IOException;
+import java.io.Reader;
+import java.util.function.Consumer;
+
+public class HollowJsonToFlatRecordTask extends AbstractHollowJsonAdaptorTask {
```
ghost, line 28: Add

```
/**
 * Incubating API; not for general use.
 */
```

I'd like to see that applied to all the FlatRecord classes or FlatRecord-specific methods until this API stabilizes. This can be added across the board in a separate PR, though.

hollow/src/main/java/com/netflix/hollow/core/schema/SimpleHollowDataset.java
```diff
@@ -0,0 +1,66 @@
...
+import com.netflix.hollow.core.HollowDataset;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A HollowDataset implementation which only describes a set of schemas comprising a dataset. 
+ */
+public class SimpleHollowDataset implements HollowDataset {
```
ghost, line 30: I wonder if there's a more descriptive name. Something is simple until it's simpler variant comes along.

ghost (approved):
Small nits only. Can be addressed in later changes.

hollow-jsonadapter/src/main/java/com/netflix/hollow/jsonadapter/HollowJsonAdapter.java
PaulSandoz: Yes, i agree with this, but i don't currently know what a better solution might be.

hollow-jsonadapter/src/main/java/com/netflix/hollow/jsonadapter/HollowJsonToFlatRecordTask.java
```diff
@@ -0,0 +1,69 @@
...
+    @Override
+    protected int processRecord(JsonParser parser) throws IOException {
+        FlatRecordWriter recWriter = getFlatRecordWriter();
+        int ordinal = adapter.processRecord(parser, recWriter);
+        FlatRecord rec = recWriter.generateFlatRecord();
+        action.accept(rec);
+        return ordinal;
+    }
+
+    private FlatRecordWriter getFlatRecordWriter() {
```
PaulSandoz, line 59: It's possible to place this logic closer to the `ThreadLocal` by extension and implementation of `initialValue`. See `ThreadLocal.withInitial`:
```java
    this.flatRecordWriter = ThreadLocal.withInitial(() -> new FlatRecordWriter(adapter.stateEngine, schemaIdMapper));
```

dkoszewnik: Oh nice tip, thanks!

PR #402, opened 2019-02-23
Name all threads started by Hollow

ghost:
* [x] audit thread creation; ensure all threads:
  * [x] are daemon threads
  * [x] have a name
  * [x] name follows the same convention
* [x] fix UISession cleanup thread ending after one sweep

### DIFF
--- a/hollow-ui-tools/src/main/java/com/netflix/hollow/ui/HollowUISession.java
+++ b/hollow-ui-tools/src/main/java/com/netflix/hollow/ui/HollowUISession.java
@@ -16,13 +16,16 @@
  */
 package com.netflix.hollow.ui;
 
-import java.util.Iterator;
+import static com.netflix.hollow.core.util.Threads.daemonThread;
+import static java.util.concurrent.TimeUnit.MINUTES;
+
 import java.util.Map;
 import java.util.Random;
 import java.util.concurrent.ConcurrentHashMap;
 import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+
 import org.eclipse.jetty.server.Response;
 
 public class HollowUISession {
@@ -84,24 +87,20 @@ public class HollowUISession {
     }
 
     static {
-        Thread sessionCleanupThread = new Thread(new Runnable() {
-            public void run() {
-                Iterator<Map.Entry<Long, HollowUISession>> iter = sessions.entrySet().iterator();
-                while(iter.hasNext()) {
-                    Map.Entry<Long, HollowUISession> entry = iter.next();
-                    if(entry.getValue().lastAccessed + SESSION_ABANDONMENT_MILLIS < System.currentTimeMillis())
-                        iter.remove();
-                }
-
-                try {
-                    Thread.sleep(60000L);
-                } catch (InterruptedException e) { }
-            }
-        });
-
-        sessionCleanupThread.setDaemon(true);
-        sessionCleanupThread.start();
+        daemonThread(HollowUISession::cleanupSessions, HollowUISession.class, "session-cleanup")
+                .start();
     }
 
+    private static void cleanupSessions() {
+        Iterator<Map.Entry<Long, HollowUISession>> iter = sessions.entrySet().iterator();
+        while (iter.hasNext()) {
+            Map.Entry<Long, HollowUISession> entry = iter.next();
+            if (entry.getValue().lastAccessed + SESSION_ABANDONMENT_MILLIS < System.currentTimeMillis())
+                iter.remove();
+        }
 
+        try {
+            MINUTES.sleep(1);
+        } catch (InterruptedException e) { }
+    }
 }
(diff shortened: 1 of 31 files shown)


hollow-ui-tools/src/main/java/com/netflix/hollow/ui/HollowUISession.java
```diff
@@ -85,24 +87,23 @@ public static HollowUISession getSession(HttpServletRequest req, HttpServletResp
...
-                        iter.remove();
-                }
-
-                try {
-                    Thread.sleep(60000L);
-                } catch (InterruptedException e) { }
-            }
-        });
-
+        Thread sessionCleanupThread = new Thread(HollowUISession::cleanupSessions, "hollow | ui-session-cleanup");
```
akhaku: Minor: might be nice to refactor thread naming into a static method to make it easier to make them consistent. Also I'm not so sure about using vertical bar as a separator - maybe there are log systems using it as a separator, how about just `hollow-`?

hollow-ui-tools/src/main/java/com/netflix/hollow/ui/HollowUISession.java
```diff
@@ -85,24 +87,23 @@ public static HollowUISession getSession(HttpServletRequest req, HttpServletResp
...
         sessionCleanupThread.setDaemon(true);
         sessionCleanupThread.start();
     }
 
+    private static void cleanupSessions() {
+        while (true) {
```
akhaku: Also minor - `!Thread.currentThread().isInterrupted()`?

akhaku (approved):
What a hero!

hollow-ui-tools/src/main/java/com/netflix/hollow/ui/HollowUISession.java
PaulSandoz: Let's create a method that accepts a String and returns a new daemon thread. That method could then be reused in many places. That method could accept a class associated with the thread as well as a name from which the thread name is composed. e.g.:

```java
Thread createDaemonThread(Runnable r, Class<?> context, String description) {
    Thread t = new Thread(r, createThreadName(context, description));
    t.setDaemon(true);
    return t;
}
```

FWIW if the thread creating the new thread is a daemon thread then new thread will be a daemon thread too.

hollow-ui-tools/src/main/java/com/netflix/hollow/ui/HollowUISession.java
```diff
@@ -85,24 +87,23 @@ public static HollowUISession getSession(HttpServletRequest req, HttpServletResp
...
+            Iterator<Map.Entry<Long, HollowUISession>> iter = sessions.entrySet().iterator();
```
PaulSandoz: This can be shrunk down to:
```java
sessions.values()
    .removeIf(s-> s.lastAccessed + SESSION_ABANDONMENT_MILLIS < System.currentTimeMillis());
```

ghost: Did it.

hollow-ui-tools/src/main/java/com/netflix/hollow/ui/HollowUISession.java
ghost: Done'd

hollow/src/main/java/com/netflix/hollow/core/util/Threads.java
```diff
@@ -0,0 +1,57 @@
...
+        Thread t = new Thread(r, namer.name(context, description));
+        t.setDaemon(true);
+        return t;
+    }
+
+    @FunctionalInterface
+    public interface ThreadNamer {
+        String platform();
+
+        default String name(Class<?> context, String description) {
```
PaulSandoz: AFAICT this is never overridden. Given it's an internal API you can simplify by removing the `ThreadNamer` type and replacing the parameter with `String platform`. The composition of the thread name can be a static method. If you want an exit define another method `daemonThread(Runnable r, String name)` that the others defer to:

```java
  public static Thread daemonThread(Runnable r, Class<?> context, String description) {
    return daemonThread(r, "hollow", context, description);
  }

  public static Thread daemonThread(Runnable r, String platform, Class<?> context, String description) {
    return daemonThread(r, name(platform, context, description));
  }

  public static Thread daemonThread(Runnable r, String name) {
    Thread t = new Thread(r, name);
    t.setDaemon(true);
    return t;
  }

  static String name(String platform, Class<?> context, String description) { ... }
```

PaulSandoz (approved):
Looks good. I think you can simplify the daemon thread construction as suggested in the comment on the related code.

My hope at some point is many uses of `SimultaneousExecutor` can be replaced with parallel streams. That does not negate this work as there will still be other long running threads that need to be named. 

When using parallel streams a shared fork join pool will be used (by default the common pool unless we decide to allocate our own).  That will mean less thread construction and more efficient parallel execution (since the fork join pool is a work stealing scheduler and sizes more optimally to the runtime configuration).  However, it does mean, by default, a computation cannot be named so it is visible in the thread name via a thread dump.

ghost:
> However, it does mean, by default, a computation cannot be named so it is visible in the thread name via a thread dump.

That doesn't feel like a big downside for me for computations that implicitly complete. It's kind of nice (with this change merged) that a thread dump during a long computation (or log lines written from its thread) have some additional context. That seems like the biggest loss when switching to parallel streams. But feels like the removal of complexity and lines of code would justify it.

PR #405, opened 2019-02-28
Check ordinal and byte data size limits.

PaulSandoz:
Limits on `ByteArrayOrdinalMap` are not checked leading to possible data corruption for large data sets that go beyond those limits but keep within the limit of the the maximum bucket array size. (The maximum ordinal value is 2^29 -1, and the maximum number of serialized bytes is 2^35).


ghost (approved):
💯

PR #406, opened 2019-03-01
Incremental producer restore should not preserve record hash positions

dkoszewnik:
When the incremental producer restarted and restored `SET` and/or `MAP` types, the preservation of hash _positions_ was causing a lot of false diffs, because the _actual_ hash codes were lost to history.  Since hash codes are encoded into the records in the write state, this caused the byte sequence representing these records in the write state to differ when the _same_ records were legitimately added back.

"Preserving hash positions" is only a useful concept in the context of the deprecated `HollowObjectHashCodeFinder`. 

I am hesitant to change the behavior of the `HollowWriteStateCreator`, so I am adding a new method to access the new behavior.  I am comfortable with changing the behavior of the `HollowIncrementalProducer`, because:

* it is unlikely that anyone is using a `HollowObjectHashCodeFinder` in the context of an incremental producer, and
* `HollowIncrementalProducer` is still marked as a beta API.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
@@ -295,7 +295,7 @@ public class HollowProducer {
 
     HollowProducer.ReadState hardRestore(long versionDesired, HollowConsumer.BlobRetriever blobRetriever) {
         return restore(versionDesired, blobRetriever,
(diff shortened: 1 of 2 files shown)


PaulSandoz (approved):
Looks good. I agree with your assessment about `HollowObjectHashCodeFinder`, especially since it is deprecated and hopefully possible to remove at some point in the near future.

PR #408, opened 2019-03-06
HollowTypeAPI implements HollowRecordDelegate.

kineshsatiya:
Hollow has two implementations for `HollowRecordDelegate` interface. An implementation of this interface is used by the `HollowFactory` when instantiating a `HollowRecord`. All the implementations of `HollowRecord` use the delegate instances to read data from the corresponding `HollowTypeDataAccess`. The flavors of the delegate interfaces are `CachedImpl` and `LookupImpl`. `LookupImpl` delegates all the reads to `HollowTypeAPI`. In order to reduce code generation and simplify the code generation, `HollowTypeAPI` now directly implements `HollowRecordDelegate` and the factory uses `HollowTypeAPI` as a delegate.


hollow/src/main/java/com/netflix/hollow/api/custom/HollowTypeAPI.java
```diff
@@ -29,20 +30,14 @@
...
-    protected final HollowAPI api;
     protected final HollowTypeDataAccess typeDataAccess;
 
-    protected HollowTypeAPI(HollowAPI api, HollowTypeDataAccess typeDataAccess) {
-        this.api = api;
+    protected HollowTypeAPI(HollowTypeDataAccess typeDataAccess) {
         this.typeDataAccess = typeDataAccess;
     }
 
-    public HollowAPI getAPI() {
```
PaulSandoz: Removing this method is a breaking change. Did you remove to simplify or was it required for all type API implementations to implement their corresponding delegate?

PaulSandoz:
I have a general concern on the change and removal of public API artifacts before they have been marked as deprecated, rather than just focusing on not generating a client API artifact that is an implementation detail.
Did you verify if these changes are compatible with generated client API code client prior to this change (since the client generated code generated by Hollow v1 can be can be operated on by a consumer using Hollow v2).

PaulSandoz:
To answer my own question. I believe the changes to the protected constructor of `HollowObjectTypeAPI` will break existing generated client API code.

For example, the test `HollowPrimitiveTypesAPIGeneratorTest` will generate the following:

```java
@SuppressWarnings("all")
public class ActorTypeAPI extends HollowObjectTypeAPI {

    private final ActorDelegateLookupImpl delegateLookupImpl;

    public ActorTypeAPI(PrimitiveTypeTestAPI api, HollowObjectTypeDataAccess typeDataAccess) {
        super(api, typeDataAccess, new String[] {
            "name",
            "role"
        });
        this.delegateLookupImpl = new ActorDelegateLookupImpl(this);
    }
    ...
```

It relies on the constructor of `HollowObjectTypeAPI` that accepts `HollowAPI` as the first parameter.

hollow/src/main/java/com/netflix/hollow/core/schema/HollowObjectSchema.java
```diff
@@ -108,6 +108,12 @@ public int getPosition(String fieldName) {
         return index.intValue();
     }
 
+    public String[] getFieldNames() {
+        String[] copyFieldNames = new String[fieldNames.length];
```
PaulSandoz: You can use `return fieldNames.clone();`

hollow/src/main/java/com/netflix/hollow/api/codegen/api/TypeAPIListJavaGenerator.java
```diff
@@ -52,30 +52,28 @@ public String generate() {
...
+        builder.append("import " + HollowListDelegate.class.getName() + ";\n");
 
         builder.append("\n@SuppressWarnings(\"all\")\n");
-        builder.append("public class ").append(className).append(" extends HollowListTypeAPI {\n\n");
+        builder.append("public class ").append(className).append(" extends HollowListTypeAPI implements "+ delegateInterfaceName(schema) +"{\n\n");
 
-        builder.append("    private final ").append(delegateLookupClassname(schema)).append(" delegateLookupImpl;\n\n");
-
-        builder.append("    public ").append(className).append("(").append(apiClassname).append(" api, HollowListTypeDataAccess dataAccess) {\n");
+        builder.append("    public ").append(className).append("(").append(apiClassname). append(" api, HollowListTypeDataAccess dataAccess) {\n");
```
PaulSandoz, line 60: Nit: rogue space `apiClassname). append`

PaulSandoz (approved):

PR #413, opened 2019-03-28
Deprecate method that does unexpected things and document alternative

akhaku:
Also minor cleanup.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/schema/HollowObjectSchema.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/schema/HollowObjectSchema.java
@@ -54,7 +54,7 @@ public class HollowObjectSchema extends HollowSchema {
     public HollowObjectSchema(String schemaName, int numFields, PrimaryKey primaryKey) {
         super(schemaName);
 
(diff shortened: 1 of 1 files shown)


PaulSandoz (commented):
Why not modify `addField(String fieldName, FieldType fieldType, String referencedType, HollowTypeReadState referencedTypeState)` to call `setReferencedTypeState` if `referencedTypeState` is non-`null`?

akhaku:
I started down that path initially, then attempted to write a test for it, which got me thinking about the use cases for it. There is no case today where we call that method with a non-null referencedTypeState, and it doesn't quite fit that a HollowObjectSchema gets a reference to a type state for another type while you're adding a field to it. It makes more sense that the type states get set when deserializing a snapshot with a full type state, and that is already exposed by the public `setReferencedTypeState`.
On a side note, AFAICT the referenced type state being accessible from the schema is really just a shortcut for internal use - I'm not convinced it's a good idea to have a reference to the data from the schema.

PaulSandoz:
That makes sense, i suspect no one uses this method, and I agree with you about the binding of type state to the schema.

PaulSandoz (approved):

PR #414, opened 2019-04-03
Final incremental producer API

PaulSandoz:
- The first commit adds incremental cycle support into `HollowProducer`.
- The second commit refactors much of the functionality of `HollowProducer` into a package private class `AbstractHollowProducer` from which `HollowProducer` and `HollowProducer.Incremental` extend from.  An incremental producer is built from the same `HollowProducer` builder API. 
- The existing beta API is deprecated.


rpalcolea (approved):
This is awesome! It's great to see the abstract producer and how incremental is evolving over time

PR #422, opened 2019-05-10
Don't fail with NPE when DuplicateDDValidator for missing type

akhaku:
If the user added a DuplicateDataDetectionValidator for a
missing type we would fail with an unhelpful NPE. Instead,
provide a useful and vaguely snarky error message.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
@@ -49,10 +49,15 @@ public class DuplicateDataDetectionValidator implements ValidatorListener {
     private static final String DUPLICATE_KEYS_FOUND_ERRRO_MSG_FORMAT =
             "Duplicate keys found for type %s. Primarykey in schema is %s. "
                     + "Duplicate IDs are: %s";
-    private static final String NO_PRIMARY_KEY_ERRRO_MSG_FORMAT =
+    private static final String NO_PRIMARY_KEY_ERROR_MSG_FORMAT =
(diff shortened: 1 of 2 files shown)


locasity (approved):

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -49,10 +49,13 @@
...
             "Duplicate keys found for type %s. Primarykey in schema is %s. "
                     + "Duplicate IDs are: %s";
-    private static final String NO_PRIMARY_KEY_ERRRO_MSG_FORMAT =
+    private static final String NO_PRIMARY_KEY_ERROR_MSG_FORMAT =
             "DuplicateDataDetectionValidator defined but unable to find primary key "
                     + "for data type %s. Please check schema definition.";
-    private static final String NOT_AN_OBJECT_ERROR_MSGR_FORMAT =
+    private static final String NO_SCHEMA_FOUND_MSG_FORMAT =
+            "DuplicateDataDetectionValidator defined for data type %s but schema not "
+                    + "found - did you forget to pass it to initializeDataModel?";
```
PaulSandoz: How about:
```
DuplicateDataDetectionValidator defined for data type %s but schema not found.
Please check that the HollowProducer is initialized with the data type's schema (see initializeDataModel)
```

PaulSandoz (approved):

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
akhaku: Sure why not

PR #426, opened 2019-06-03
Bounds checks when removing ordinal from index.

PaulSandoz:
A `HollowPrimaryKeyIndex` will perform a delta-update on the index if the previous and current state results in the same hash table size, same bits per element, and the number of removals is within 10% of the previous states cardinality.

Bounds checks are explicitly performed for ordinal removal in case there is an issue with the hash table.  A test is added to exercise removal and addition.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/index/HollowPrimaryKeyIndex.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/index/HollowPrimaryKeyIndex.java
@@ -38,6 +38,8 @@ import java.util.BitSet;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * A HollowPrimaryKeyIndex is the go-to mechanism for indexing and querying data in a Hollow blob.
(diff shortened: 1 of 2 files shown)


hollow/src/main/java/com/netflix/hollow/core/index/HollowPrimaryKeyIndex.java
```diff
@@ -506,6 +503,27 @@ private void deltaUpdate(int hashTableSize, int bitsPerElement) {
...
+        int startBucket = hashCode & hashMask;
+        int bucket = startBucket;
+        long value;
+        do {
+            value = hashedArray.getElementValue((long)bucket * (long)bitsPerElement, bitsPerElement);
+            if (prevOrdinal + 1 == value) {
+                return bucket;
+            }
+            bucket = (bucket + 1) & hashMask;
+        } while (value != 0 && bucket != startBucket);
```
akhaku, line 540: this is meant to be zero right, not ORDINAL_NONE?

PaulSandoz: Yes, it's meant to be zero. A zero `long` value in the hash map corresponds to an empty ordinal.  
e.g. this is code further down when adding new ordinals to find an empty bucket:
```
            while(hashedArray.getElementValue((long)bucket * (long)bitsPerElement, bitsPerElement) != 0)
                bucket = (bucket + 1) & hashMask;
```
(Note although this lacks a bound check for wrapping around the hash table is sized such that it is guaranteed that empty buckets will exist for the cardinality, so the loop will terminate.)

PR #428, opened 2019-06-10
Allow creating indices when schema name != generated class name

### LINKED ISSUE #427: SchemaNotFoundException when using UniqueKeyIndex
### When hollow-generated class' name is postfixed, UniqueKeyIndex cannot find HollowSchema.  

Example of HollowAPIGenerator creation:
`  new HollowAPIGenerator.Builder().withClassPostfix("Hollow")...
`

In this case, the name of hollow-generated class from https://github.com/Netflix/hollow/releases/tag/v4.1.0 example would be _MovieHollow_.
Schema name would still be just _Movie_. 

In this case, UniqueKeyIndex **cannot** be built using this piece of code:
`UniqueKeyIndex.from(consumer, Movie.class).bindToPrimaryKey()...`

This is caused by _bindToPrimaryKey_ method implementation, which seeks for HollowSchema using _uniqueType_'s simple name, which in this case is _MovieHollow_, whereas schema name is _Movie_.

Mentioned method seeks for HollowSchema using the following piece of code:

` String primaryTypeName = uniqueType.getSimpleName();
HollowSchema schema = consumer.getStateEngine().getNonNullSchema(primaryTypeName);
`

akhaku:
- generate classes with @HollowTypeName if name does not match schema
- read the schema name using any @HollowTypeName when inspecting types during index creation

Fixes #427.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/codegen/objects/HollowObjectJavaGenerator.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/codegen/objects/HollowObjectJavaGenerator.java
@@ -39,6 +39,7 @@ import com.netflix.hollow.core.HollowDataset;
 import com.netflix.hollow.core.index.key.PrimaryKey;
 import com.netflix.hollow.core.schema.HollowObjectSchema;
 import com.netflix.hollow.core.schema.HollowObjectSchema.FieldType;
+import com.netflix.hollow.core.write.objectmapper.HollowTypeName;
 import com.netflix.hollow.tools.stringifier.HollowRecordStringifier;
 import java.util.LinkedHashMap;
 import java.util.Map;
@@ -89,6 +90,7 @@ public class HollowObjectJavaGenerator extends HollowConsumerJavaFileGenerator {
     public String generate() {
         StringBuilder classBuilder = new StringBuilder();
         appendPackageAndCommonImports(classBuilder, apiClassname);
+        boolean requiresHollowTypeName = !className.equals(schema.getName());
 
         classBuilder.append("import " + HollowConsumer.class.getName() + ";\n");
         if (schema.getPrimaryKey() != null && schema.getPrimaryKey().numFields() > 1) {
@@ -99,12 +101,18 @@ public class HollowObjectJavaGenerator extends HollowConsumerJavaFileGenerator {
         }
         classBuilder.append("import " + HollowObject.class.getName() + ";\n");
         classBuilder.append("import " + HollowObjectSchema.class.getName() + ";\n");
+        if (requiresHollowTypeName) {
+            classBuilder.append("import " + HollowTypeName.class.getName() + ";\n");
+        }
         if (config.isUseVerboseToString()) {
             classBuilder.append("import " + HollowRecordStringifier.class.getName() + ";\n");
         }
         classBuilder.append("\n");
 
         classBuilder.append("@SuppressWarnings(\"all\")\n");
+        if (requiresHollowTypeName) {
+            classBuilder.append("@" + HollowTypeName.class.getSimpleName() + "(name=\"" + schema.getName() + "\")\n");
+        }
         classBuilder.append("public class " + className + " extends HollowObject {\n\n");
 
         appendConstructor(classBuilder);
--- a/hollow/src/main/java/com/netflix/hollow/api/consumer/index/SelectFieldPathResultExtractor.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/consumer/index/SelectFieldPathResultExtractor.java
@@ -24,6 +24,7 @@ import com.netflix.hollow.core.HollowDataset;
 import com.netflix.hollow.core.index.FieldPaths;
 import com.netflix.hollow.core.schema.HollowObjectSchema;
 import com.netflix.hollow.core.schema.HollowSchema;
+import com.netflix.hollow.core.write.objectmapper.HollowObjectTypeMapper;
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.MethodType;
@@ -71,7 +72,7 @@ final class SelectFieldPathResultExtractor<T> {
     static <T> SelectFieldPathResultExtractor<T> from(
             Class<? extends HollowAPI> apiType, HollowDataset dataset, Class<?> rootType, String fieldPath,
             Class<T> selectType) {
-        String rootTypeName = rootType.getSimpleName();
+        String rootTypeName = HollowObjectTypeMapper.getDefaultTypeName(rootType);
         FieldPaths.FieldPath<FieldPaths.FieldSegment> fp =
                 FieldPaths.createFieldPathForHashIndex(dataset, rootTypeName, fieldPath);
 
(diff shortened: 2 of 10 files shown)


hollow/src/test/java/com/netflix/hollow/api/codegen/AbstractHollowAPIGeneratorTest.java
```diff
@@ -55,7 +61,15 @@ void assertNonEmptyFileExists(String relativePath) throws IOException {
         assertTrue("File at " + relativePath + " should exist", f.exists() && f.length() > 0L);
     }
 
-    void assertFileDoesNotExist(String relativePath) throws IOException {
+    void assertClassHasHollowTypeName(String clazz, String typeName) throws IOException, ClassNotFoundException {
```
akhaku, line 64: Couldn't find a good way to test what we really want (that we can create generated indices) so went with the best I could - verify that a HollowTypeName is set.

hollow/src/main/java/com/netflix/hollow/api/codegen/objects/HollowObjectJavaGenerator.java
```diff
@@ -105,6 +106,9 @@ public String generate() {
         classBuilder.append("\n");
 
         classBuilder.append("@SuppressWarnings(\"all\")\n");
+        if (!className.equals(schema.getName())) {
+            classBuilder.append("@" + HollowTypeName.class.getName() + "(name=\"" + schema.getName() + "\")\n");
```
PaulSandoz: That's gonna use the fully qualified name.  Can you declare an import?

hollow/src/test/java/com/netflix/hollow/api/codegen/AbstractHollowAPIGeneratorTest.java
PaulSandoz: That seems reasonable.  We can work on an additional fix to the indexes.  If you look in the test class `com.netflix.hollow.api.consumer.index.DataModel.Consumer` we can add something explicit.

hollow/src/main/java/com/netflix/hollow/api/codegen/objects/HollowObjectJavaGenerator.java
akhaku: Sure, this was the lazy way to go

hollow/src/test/java/com/netflix/hollow/api/codegen/AbstractHollowAPIGeneratorTest.java
akhaku: I don't think I understand your comment - is there a way to test the generated index, or are you suggesting adding a test to make sure that the HollowTypeName is picked up by indexes?

hollow/src/main/java/com/netflix/hollow/api/consumer/index/SelectFieldPathResultExtractor.java
```diff
@@ -71,7 +72,7 @@ static IllegalArgumentException incompatibleSelectType(Class<?> selectType, Stri
     static <T> SelectFieldPathResultExtractor<T> from(
             Class<? extends HollowAPI> apiType, HollowDataset dataset, Class<?> rootType, String fieldPath,
             Class<T> selectType) {
-        String rootTypeName = rootType.getSimpleName();
+        String rootTypeName = HollowObjectTypeMapper.getDefaultTypeName(rootType);
```
akhaku, line 75: I had some trouble testing the changes in this file, Paul's eyes would be useful here

PaulSandoz: Here's an example:
```
HashIndexSelect<Movie, Actor, String> index = HashIndex.from(consumer, Movie.class)
    .selectField("cast.element.actor", Actor.class)
    .usingPath("title.value", String.class);

String title = "The Imitation Game";
Stream<Actor> m = index.findMatch(title);  
```
It's a bit like:
```
SELECT Actor FROM Movie
WHERE title='The Imitation Game';
```
So you need to add another annotated test class for selection type referenced from the top type.

hollow/src/test/java/com/netflix/hollow/api/consumer/index/HashIndexTest.java
```diff
@@ -57,7 +58,11 @@ public static void setup() {
 
             for (int i = 0; i < 100; i++) {
                 ws.add(new DataModel.Producer.TypeA(1, "TypeA" + i));
+                ws.add(new DataModel.Producer.TypeB(1, "TypeB" + i));
```
akhaku, line 61: Something odd is going  on - somehow adding this here causes TestManyMatches#testWithSelectGenericHollowObject,testWithSelect to fail. Digging further...

akhaku: I can reproduce using an old-style manually constructed HollowHashIndex. It only happens when I pass in a select field path for the string and I _don't_ explicitly as for `.value` as the select field path. This is explicitly disallowed in the new-style hash indexes - need to sync up with Paul on this.

PaulSandoz: The field path parsing should be consistent with `com.netflix.hollow.core.index.HollowHashIndex` and `com.netflix.hollow.api.consumer.index.HashIndex`. See use of `FieldPaths.createFieldPathForHashIndex`.  Auto-expansion is not supported for the select and match field paths of a hash index.

hollow/src/test/java/com/netflix/hollow/api/codegen/AbstractHollowAPIGeneratorTest.java
PaulSandoz: I found the cause.  Select path never quite worked correctly, it currently requires the path have a last segment reference a field value, and the ordinal returned will correspond to the enclosing record.

The problematic line is at https://github.com/Netflix/hollow/blob/master/hollow/src/main/java/com/netflix/hollow/core/index/traversal/TraversalTreeBuilder.java#L103:
```java
                            if(schema.getFieldType(fieldSchemaPositions[i]) == FieldType.REFERENCE) {
                                fieldSchemaPositions[i] = -1;
                                fieldTypeDataAccess[i] = getChildDataAccess(typeDataAccess, pathElement);
                            } else {
                                fieldSchemaPositions[i] = schema.getPosition(pathElement);
                                fieldTypeDataAccess[i] = typeDataAccess;
                            }
```

It should be:
```java
                            if(schema.getFieldType(pathElement) == FieldType.REFERENCE) {
```

hollow/src/test/java/com/netflix/hollow/api/consumer/index/HashIndexTest.java
PaulSandoz: I found the cause.  Select path never quite worked correctly, it currently requires the path have a last segment reference a field value, and the ordinal returned will correspond to the enclosing record.

The problematic line is at https://github.com/Netflix/hollow/blob/master/hollow/src/main/java/com/netflix/hollow/core/index/traversal/TraversalTreeBuilder.java#L103:
```java
                            if(schema.getFieldType(fieldSchemaPositions[i]) == FieldType.REFERENCE) {
                                fieldSchemaPositions[i] = -1;
                                fieldTypeDataAccess[i] = getChildDataAccess(typeDataAccess, pathElement);
                            } else {
                                fieldSchemaPositions[i] = schema.getPosition(pathElement);
                                fieldTypeDataAccess[i] = typeDataAccess;
                            }
```

It should be:
```java
                            if(schema.getFieldType(pathElement) == FieldType.REFERENCE) {
```

akhaku: nice - great find!

PaulSandoz (approved):

PR #429, opened 2019-06-12
Maintain correct read states after integrity check

akhaku:
Previously, we were swapping the current and pending read states (after applying the delta and reverse delta, respectively), for efficiency. However, this is not the correct thing to do when the two read states don't have the same schemas. For those cases, apply the delta back onto the pending state, and apply the reverse delta back onto the current state.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
@@ -654,16 +654,16 @@ abstract class AbstractHollowProducer {
 
     /**
      * Given these read states
-     * <p>
-     * * S(cur) at the currently announced version
-     * * S(pnd) at the pending version
-     * <p>
-     * Ensure that:
-     * <p>
-     * S(cur).apply(forwardDelta).checksum == S(pnd).checksum
-     * S(pnd).apply(reverseDelta).checksum == S(cur).checksum
      *
-     * @return updated read states
+     * * S(cur) at the currently announced version
+     * * S(pnd) = empty read state
+     *
+     * 1. Read in the snapshot artifact to initialize S(pnd)
+     * 2. Ensure that:
+     *   - S(cur).apply(forwardDelta).checksum == S(pnd).checksum
+     *   - S(pnd).apply(reverseDelta).checksum == S(cur).checksum
+     *
+     * @return S(cur) and S(pnd)
      */
     private ReadStateHelper checkIntegrity(
             ListenerSupport.Listeners listeners, ReadStateHelper readStates, Artifacts artifacts) throws Exception {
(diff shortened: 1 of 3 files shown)


akhaku:
I see a separate but related bug in #425 - will fix in a separate change.

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -702,7 +702,9 @@ private ReadStateHelper checkIntegrity(
                     if (!reverseChecksum.equals(currentChecksum)) {
                         throw new HollowProducer.ChecksumValidationException(HollowProducer.Blob.Type.REVERSE_DELTA);
                     }
-                    result = readStates.swap();
+                    // undo the changes we made to the read states
```
dkoszewnik: Deltas for large datasets can take a relatively long time to apply.  It seems wasteful to do this every time.

There must be a better way to accomplish this -- e.g. detect when the schemas differ, then apply this logic selectively in those rare cases.

dkoszewnik (changes requested):

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
akhaku: Hm, yeah comparing all the object schemas would work. I'll probably have to add hashCode to make the comparisons easier but it's probably worth it.

hollow/src/main/java/com/netflix/hollow/core/read/engine/HollowReadStateEngine.java
```diff
@@ -288,6 +288,18 @@ public boolean updatedLastCycle() {
         return false;
     }
 
+    public boolean hasIdenticalSchemas(HollowReadStateEngine other) {
+        if (!other.getAllTypes().equals(getAllTypes())) {
+            return false;
+        }
+        for (String type : getAllTypes()) {
+            if (!getSchema(type).equals(other.getNonNullSchema(type))) {
```
PaulSandoz, line 296: Minor point. Since the prior check determines that both read states have the same set of types there is no need to use `getNonNullSchema`, its usage stands out and one wonders if a `SchemaNotFoundException` can be thrown.

Alas that prior check is somewhat fragile given `getAllTypes` returns a `Collection` and not a `Set` but we just have to assume the implementation returns a set or a sorted sequence.

PaulSandoz (approved):

hollow/src/main/java/com/netflix/hollow/core/read/engine/HollowReadStateEngine.java
dkoszewnik: Can this functionality be moved up to the `HollowDataset`?  I've written this or similar a number of times because it didn't exist there.

dkoszewnik (commented):
LGTM, but would suggest moving the `hasIdenticalSchemas` functionality up to the `HollowDataset`, since this is broadly applicable (I've needed this more than a few times).

PR #431, opened 2019-06-28
Ignore failed transitions when no double snapshot

PaulSandoz:
This is a temporary solution until it is decided what to do with the feature, likely removing it or enhancing to make it more configurable/accurate/actionable.


dkoszewnik (approved):
Awesome, thanks Paul.

PR #435, opened 2019-07-13
Blob write to temp file then rename instead of write to destination

Sunjeet:
Protection against corrupt snapshot getting cached on disk because of partial write.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/consumer/fs/HollowFilesystemBlobRetriever.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/consumer/fs/HollowFilesystemBlobRetriever.java
@@ -26,6 +26,8 @@ import java.io.OutputStream;
 import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.UUID;
 
 public class HollowFilesystemBlobRetriever implements HollowConsumer.BlobRetriever {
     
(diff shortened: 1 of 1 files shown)


hollow/src/main/java/com/netflix/hollow/api/consumer/fs/HollowFilesystemBlobRetriever.java
```diff
@@ -196,16 +198,20 @@ public InputStream getInputStream() throws IOException {
...
-                    OutputStream os = Files.newOutputStream(path)
+                    OutputStream os = Files.newOutputStream(tempPath)
             ) {
                 byte buf[] = new byte[4096];
                 int n;
                 while (-1 != (n = is.read(buf)))
                     os.write(buf, 0, n);
             }
 
+            File f = tempPath.toFile();
```
PaulSandoz: See `Files.move(tempPath, path, REPLACE_EXISTING)`

hollow/src/main/java/com/netflix/hollow/api/consumer/fs/HollowFilesystemBlobRetriever.java
```diff
@@ -196,16 +198,20 @@ public InputStream getInputStream() throws IOException {
         @Override
         public InputStream getInputStream() throws IOException {
 
+            Path tempPath = Paths.get(path.toString() + "-" + UUID.randomUUID().toString());
```
PaulSandoz: See `path.resolveSibling(path.getName(0) + "-" + UUID....`

Sunjeet: Done, thanks

PaulSandoz (approved):

PR #439, opened 2019-09-10
Control fallback for local blob store.

PaulSandoz:
Allow no fallback from local blob store if there exists a snapshot
in the local blob store.


toolbear:
I like it. I was originally going to suggest calling the flag `noRemote`, but I see now that this handles the case of using the fallback when the local blob store is empty. That nicely avoids the problems of users having to temporarily enable the fallback to prime their local cache.

ghost (approved):
I like it.

I was originally going to suggest calling the flag `noRemote`, but I see now that this handles the case of using the fallback when the local blob store is empty. That nicely avoids the problems of users having to temporarily enable the fallback to prime their local cache.

dkoszewnik (approved):

PR #447, opened 2020-01-24
Allow restoring producer to non-exact version

akhaku:
For cases where users decide to pin to a non-exact version,
we want to be able to restore the version before the
provided version.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
@@ -264,31 +264,25 @@ abstract class AbstractHollowProducer {
             if (versionDesired != HollowConstants.VERSION_NONE) {
                 HollowConsumer client = HollowConsumer.withBlobRetriever(blobRetriever).build();
                 client.triggerRefreshTo(versionDesired);
-                if (client.getCurrentVersionId() == versionDesired) {
-                    readState = ReadStateHelper.newReadState(client.getCurrentVersionId(), client.getStateEngine());
-                    readStates = ReadStateHelper.restored(readState);
+                readState = ReadStateHelper.newReadState(client.getCurrentVersionId(), client.getStateEngine());
+                readStates = ReadStateHelper.restored(readState);
 
-                    // Need to restore data to new ObjectMapper since can't restore to non empty Write State Engine
-                    Collection<HollowSchema> schemas = objectMapper.getStateEngine().getSchemas();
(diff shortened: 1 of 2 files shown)


hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -264,31 +264,25 @@ public void initializeDataModel(HollowSchema... schemas) {
...
-                    Collection<HollowSchema> schemas = objectMapper.getStateEngine().getSchemas();
-                    HollowWriteStateEngine writeEngine = hashCodeFinder == null
-                            ? new HollowWriteStateEngine()
-                            : new HollowWriteStateEngine(hashCodeFinder);
-                    HollowWriteStateCreator.populateStateEngineWithTypeWriteStates(writeEngine, schemas);
-                    HollowObjectMapper newObjectMapper = new HollowObjectMapper(writeEngine);
-                    if (hashCodeFinder != null) {
-                        newObjectMapper.doNotUseDefaultHashKeys();
-                    }
+                readState = ReadStateHelper.newReadState(client.getCurrentVersionId(), client.getStateEngine());
```
dkoszewnik, line 267: A tradeoff to consider: if the `HollowConsumer` here somehow fails to retrieve the latest version, then this producer might inadvertently force a double snapshot on all clients after retrieving an earlier version.

dkoszewnik (approved):

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
ghost: When I reviewed the snapshot and delta applying code, it looks like they all rethrow exceptions. I can't think of any sources of silent failures there. So `triggerRefreshTo()` should end up throwing.

The scenario I can think of where the consumer doesn't arrive at the correct version *and* doesn't throw an exception is when the number of deltas necessary would exceed the max deltas config.

@dkoszewnik, can you think of any other scenario where that would happen?

ghost (approved):

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
akhaku: You mentioned this to me in person, but just to state here - that also implies that the producer is configured to not create snapshots every time, and that the number of deltas between snapshots is larger than the max deltas to apply on a single refresh.

PR #448, opened 2020-01-28
Producer Listener support refactored for code reuse

Sunjeet:
For reuse in abstractions that compose a producer (eg. transformer).

Note the following breaking changes:
`final class ListenerSupport` -> `final class ProducerListenerSupport extends ListenerSupport`
`public interface HollowProducerEventListener` -> `public interface HollowProducerEventListener extends EventListener`

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/ProducerListenerSupport.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/ProducerListenerSupport.java
@@ -68,8 +68,6 @@ final class ProducerListenerSupport extends ListenerSupport {
     }
 
     ProducerListenerSupport() {
-        super();
-
         // @@@ This is used only by HollowIncrementalProducer, and should be
         // separated out
         incrementalCycleListeners = new CopyOnWriteArraySet<>();
@@ -84,7 +82,7 @@ final class ProducerListenerSupport extends ListenerSupport {
     }
 
     ProducerListenerSupport(ProducerListenerSupport that) {
-        super();
+        super(that);
 
         // @@@ This is used only by HollowIncrementalProducer, and should be
         // separated out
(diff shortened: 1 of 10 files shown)


ghost (approved):

hollow/src/main/java/com/netflix/hollow/api/producer/ProducerListenerSupport.java
```diff
@@ -68,81 +67,45 @@ static boolean isValidListener(HollowProducerEventListener l) {
...
 
         // @@@ This is used only by HollowIncrementalProducer, and should be
         // separated out
         incrementalCycleListeners = new CopyOnWriteArraySet<>();
     }
 
-    ListenerSupport(ListenerSupport that) {
-        eventListeners = new CopyOnWriteArrayList<>(that.eventListeners);
+    ProducerListenerSupport(ProducerListenerSupport that) {
+        super();
```
akhaku: did you mean to pass in `that` to the `super` call here?

hollow/src/main/java/com/netflix/hollow/api/producer/ProducerListenerSupport.java
```diff
@@ -68,81 +67,45 @@ static boolean isValidListener(HollowProducerEventListener l) {
         return LISTENERS.stream().anyMatch(c -> c.isInstance(l));
     }
 
-    private final CopyOnWriteArrayList<HollowProducerEventListener> eventListeners;
-
-    ListenerSupport() {
-        eventListeners = new CopyOnWriteArrayList<>();
+    ProducerListenerSupport() {
+        super();
```
akhaku: Super minor, but I think the no-argument `super` call is implicit.

akhaku (approved):

hollow/src/main/java/com/netflix/hollow/api/producer/ProducerListenerSupport.java
Sunjeet: Yes, thanks for `that`

hollow/src/main/java/com/netflix/hollow/api/producer/ProducerListenerSupport.java
Sunjeet: Dropped it

PR #449, opened 2020-01-31
Resume partial blob downloads

Sunjeet:
This is useful for local dev on spotty internet connection

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/consumer/fs/HollowFilesystemBlobRetriever.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/consumer/fs/HollowFilesystemBlobRetriever.java
@@ -28,9 +28,12 @@ import java.io.OutputStream;
 import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.UUID;
+import java.nio.file.StandardOpenOption;
 
 public class HollowFilesystemBlobRetriever implements HollowConsumer.BlobRetriever {
+
+    public static final String PARTIAL_DOWNLOAD = "partial.";
+
     private final Path blobStorePath;
     private final HollowConsumer.BlobRetriever fallbackBlobRetriever;
     private final boolean noFallBackForExistingSnapshot;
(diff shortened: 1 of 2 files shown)


hollow/src/main/java/com/netflix/hollow/api/consumer/fs/HollowFilesystemBlobRetriever.java
```diff
@@ -223,12 +226,17 @@ public InputStream getInputStream() throws IOException {
...
             try(
-                    InputStream is = remoteBlob.getInputStream();
-                    OutputStream os = Files.newOutputStream(tempPath)
+                InputStream is = remoteBlob.getInputStream();
+                OutputStream os = Files.newOutputStream(tempPath, StandardOpenOption.CREATE, StandardOpenOption.APPEND);
             ) {
-                byte buf[] = new byte[4096];
+                // resume unfinished downloads
+                File outFile = new File(tempPath.toAbsolutePath().toString());
+                is.skip(outFile.length());
```
akhaku, line 237: I suppose this is better than we have currently, but I'm guessing this results in effectively the same behavior from a stream consumption perspective, i.e. we have to read the bytes (even if we drop them)?

hollow/src/main/java/com/netflix/hollow/api/consumer/fs/HollowFilesystemBlobRetriever.java
```diff
@@ -223,12 +226,17 @@ public InputStream getInputStream() throws IOException {
         @Override
         public InputStream getInputStream() throws IOException {
 
-            Path tempPath = path.resolveSibling(path.getName(path.getNameCount()-1) + "-" + UUID.randomUUID().toString());
+            Path tempPath = path.resolveSibling(PARTIAL_DOWNLOAD + path.getName(path.getNameCount()-1));
```
akhaku, line 229: nice, although FWIW most  partial download files I've seen end in `.part` or similar (rather than being prefixed)

hollow/src/test/java/com/netflix/hollow/api/consumer/LocalBlobStoreTest.java
```diff
@@ -1,14 +1,131 @@
...
+
+        // Consumer tries to refresh to published version, fails midway
+        HollowConsumer.Builder consumerBuilder = HollowConsumer
+                .withBlobRetriever(mockBlobRetriever)
+                .withLocalBlobStore(localDir);
+        consumerBuilder.noFallBackForExistingSnapshot = false;
+        HollowConsumer consumer = consumerBuilder.build();
+        boolean downloadInterrupted = false;
+        try {
+            consumer.triggerRefreshTo(testVersion);
```
akhaku, line 105: minor cleanup -  you could move the `fail` call here and not need the `downloadInterrupted` boolean at all

akhaku (approved):

hollow/src/main/java/com/netflix/hollow/api/consumer/fs/HollowFilesystemBlobRetriever.java
Sunjeet: There is some fragile filename parsing at other places in code, and it seemed to work better to have partial in the prefix as opposed to suffix. One eg. is checking if filename startswith "snapshot.".

PR #455, opened 2020-03-04
Introduce TypeFilter as a future HollowFilterConfig replacement [DABP-205]

ghost:
`TypeFilter` is intended to address limitations in `HollowFilterConfig` (see "Limitations" below).

Example when used in conjunction with `HollowConsumer.Builder`:

```java
consumer = newHollowConsumer()
  .withTypeFilter(filter -> filter
    .excludeAll()
    .include("Type1")
    .includeRecursive("Type2")
    .exclude("Type2", "field1"))
  .build();
```

or

```
filter = newTypeFilter()
  .excludeRecursive("Type")
  .build();
consumer = newHollowConsumer()
  .withTypeFilter(filter)
  .build();
```

## Limitations

`HollowFilterConfig` limitations:

* cannot mix inclusion and exclusion actions, it is either all inclusions or all exclusions
* expressing recursive actions requires that callers already have the datasets schema. This is a chicken-and-egg problem for some use cases (notably `HollowConsumer.Builder`)

`TypeFilter` limitations:

* ~doesn't yet implemented field-specific filtering (I might get that working before merging)~ implemented
* no ser/de to a String like `HollowFilterConfig` has

Intent for now is to leave both, improve upon `TypeFilter` over time, then deprecated `HollowFilterConfig` once `TypeFilter` is a suitable replacement for all use-cases.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
@@ -19,6 +19,8 @@ package com.netflix.hollow.api.consumer;
 import static com.netflix.hollow.core.util.Threads.daemonThread;
 import static java.util.concurrent.Executors.newSingleThreadExecutor;
 
+import com.netflix.hollow.PublicApi;
+import com.netflix.hollow.PublicSpi;
 import com.netflix.hollow.api.client.FailedTransitionTracker;
 import com.netflix.hollow.api.client.HollowAPIFactory;
 import com.netflix.hollow.api.client.HollowClientUpdater;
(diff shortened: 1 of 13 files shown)


hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
```diff
@@ -958,11 +1005,58 @@ public B withGeneratedAPIClass(Class<? extends HollowAPI> generatedAPIClass) {
...
         }
 
+        /**
+         * Specifies the type filter. Clears any filter config specified with {@code withFilterConfig()}
+         * @param typeFilter
+         * @return this builder
+         * @see #withTypeFilter(UnaryOperator)
+         * @see #withFilterConfig(HollowFilterConfig)
+         */
+        public B withTypeFilter(TypeFilter typeFilter) {
```
Sunjeet, line 1043: I wonder if it would be beneficial to document the limitations of `TypeFilter` and `HollowFilterConfig` in javadoc for `withTypeFilter` and `withHollowFilterConfig`. Depending on the time frame for `TypeFilter` feature completeness and when we can drop or at least deprecate `HollowFilterConfig`, this might reduce support cost for questions like "which one should I use?". Ofcourse we'd have to keep the `withTypeFilter` javadoc up to date as we develop it, or else the failure mode is that a feature is ready but not used.

ghost: I have field level filtering working, so I can deprecate `HollowFilterConfig` now.

👍 to documenting the limitations in the JavaDoc. `TypeFilter` now has only the 1 limitation: has no serialization/deserialization like `HollowFilterConfig`.

ghost: Done'd.

PR #458, opened 2020-03-10
Fix producer metric reporting non-published version

AlexandruGhergut:
On each cycle that is run, a `toVersion` is computed.
The `currentVersion` metric of the producer is updated on each cycle to the above version. I find that misleading because when there is NO state change, there will be no new version published but the metric will still be updated, pointing to a non-existent version.

If I understood correctly, the producer's `readState` will be null when there is no state change so I removed the else branch in which the version update happened.

Also added a gradle setting to build the subprojects in parallel.

**Use-case for this metric**: we want to add an alert for consumers that are out-of-sync with the producer for a given amount of time. For that we look at the consumer's reported version and at the producer's version.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,0 +1,1 @@
+org.gradle.parallel=true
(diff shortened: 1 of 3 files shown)


AlexandruGhergut:
ping @toolbear @rpalcolea @akhaku

AlexandruGhergut:
Closing this in favor of https://github.com/Netflix/hollow/pull/459

PR #461, opened 2020-04-22
Specific msg when NPE on index creation

Sunjeet:


hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
```diff
@@ -85,9 +86,9 @@ public HollowPrefixIndex(HollowReadStateEngine readStateEngine, String type, Str
     @SuppressWarnings("WeakerAccess")
     public HollowPrefixIndex(HollowReadStateEngine readStateEngine, String type, String fieldPath,
             int estimatedMaxStringDuplicates) {
+        Objects.requireNonNull(readStateEngine, "Hollow Prefix Key Index creation failed because read state wasn't initialized");
```
akhaku: Did you intentionally flip the order here? (as compared to the other cases, not as compared to the previous code here) Might be good to keep all three consistent

akhaku (approved):

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
Sunjeet: ACK, and added type to the message like the others

hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
```diff
@@ -52,6 +53,10 @@
      * @param matchFields The query will match on the specified match fields.  The match fields may span collection elements and/or map keys or values.
      */
     public HollowHashIndex(HollowReadStateEngine stateEngine, String type, String selectField, String... matchFields) {
+        Objects.requireNonNull(type, "Hollow Hash Index creation failed because type was null");
```
ghost: I would static import `requireNonNull`.

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
```diff
@@ -85,9 +86,10 @@ public HollowPrefixIndex(HollowReadStateEngine readStateEngine, String type, Str
     @SuppressWarnings("WeakerAccess")
     public HollowPrefixIndex(HollowReadStateEngine readStateEngine, String type, String fieldPath,
             int estimatedMaxStringDuplicates) {
+        Objects.requireNonNull(type, "Hollow Prefix Key Index creation failed because type was null");
+        Objects.requireNonNull(readStateEngine, "Hollow Prefix Key Index creation for type [" + type
```
ghost: I'm glad that this changes from `IllegalArgumentException` to `NullPointerException`. Technically this could be a breaking change, but I think the risk is fine. For it to break a caller they'd have to have been specifically catching `IllegalArgumentException` to guard against this case. Probably worth calling out in the release notes and at least bumping the minor version instead of just the patch. Or maybe we do a major version bump and be good semver citizens.

I think the risk is low so don't feel strongly.

ghost (approved):
👍

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrefixIndex.java
Sunjeet: Thanks for pointing that out. I had thought of it while making the code change but forgot when creating the PR and if you hadn't pointed it out I would've probably missed it. I will add it to the release notes and bump the minor version given low risk.

Issue #389, opened 2019-01-30
Hash indexes do not support null values


DanielThomas:
Given the index definition where the field `metadata` is nullable:

```
    private val modulesByOrigin = HashIndex.from(consumer, ModuleDescriptor::class.java)
        .usingPath("metadata.origin.value", String::class.java)
```

Index creation will fail with:
```
java.lang.ArrayIndexOutOfBoundsException: -1

    at com.netflix.hollow.core.memory.encoding.FixedLengthElementArray.getElementValue(FixedLengthElementArray.java:133)
    at com.netflix.hollow.core.memory.encoding.FixedLengthElementArray.getElementValue(FixedLengthElementArray.java:111)
    at com.netflix.hollow.core.read.engine.object.HollowObjectTypeReadStateShard.readFixedLengthFieldValue(HollowObjectTypeReadStateShard.java:166)
    at com.netflix.hollow.core.read.engine.object.HollowObjectTypeReadStateShard.readOrdinal(HollowObjectTypeReadStateShard.java:81)
    at com.netflix.hollow.core.read.engine.object.HollowObjectTypeReadState.readOrdinal(HollowObjectTypeReadState.java:145)
    at com.netflix.hollow.core.index.HollowHashIndexBuilder.getMatchHash(HollowHashIndexBuilder.java:395)
    at com.netflix.hollow.core.index.HollowHashIndexBuilder.buildIndex(HollowHashIndexBuilder.java:121)
    at com.netflix.hollow.core.index.HollowHashIndex.reindexHashIndex(HollowHashIndex.java:71)
    at com.netflix.hollow.core.index.HollowHashIndex.<init>(HollowHashIndex.java:62)
    at com.netflix.hollow.api.consumer.index.HashIndexSelect.<init>(HashIndexSelect.java:84)
    at com.netflix.hollow.api.consumer.index.HashIndexSelect.<init>(HashIndexSelect.java:107)
    at com.netflix.hollow.api.consumer.index.HashIndex.<init>(HashIndex.java:55)
    at com.netflix.hollow.api.consumer.index.HashIndex$Builder.usingPath(HashIndex.java:121)
    at com.netflix.astrid.source.DefaultSourceRepository.<init>(DefaultSourceRepository.kt:14)
```
In this case nulls could be skipped over. In the case where the value ending the path is null, I wonder whether null should be supported explicitly to allow query by null?

PR #463, opened 2020-06-01
Fixed field traversal in hash index when parent field path is null.

mikemuske:
This change addresses an issue we are encountering where a hash index is defined with a field that is nested several levels deep. Sometimes the root item has a null value for one of the nested types, which breaks the field traversal code.

My fix was to simply check if the last ordinal retrieved was -1 indicating the value is null. In this case, we don't continue to attempt to traverse the field path by looking up any child ordinals.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
@@ -163,6 +163,10 @@ public class HollowHashIndex implements HollowTypeStateListener {
                     HollowObjectTypeReadState objectAccess = (HollowObjectTypeReadState)readState;
                     readState = objectAccess.getSchema().getReferencedTypeState(fieldPath[j]);
                     hashOrdinal = objectAccess.readOrdinal(hashOrdinal, fieldPath[j]);
+                    // Cannot find nested ordinal for null parent
+                    if(hashOrdinal == -1) {
+                        break;
+                    }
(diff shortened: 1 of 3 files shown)


hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
```diff
@@ -163,6 +163,10 @@ private boolean matchIsEqual(FixedLengthElementArray matchHashTable, long hashBu
                     HollowObjectTypeReadState objectAccess = (HollowObjectTypeReadState)readState;
                     readState = objectAccess.getSchema().getReferencedTypeState(fieldPath[j]);
                     hashOrdinal = objectAccess.readOrdinal(hashOrdinal, fieldPath[j]);
+                    // Cannot find nested ordinal for null parent
+                    if(hashOrdinal == -1) {
```
akhaku: Minor note: can we update these `-1`s to be `HollowConstants.ORDINAL_NONE`?

akhaku (changes requested):
Thank you very much! Looks like this fixes #389

hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
mikemuske: Sure, done.

akhaku (approved):
Thank you very much for the fix!

PR #465, opened 2020-06-04
Shared-memory mode for Consumer

Sunjeet:
### Description
Shared memory mode for Hollow Consumer `mmap`s a snapshot blob to virtual memory and lazily loads only 4k-sized pages of accessed portions of data to physical memory (off JVM heap). This is useful for local debugging using large Hollow datasets, that wouldn't otherwise be possible if the entire dataset were to be eagerly loaded on heap. In future, shared memory mode can also prove useful in deployed environments for faster app init, RAM savings, and to support virtual-memory-scale dataset size (10 TBs) instead of RAM-scale(100 GB). For now,
* only initial load using a snapshot is supported i.e., no subsequent version transitions using deltas or snapshots (note that double-snapshot transitions can be supported with a small code change) 
* type filtering is not supported, and
* indices live on-heap.

**Notable changes:** 
A lot of it is wiring and refactor; significant additions are in the following files:
- Storage and retrieval from `ByteBuffer`s: 

> [BlobByteBuffer.java](https://github.com/Netflix/hollow/blob/shared-memory-converge/hollow/src/main/java/com/netflix/hollow/core/memory/encoding/BlobByteBuffer.java)
> [EncodedLongBuffer.java](https://github.com/Netflix/hollow/blob/shared-memory-converge/hollow/src/main/java/com/netflix/hollow/core/memory/encoding/EncodedLongBuffer.java)
> [EncodedByteBuffer.java](https://github.com/Netflix/hollow/blob/shared-memory-converge/hollow/src/main/java/com/netflix/hollow/core/memory/EncodedByteBuffer.java)

- Unified access across on-heap and shared-memory modes: 
> [HollowBlobInput.java](https://github.com/Netflix/hollow/blob/shared-memory-converge/hollow/src/main/java/com/netflix/hollow/core/read/HollowBlobInput.java)
> [MemoryMode.java](https://github.com/Netflix/hollow/blob/shared-memory-converge/hollow/src/main/java/com/netflix/hollow/core/memory/MemoryMode.java)
> [FixedLengthData.java](https://github.com/Netflix/hollow/blob/shared-memory-converge/hollow/src/main/java/com/netflix/hollow/core/memory/FixedLengthData.java)
> [FixedLengthDataFactory.java](https://github.com/Netflix/hollow/blob/shared-memory-converge/hollow/src/main/java/com/netflix/hollow/core/memory/FixedLengthDataFactory.java)
> [VariableLengthData.java](https://github.com/Netflix/hollow/blob/shared-memory-converge/hollow/src/main/java/com/netflix/hollow/core/memory/VariableLengthData.java)
> [VariableLengthDataFactory.java](https://github.com/Netflix/hollow/blob/shared-memory-converge/hollow/src/main/java/com/netflix/hollow/core/memory/VariableLengthDataFactory.java)

- Testing parity between on-heap array vs. off heap buffer reads:
> [OnHeapArrayVsOffHeapBufferAcceptanceTest.java](https://github.com/Netflix/hollow/blob/shared-memory-converge/hollow/src/test/java/com/netflix/hollow/core/memory/encoding/OnHeapArrayVsOffHeapBufferAcceptanceTest.java)

### Usage
**Scenario A:** Specify memory mode on a `HollowConsumer`-
```
HollowConsumer
      .withBlobRetriever(blobRetriever)   /// required: a BlobRetriever
      .withMemoryMode(memoryMode)         /// optional: memory mode defaults to ON_HEAP
   .build();
```
**Scenario B:** Initialize a ReadStateEngine from a snapshot file. Traditionally, one would write-
```
HollowReadStateEngine readState = new HollowReadStateEngine();
HollowBlobReader reader = new HollowBlobReader(readState);
reader.readSnapshot(new BufferedInputStream(new FileInputStream(file)));
```
For shared memory mode that code would look like-
```
HollowReadStateEngine readState = new HollowReadStateEngine();
HollowBlobReader reader = new HollowBlobReader(readState, MemoryMode.SHARED_MEMORY_LAZY);
reader.readSnapshot(HollowBlobInput.randomAccess(new File(file)));
```

### Backwards compatibility
The producer side is unchanged, so shared memory consumer is backwards compatible with previously produced blobs. As for code changes, breaking API changes are limited to classes that aren’t expected to be consumed externally and have no searchable usages. This will be a major version release. 

### Performance
A performance hit is expected on the read path due to page faults. App initialization is typically faster (depending on use of indices) because data is mmap-ed to virtual memory instead of being copied into on-heap arrays. More work is required for performance profiling and operational viability. Initially only the local dev use case is targeted.


hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowHistoryKeyIndex.java
```diff
@@ -182,11 +182,12 @@ private HollowReadStateEngine roundTripStateEngine(boolean isInitialUpdate, bool
                 }
             });
 
-            BufferedInputStream bin = new BufferedInputStream(in);
```
Sunjeet, line 185: There was a memory leak in the old code (`bin` wasn't `close`d).

PR #469, opened 2020-06-26
Document HollowConsumer.Builder#withLocalBlobStore

akhaku:


hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
```diff
@@ -969,11 +969,36 @@ public B withBlobRetriever(HollowConsumer.BlobRetriever blobRetriever) {
...
+         * to retrieving using the configured {@link HollowConsumer.BlobRetriever}.
+         * If it does not find the blob in the configured local blob store directory, it will
+         * retrieve the blob using the configured blobRetriever and then use it from there.
+         * Note that cached files are never deleted from {@code localBlobStoreDir}, meaning if you
+         * use this on a long-running instance you will eventually run out of disk space. As such, a
+         * local blob store is better suited for local development, testing, or tooling purposes.
+         *
+         * @param localBlobStoreDir the directory used to store cached blobs. This will be created
+         *   if it does not already exist.
+         * @param noFallBackForExistingSnapshot
```
Sunjeet: Thanks for the adding the description- definitely helps. Should we also rename the param to `useWhatsAvailable`?

akhaku: Yes! I'm going to rename it, trying to reason about `noFallbackForExistingSnapshot` is making my head explode

PR #470, opened 2020-06-29
Add per-dataset restore flow with support for referenced types

cosmin-ionita:
The context for this PR is described here: https://github.com/Netflix/hollow/issues/471

Any guidance on how to add a unit test (inline with the existing project setup) for that would be highly appreciated.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/write/HollowWriteStateEngine.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/write/HollowWriteStateEngine.java
@@ -29,8 +29,14 @@ import com.netflix.hollow.core.util.SimultaneousExecutor;
 import com.netflix.hollow.core.write.objectmapper.HollowObjectMapper;
 import com.netflix.hollow.core.write.objectmapper.HollowTypeMapper;
 import com.netflix.hollow.tools.traverse.TransitiveSetTraverser;
-
-import java.util.*;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Random;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.BitSet;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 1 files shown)


AlexandruGhergut (approved):
That's what I thought `addAllObjectsFromPreviousCycle` would actually do. This is useful for providing more flexibility in deciding how the cycle ends during failures when producing one of your datasets

Sunjeet:
(duplicated at the issue #471 ) Why not have a producer per dataset and use the producer's restore functionality? Or I guess that could be another option.

cosmin-ionita:
@Sunjeet Thank you for the question. I replied on the issue.

hollow/src/main/java/com/netflix/hollow/core/write/HollowWriteStateEngine.java
```diff
@@ -245,6 +247,28 @@ public void addAllObjectsFromPreviousCycle() {
...
     }
 
+    /**
+     * Add all objects of the provided type from the previous cycle, exactly as they were in the previous cycle, while
+     * also considering referenced records.
+     * @param typeName The type name of the records that should added from the previous cycle
+     * @param readStateEngine The read state of the previous cycle (should be retrieved using the method linked below)
+     * {@link com.netflix.hollow.api.producer.HollowProducer.WriteState#getPriorState()}
+     */
+    public void addAllObjectsFromPreviousCycle(String typeName, HollowReadStateEngine readStateEngine) {
```
ghost, line 257: As you point out, this is what `TransitiveSetTraverser` is for. There's a few reasons I'm against adding this helper method:

* your use use-cases is one of several possible use cases. I'm not confident it's the primary use-case, for example  the incremental producer does something similar with `TransitiveSetTraverser`

* I want to avoid bloating `HollowWriteStateEngine` with helper methods, especially per above if there are a number of slight variations for which a helper method would be justified

* The contract and the API are inconsistent. Nothing prevents users from using an arbitrary HollowReadStateEngine as the second arguments. More likely to cause confusion than be helpful.

ghost (commented):
@cosmin-ionita, I'm going to reject your PR, but I want to thank you for the contribution.

I think you highlight a use-case that we can do a better job guiding people to solutions through docs, but that doesn't warrant this helper method (details in code comment).

hollow/src/main/java/com/netflix/hollow/core/write/HollowWriteStateEngine.java
cosmin-ionita: @toolbear I totally agree with your reasons, but don't you think it's useful to have the capability of independently publishing the datasets somewhere in the producer? I think that restoring a dataset with reference-type awareness should be an internal detail of hollow, and not something the the user should be aware of.

PR #474, opened 2020-07-14
Expose getPrimaryKeys() API at HollowStateEngine

davesu:
There are many use cases of HollowStateEngine's Primary Keys so instead of having many projects copy/paste code, it would be great to have it part of HollowStateEngine itself.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/HollowStateEngine.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/HollowStateEngine.java
@@ -65,6 +65,16 @@ public interface HollowStateEngine extends HollowDataset {
 
     String getHeaderTag(String name);
 
+    default PrimaryKey getPrimaryKey(String typeName) {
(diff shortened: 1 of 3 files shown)


hollow/src/main/java/com/netflix/hollow/core/HollowStateEngine.java
```diff
@@ -59,4 +64,29 @@
     Map<String, String> getHeaderTags();
 
     String getHeaderTag(String name);
+
+    default PrimaryKey getPrimaryKey(String typeName) {
+        PrimaryKey pk=null;
+        HollowSchema schema = getSchema(typeName);
+        if (schema.getSchemaType() == HollowSchema.SchemaType.OBJECT) {
```
Sunjeet: nit: null check here?

davesu: Good idea.

PR #480, opened 2020-07-27
Performance enhancements for high cardinality types

zurdoron:
during the population stage in the cycle the ordinal map of each type is populated sequentially, in high cardinality types this can relatively a long time  (for us a type with 20M values took around 25s), this enhancement allows the population of the internal hashmap of the ByteArrayOrdinalMap to scale to multiple cores and thus dramatically reduce the time of the population stage


brachi-wernick:
We also faced this slowness when using high cardinality types.
I would want to use this improvement, what is the status of this PR?

PR #482, opened 2020-08-12
Deal with race condition between snapshot application and HollowConsumer#getAPI

akhaku:
If a user calls HollowConsumer#getAPI in the middle of a
snapshot update, they will get back null if the hollowDataHolder
volatile field is assigned but the snapshot is not yet applied.
To work around this, only set the volatile field after the
new API is initialized (and so the snapshot is processed)
but before refresh handlers are called.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
@@ -151,12 +151,19 @@ public class HollowClientUpdater {
             if (updatePlan.isSnapshotPlan()) {
                 HollowDataHolder oldDh = hollowDataHolderVolatile;
                 if (oldDh == null || doubleSnapshotConfig.allowDoubleSnapshot()) {
-                    // The volatile field needs to be assigned here because it may be accessed when
-                    // executing the update plan, for example a refresh listener may access the consumer
-                    // after the plan has been applied (such as a unique key indexer).
-                    HollowDataHolder newDh = hollowDataHolderVolatile = newHollowDataHolder();
+                    HollowDataHolder newDh = newHollowDataHolder();
                     try {
-                        newDh.update(updatePlan, localListeners);
+                        /* We need to assign the volatile field after API init since it may be
+                         * accessed during the update plan application, for example via a refresh
+                         * listener (such as a unique key indexer) that calls getAPI. If we do it after
+                         * newDh.update(), refresh listeners will see the old API. If we do it
+                         * before then we open ourselves up to a race where a caller will get back
+                         * null if they call getAPI after assigning the volatile but before the API
+                         * is initialized in HollowDataHolder#initializeAPI.
+                         * Also note that hollowDataHolderVolatile only changes for snapshot plans,
+                         * and it is only for snapshot plans that HollowDataHolder#initializeAPI is
+                         * called. */
+                        newDh.update(updatePlan, localListeners, () -> hollowDataHolderVolatile = newDh);
                     } catch (Throwable t) {
                         // If the update plan failed then revert back to the old holder
                         hollowDataHolderVolatile = oldDh;
@@ -165,7 +172,7 @@ public class HollowClientUpdater {
                     forceDoubleSnapshot = false;
                 }
             } else {
-                hollowDataHolderVolatile.update(updatePlan, localListeners);
+                hollowDataHolderVolatile.update(updatePlan, localListeners, () -> {});
             }
 
             for(HollowConsumer.RefreshListener refreshListener : localListeners)
--- a/hollow/src/main/java/com/netflix/hollow/api/client/HollowDataHolder.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/client/HollowDataHolder.java
@@ -99,7 +99,8 @@ class HollowDataHolder {
         return this;
     }
 
-    void update(HollowUpdatePlan updatePlan, HollowConsumer.RefreshListener[] refreshListeners) throws Throwable {
+    void update(HollowUpdatePlan updatePlan, HollowConsumer.RefreshListener[] refreshListeners,
+            Runnable apiInitCallback) throws Throwable {
         // Only fail if double snapshot is configured.
         // This is a short term solution until it is decided to either remove this feature
         // or refine it.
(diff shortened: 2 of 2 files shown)


hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
```diff
@@ -151,12 +151,19 @@ public synchronized boolean updateTo(long requestedVersion) throws Throwable {
...
-                    HollowDataHolder newDh = hollowDataHolderVolatile = newHollowDataHolder();
+                    HollowDataHolder newDh = newHollowDataHolder();
                     try {
-                        newDh.update(updatePlan, localListeners);
+                        /* We need to assign the volatile field after API init since it may be
+                         * accessed during the update plan application, for example via a refresh
+                         * listener (such as a unique key indexer) that calls getAPI. If we do it after
+                         * newDh.update(), refresh listeners will see the old API. If we do it
+                         * before then we open ourselves up to a race where a caller will get back
+                         * null if they call getAPI after assigning the volatile but before the API
```
Sunjeet, line 161: Would this be another viable option? Make oldDh a class variable and modify getAPI() below to-
```
return hollowDataHolderLocal == null ? null :
                hollowDataHolderLocal.getAPI() == null ? oldDh :
                    hollowDataHolderLocal.getAPI();
```
I think that fixes the NPE. There may still be a race in that getAPI() returns oldDh but refresh listeners are called on the new state- I suppose thats ok and happens anyway but I'm not sure. 

The fix as-is seems reasonable to me too. Perhaps one suggestion would be to replace use of Runnable with Consumer<Void> and call callback.accept(null)

ghost: > Make oldDh a class variable

IIUC we would then have to lock on `oldDh` for the case of multiple consumer instances and now multiple consumers have contention.

ghost: > Runnable with Consumer and call callback.accept(null)

I like this suggestion.

Sunjeet: > lock on oldDh for the case of multiple consumer instances

I'm not seeing it, can you elaborate? oldDh would be a class member of HollowClientUpdater and the only assignment to it would be in HollowClientUpdater::updateTo()?

akhaku: RE: making `oldDh` a volatile class variable - I think that includes a bit of complexity where we need to worry about old vs new data holder, where the only time it changes is for snapshot updates. getAPI would also have to be more complex, essentially what you said plus we'll have to atomically copy both, the (volatile) oldDh _and_ the (volatile) hollowDataHolder.
I also don't understand why a `Consumer<what type?>` with null is cleaner - since we don't need to pass anything in, is it actually cleaner to pass in a generic lamba (so a Runnable)?

akhaku: For some additional context: here's a (very old) commit where I tried to fix similar races (but clearly missed this one :( ) https://github.com/Netflix/hollow/commit/0e1d898537a2e5302de302dd4e9e4d368161da70

ghost: I'm thinking of two consumer instances each performing an update in separate threads.

Each instance has its own refresh lock, so they could both be updating concurrently. `HollowClientUpdater::updateTo` is `synchronized` but each consumer instance has its own `HollowClientUpdater` instance and the synchronized guarantees only apply to calls on the same instance, not across instances. If `oldDh` was a static class variable this would introduce a race condition in `updateTo` unless we introduced some class-level locking/synchronization to mitigate.

Otherwise, given consumers A and B updating concurrently it could be possible for `A`'s `hollowDataHolderVolatile` to be assigned `B`'s data holder, or vice-versa.

Sunjeet: @toolbear would oldDh need to be a static variable?

@akhaku Not a strong reservation, but Consumer<Void> instead of Runnable just feels more java8. I think it also carries a trivial advantage in that we don't create a thread to run the callback on.

Sunjeet: It'd be a Consumer < Void > , if you choose to incorporate it

akhaku: Ah `Runnable` isn't creating a new thread, it's effectively the same as a `Consumer` without an argument. You may be confusing it with passing a `Runnable` to `new Thread` and then calling `.start().join()`. I think a `Consumer` would make more sense if we were actually consuming something; here we really want a `Consumer` with no argument, which is a `Runnable`. E.g see the second answer here https://stackoverflow.com/questions/29945627/java-8-lambda-void-argument

Regarding `oldDh` yes it doesn't need to be class-level, but we still need to worry about atomically getting a handle to the "current" oldDh and hollowDataHolder, which gets a little tricky without synchronization.

Sunjeet: @toolbear nvm, I indicated static variable when I used class variable / class member to describe instance variable.

Sunjeet: @akhaku Thanks for the distinction, yes, I was confusing it because Runnable is typically used when running in a new Thread.

hollow/src/main/java/com/netflix/hollow/api/client/HollowDataHolder.java
```diff
@@ -111,18 +112,21 @@ void update(HollowUpdatePlan updatePlan, HollowConsumer.RefreshListener[] refres
...
-        else
+        if (updatePlan.isSnapshotPlan()) {
+            applySnapshotPlan(updatePlan, refreshListeners, apiInitCallback);
+        } else {
             applyDeltaOnlyPlan(updatePlan, refreshListeners);
+        }
     }
 
-    private void applySnapshotPlan(HollowUpdatePlan updatePlan, HollowConsumer.RefreshListener[] refreshListeners) throws Throwable {
-        applySnapshotTransition(updatePlan.getSnapshotTransition(), refreshListeners);
```
dkoszewnik, line 125: A few lines below this, `refreshListener.snapshotUpdateOccurred()` is called, might this be a problem if the `HollowDataHolder` is not assigned in the updater yet?

hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
```diff
@@ -151,12 +151,19 @@ public synchronized boolean updateTo(long requestedVersion) throws Throwable {
             if (updatePlan.isSnapshotPlan()) {
                 HollowDataHolder oldDh = hollowDataHolderVolatile;
                 if (oldDh == null || doubleSnapshotConfig.allowDoubleSnapshot()) {
-                    // The volatile field needs to be assigned here because it may be accessed when
-                    // executing the update plan, for example a refresh listener may access the consumer
-                    // after the plan has been applied (such as a unique key indexer).
```
dkoszewnik, line 156: What about this comment that's getting deleted -- it seems like a valid point?
Under the `HollowDataHolder.update()` method, calls to the refreshListeners are made, right?

akhaku: Yes, this is expected - the comment below tries to capture that it maintains the same invariant. The short version is that we want to be sure that the data holder has been swapped before we call any listeners. That is still the case in this PR.

hollow/src/main/java/com/netflix/hollow/api/client/HollowDataHolder.java
akhaku: Yes indeed, that would be a problem. This PR doesn't change that though (more detail in the response to your other comment)

PR #483, opened 2020-08-17
Hollow schema precondition that name should not be null or empty

Sunjeet:


hollow/src/main/java/com/netflix/hollow/core/schema/HollowSchema.java
```diff
@@ -49,6 +49,9 @@
     private final String name;
 
     public HollowSchema(String name) {
+        if (name == null || name == "") {
```
akhaku: I suppose the test passes because `""` happens to be interned, but might be safer to use `.equals` anyway?

Sunjeet: Derp. Thanks!

akhaku (approved):

akhaku:
Hm you know this _may_ be a breaking change for people who don't realize their schema contains null/empty types. At least internally we can figure out who is in that set of people though.

EDIT we talked via IM, sounds like it isn't possible for anyone to be in that set and not already be broken.

PR #493, opened 2020-11-16
Bugfix: HollowRecordJsonStringifier append newline only when prettyprint

Sunjeet:


akhaku (approved):
LGTM, a test would be 💯

PR #494, opened 2020-11-16
Feature: search in history by composite primary key

Sunjeet:
Status quo: History UI's search can accept a single field- so one can search based on any indexed field, and by default all fields comprising a primary key are indexed.

New: If a primary key is composed of multiple fields, one can search using primary key by colon delimiting the values, for eg. "1234:US"

### DIFF
--- a/hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/pages/BrowseSelectedTypePage.java
+++ b/hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/pages/BrowseSelectedTypePage.java
@@ -16,15 +16,15 @@
  */
 package com.netflix.hollow.explorer.ui.pages;
 
-import com.netflix.hollow.core.index.HollowPrimaryKeyIndex;
+import static com.netflix.hollow.tools.util.SearchUtils.getFieldPathIndexes;
+import static com.netflix.hollow.tools.util.SearchUtils.getOrdinalToDisplay;
+import static com.netflix.hollow.tools.util.SearchUtils.getPrimaryKey;
+import static com.netflix.hollow.tools.util.SearchUtils.parseKey;
+
 import com.netflix.hollow.core.index.key.PrimaryKey;
 import com.netflix.hollow.core.read.HollowReadFieldUtils;
 import com.netflix.hollow.core.read.engine.HollowTypeReadState;
-import com.netflix.hollow.core.read.engine.HollowTypeStateListener;
 import com.netflix.hollow.core.read.engine.object.HollowObjectTypeReadState;
-import com.netflix.hollow.core.schema.HollowObjectSchema;
-import com.netflix.hollow.core.schema.HollowSchema;
-import com.netflix.hollow.core.schema.HollowSchema.SchemaType;
 import com.netflix.hollow.explorer.ui.HollowExplorerUI;
 import com.netflix.hollow.explorer.ui.model.QueryResult;
 import com.netflix.hollow.explorer.ui.model.TypeKey;
@@ -81,7 +81,7 @@ public class BrowseSelectedTypePage extends HollowExplorerPage {
         }
         
         PrimaryKey primaryKey = getPrimaryKey(typeState.getSchema());
-        int fieldPathIndexes[][] = getFieldPathIndexes(primaryKey); 
+        int fieldPathIndexes[][] = getFieldPathIndexes(ui.getStateEngine(), primaryKey);
         
         List<TypeKey> keys = new ArrayList<>(pageSize);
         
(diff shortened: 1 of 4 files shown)


hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/pages/BrowseSelectedTypePage.java
```diff
@@ -94,14 +94,15 @@ protected void setUpContext(HttpServletRequest req, HollowUISession session, Vel
...
+            parsedKey = parseKey(ui.getStateEngine(), primaryKey, key);
         } catch(Exception e) {
             key = "";
         }
 
-        Integer ordinal =
-            getOrdinalToDisplay(req, key, parsedKey, selectedOrdinals, fieldPathIndexes);
-        if (ordinal != null && !ordinal.equals(-1) && "".equals(key)
+        HollowTypeReadState readTypeState = getTypeState(req);
+        int ordinal = req.getParameter("ordinal") == null ? -1 : Integer.parseInt(req.getParameter("ordinal"));
```
davesu: Minor: it would be great if there is a constant for all those  -1.

hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowHistoryTypeKeyIndex.java
```diff
@@ -213,6 +217,28 @@ public IntList queryIndexedFields(final String query) {
             return matchingKeys;
         }
 
+        String[] parts;
+        if (query.contains(":")) {  // composite field query, uses ':' as separator
```
davesu: minor: it would be great to have constant for ":" as well

hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowHistoryTypeKeyIndex.java
```diff
@@ -213,6 +217,28 @@ public IntList queryIndexedFields(final String query) {
...
+            parts = query.split(":");
+            if (parts.length != primaryKey.numFields()) {
+                return matchingKeys;
+            }
+
+            Object[] parsedKey;
+            try {
+                parsedKey = parseKey(readStateEngine, primaryKey, query);
+            } catch(Exception e) {
+                return matchingKeys;
```
davesu, line 238: Wonder if some logging here would be useful for debugging purpose.

hollow/src/main/java/com/netflix/hollow/tools/util/SearchUtils.java
```diff
@@ -0,0 +1,149 @@
...
+                case LONG:
+                    key[i] = Long.parseLong(fields[i]);
+                    break;
+                case DOUBLE:
+                    key[i] = Double.parseDouble(fields[i]);
+                    break;
+                case FLOAT:
+                    key[i] = Float.parseFloat(fields[i]);
+                    break;
+                case BYTES:
```
davesu, line 54: Does PrimaryKey support BYTES field type?

hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowHistoryTypeKeyIndex.java
Sunjeet: done

hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/pages/BrowseSelectedTypePage.java
Sunjeet: done

hollow/src/main/java/com/netflix/hollow/tools/util/SearchUtils.java
```diff
@@ -0,0 +1,149 @@
...
+    /**
+     * Returns primary key index for a given type if it exists.
+     */
+    public static HollowPrimaryKeyIndex findPrimaryKeyIndex(HollowTypeReadState typeState) {
+        if(getPrimaryKey(typeState.getSchema()) == null)
+            return null;
+
+        for(HollowTypeStateListener listener : typeState.getListeners()) {
+            if(listener instanceof HollowPrimaryKeyIndex) {
+                if(((HollowPrimaryKeyIndex) listener).getPrimaryKey().equals(getPrimaryKey(typeState.getSchema())))
```
davesu: Minor: Wonder if it helps to extract "getPrimaryKey(typeState.getSchema())" outside the loop.

hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowHistoryTypeKeyIndex.java
Sunjeet: Good call

hollow/src/main/java/com/netflix/hollow/tools/util/SearchUtils.java
Sunjeet: Primary key does support BYTES, but search in explorer/history doesn't. I changed it slightly to now throw an exception if an attempt is made to search using primary key when primary key contains a BYTES field. This exception will be logged.

PR #496, opened 2020-12-02
Optimization to delay creation of Hollow Objects only when needed.

davesu:
- Enhance computeDataChange to capture ordinals only and not create Hollow Objects
- Benefits will be significant for large state changes

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/consumer/data/AbstractHollowDataAccessor.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/consumer/data/AbstractHollowDataAccessor.java
@@ -101,15 +101,18 @@ public abstract class AbstractHollowDataAccessor<T> {
     protected void computeDataChange(String type, HollowReadStateEngine stateEngine, PrimaryKey primaryKey) {
         HollowTypeReadState typeState = stateEngine.getTypeDataAccess(type).getTypeState();
 
+        BitSet previousOrdinals = typeState.getPreviousOrdinals();
+        BitSet currentOrdinals = typeState.getPopulatedOrdinals();
+
         // track removed ordinals
         removedOrdinals = new BitSet();
-        removedOrdinals.or(typeState.getPreviousOrdinals());
-        removedOrdinals.andNot(typeState.getPopulatedOrdinals());
+        removedOrdinals.or(previousOrdinals);
+        removedOrdinals.andNot(currentOrdinals);
 
         // track added ordinals
         addedOrdinals = new BitSet();
-        addedOrdinals.or(typeState.getPopulatedOrdinals());
-        addedOrdinals.andNot(typeState.getPreviousOrdinals());
+        addedOrdinals.or(currentOrdinals);
+        addedOrdinals.andNot(previousOrdinals);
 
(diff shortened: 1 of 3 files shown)


hollow/src/main/java/com/netflix/hollow/api/consumer/data/AbstractHollowDataAccessor.java
```diff
@@ -99,59 +102,33 @@ protected void computeDataChange(String type, HollowReadStateEngine stateEngine,
...
 
                 addedOrdinal = addedOrdinals.nextSetBit(addedOrdinal + 1);
             }
         }
-
-        { // determine removed records
-            int removedOrdinal = removedOrdinals.nextSetBit(0);
-            while (removedOrdinal != -1) {
-                // exclude records that was removed but re-added with different ordinal since that is considered as updated record
-                if (!updatedRecordOrdinals.get(removedOrdinal)) {
```
Sunjeet, line 143: This logic might be needed to correctly compute removed ordinals, without it updates will also be returned as removed

Sunjeet:
Is this cost currently incurred in delta application? Does moving the deserialization cost to access time move the problem elsewhere? Do getAllRecords() etc. get called repeatedly, are the deserialized records cached?

hollow/src/main/java/com/netflix/hollow/api/consumer/data/AbstractHollowDataAccessor.java
davesu: yeah, I did notice the issue with the unit test but was not able to run it locally due to build/dependency (jcenter repo) issues.  

Now that the above jcenter repo issue was corrected, I was able to run unit test and fix the issues.  Please take another look at the PR.

davesu:
> Is this cost currently incurred in delta application? Does moving the deserialization cost to access time move the problem elsewhere? Do getAllRecords() etc. get called repeatedly, are the deserialized records cached?

Most people don't currently use the generated DataAccessor.  Also one will not notice performance overhead with small dataset or small deltas.  But with large datasets creating all those POJO during change calculate could be really costly both CPU and large heap.   Specially for use cases which only cares about modified records and not added nor removed ones.   

Therefore, this change falls inline with hollow practice on only hydrating POJOs when needed.

hollow/src/main/java/com/netflix/hollow/api/consumer/data/AbstractHollowDataAccessor.java
```diff
@@ -27,19 +27,22 @@
...
+import java.util.Objects;
 
 public abstract class AbstractHollowDataAccessor<T> {
     protected final String type;
     protected final PrimaryKey primaryKey;
     protected final HollowReadStateEngine rStateEngine;
 
-    private List<T> removedRecords = Collections.emptyList();
-    private List<T> addedRecords = Collections.emptyList();
+    private BitSet removedOrdinals = new BitSet();
```
Sunjeet, line 44: nit: initialized twice with `= new BitSet()`, we can drop it from one place

Sunjeet (approved):

PR #497, opened 2020-12-12
SimultaneousExecutor support for Thread Priority

davesu:


hollow/src/main/java/com/netflix/hollow/core/util/SimultaneousExecutor.java
```diff
@@ -144,8 +145,38 @@ public SimultaneousExecutor(int numThreads) {
      * @param description brief description used to name created threads; combined with {@code context}
      */
     public SimultaneousExecutor(int numThreads, Class<?> context, String description) {
-        super(numThreads, numThreads, 100, TimeUnit.SECONDS, new LinkedBlockingQueue<>(),
-                r -> daemonThread(r, context, description));
+        this(numThreads, context, description, Thread.NORM_PRIORITY);
```
nayanika-u, line 148: I think it will be worthwhile to add a unit test case for this code change.

davesu: I tried to add unit test but SimultaneousExecutor does not expose the Thread instances to verify the priority.

davesu: I should mention that I did run existing SimultaneousExecutorTest and make sure it passes existing tests.

nayanika-u: Okay. I checked. The existing UTs should suffice then. They are covering the newly added constructors.

PR #499, opened 2021-01-07
Augment RecordCountVarianceValidator to support optional behavior to only fail the publish if the record count goes down by the specified percentage

Helmsdown:
This will allow clients to allow as many additions for a given type as needed by the business, but will prevent sudden loss of data beyond the specified threshold.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/validation/RecordCountVarianceValidator.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/validation/RecordCountVarianceValidator.java
@@ -45,6 +45,7 @@ public class RecordCountVarianceValidator implements ValidatorListener {
     private static final String NAME = RecordCountVarianceValidator.class.getName();
 
     private final String typeName;
+    private final boolean onlyValidateDropsInCardinality;
 
     private final float allowableVariancePercent;
 
@@ -58,7 +59,24 @@ public class RecordCountVarianceValidator implements ValidatorListener {
      * Anything more results in failure of validation.
      */
     public RecordCountVarianceValidator(String typeName, float allowableVariancePercent) {
+        this(typeName, allowableVariancePercent, false);
+    }
+
+    /**
+     * @param typeName type name
+     * @param allowableVariancePercent: Used to validate if the cardinality change in current cycle is with in the
+     * @param onlyValidateDropsInCardinality: If true, only drops in cardinality will cause a failure of validation
+     * allowed percent.
+     * Ex: 0% allowableVariancePercent ensures type cardinality does not vary at all for cycle to cycle.
+     * Ex: Number of state in United States.
+     * 10% allowableVariancePercent: from previous cycle any addition or removal within 10% cardinality is valid.
+     * Anything more results in failure of validation.
+     */
+    public RecordCountVarianceValidator(String typeName,
+                                        float allowableVariancePercent,
+                                        boolean onlyValidateDropsInCardinality) {
         this.typeName = typeName;
+        this.onlyValidateDropsInCardinality = onlyValidateDropsInCardinality;
         if (allowableVariancePercent < 0) {
             throw new IllegalArgumentException("RecordCountVarianceValidator for type " + typeName
                     + ": cannot have allowableVariancePercent less than 0. Value provided: "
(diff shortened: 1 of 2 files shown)


hollow/src/main/java/com/netflix/hollow/api/producer/validation/RecordCountVarianceValidator.java
```diff
@@ -93,9 +111,12 @@ public ValidationResult onValidate(ReadState readState) {
         vrb.detail(ACTUAL_CHANGE_PERCENT_NAME, actualChangePercent);
 
         if (Float.compare(actualChangePercent, allowableVariancePercent) > 0) {
-            String message = String.format(FAILED_RECORD_COUNT_VALIDATION, typeName, actualChangePercent,
-                    allowableVariancePercent);
-            return vrb.failed(message);
+            if(!onlyValidateDropsInCardinality || latestCardinality < previousCardinality) {
```
nayanika-u: Just a minor one : When would this flag value be used as false? could you add a test case where onlyValidateDropsInCardinality is false. The test cases you've added test the changes in cardinality but not this flag as such.

hollow/src/test/java/com/netflix/hollow/api/producer/validation/RecordCountVarianceValidatorTests.java
```diff
@@ -99,6 +99,36 @@ public void populate(WriteState newState) throws Exception {
...
+
+            producer.runCycle(new Populator() {
+                public void populate(WriteState newState) throws Exception {
+                    newState.add(new TypeWithPrimaryKey(1, "Brad Pitt", "klsdjfla;sdjkf"));
+                    newState.add(new TypeWithPrimaryKey(1, "Angelina Jolie", "as;dlkfjasd;l"));
+                }
+            });
+            Assert.fail();
+        } catch (ValidationStatusException expected) {
+            //System.out.println("Message:"+expected.getIndividualFailures().get(0).getMessage());
```
nayanika-u: Please remove the commented out statement.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/RecordCountVarianceValidator.java
nayanika-u: Minor : Space after if is missing.

nayanika-u: Just to add : existing test cases should suffice for testing false value. So no need. I just wanted to clarify if there would be a case where you'd explicitly pass this flag value as false (apart from the default case).

hollow/src/test/java/com/netflix/hollow/api/producer/validation/RecordCountVarianceValidatorTests.java
```diff
@@ -127,6 +157,36 @@ public void populate(WriteState newState) throws Exception {
                 .cardinality());
     }
 
+    @Test
+    public void passTestNoMoreChangeThanExpected_onlyValidateDropsInCardinality() {
+        HollowProducer producer = HollowProducer.withPublisher(blobStore).withBlobStager(new HollowInMemoryBlobStager())
+                .withListener(new RecordCountVarianceValidator("TypeWithPrimaryKey", 1f, true)).build();
+
+        // runCycle(producer, 1);
```
nayanika-u: minor : remove commented line.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/RecordCountVarianceValidator.java
Helmsdown: I would never choose to use the three argument constructor AND pass false for the `onlyValidateDropsInCardinality` parameter. I chose to override the 2 argument constructor so we don't break existing integrations when they upgrade.

nayanika-u: Okay. That default part I get. I was wondering if you were going to use it for local testing or some other case which I was missing.

nayanika-u: Changes look good. Merging this.

Sunjeet:
A potential issue with validation based on drop in # of ordinals in the bitset (or in the context of this PR drop in cardinality) is that a mass removal of records might not translate to a drop in cardinality if there was also a mass addition in the same version. For eg. if a version dropped 1k records in error, but also added 1.1k new records as expected. 
To achieve the desired validation we'll need to count % of records removed, instead of % of ordinals removed.

Sunjeet:
As per our chat, the existing validator code also seems subject to above false negative case. As we add the proposed validator, a couple of things that would be good to incorporate are-
* count the # records removed based on ordinals reset in new state, based on value of primary key (similar to what ObjectModificationValidator does)
* implement this as a separate validator because of reliance on presence of primary key
We also decided to iterate on this in Russel's app, and later pull down the feature into Hollow. For now this commit will be reverted.

PR #500, opened 2021-01-07
DataAccessor tweaks

davesu:
- Fix Generator to use schema name for data type instead of implClass since classname does not always match schema type (e.g. when using PostFix for classname)

- Also introduce DataAccessor.hasPriorState() since it is important data point when using DataAccessor.  Loading Snapshot does not contain prior state info so can't computeChange.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/codegen/api/HollowDataAccessorGenerator.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/codegen/api/HollowDataAccessorGenerator.java
@@ -46,7 +46,7 @@ public class HollowDataAccessorGenerator extends HollowConsumerJavaFileGenerator
         super(packageName, SUB_PACKAGE_NAME, dataset, config);
         this.className = getClassName(schema);
         this.apiclassName = apiclassName;
-        this.type =  hollowImplClassname(schema.getName());
+        this.type = schema.getName();
         this.schema = schema;
     }
 
--- a/hollow/src/main/java/com/netflix/hollow/api/codegen/api/HollowDataAccessorGenerator.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/codegen/api/HollowDataAccessorGenerator.java
@@ -46,7 +46,7 @@ public class HollowDataAccessorGenerator extends HollowConsumerJavaFileGenerator
         super(packageName, SUB_PACKAGE_NAME, dataset, config);
         this.className = getClassName(schema);
         this.apiclassName = apiclassName;
-        this.type = schema.getName();
+        this.type =  hollowImplClassname(schema.getName());
         this.schema = schema;
     }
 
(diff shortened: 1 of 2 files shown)


hollow/src/main/java/com/netflix/hollow/api/consumer/data/AbstractHollowDataAccessor.java
```diff
@@ -81,6 +82,17 @@ public AbstractHollowDataAccessor(HollowReadStateEngine rStateEngine, String typ
...
     }
 
+    /**
+     * Indicate whether Data Accessor contains prior state
+     *
+     * NOTE: This is critical since loading a Snapshot will not contain any information about changes from prior state
+     *
+     * @return true indicate it contains prior state
+     */
+    public boolean hasPriorState() {
```
Sunjeet, line 92: Can you plz share any usage of this method? It'll help me better understand why its needed

hollow/src/main/java/com/netflix/hollow/api/codegen/api/HollowDataAccessorGenerator.java
```diff
@@ -68,7 +68,7 @@ public String generate() {
         builder.append("@SuppressWarnings(\"all\")\n");
         builder.append("public class " + className + " extends " + AbstractHollowDataAccessor.class.getSimpleName() + "<" + type  +"> {\n\n");
 
-        builder.append("    public static final String TYPE = \"" + type + "\";\n");
+        builder.append("    public static final String TYPE = \"" + schema.getName() + "\";\n");
```
Sunjeet, line 71: Not very familiar with this code path. Can you plz share example generated DataAccessor BEFORE and AFTER?

hollow/src/main/java/com/netflix/hollow/api/consumer/data/AbstractHollowDataAccessor.java
davesu: On a snapshot, getAddedRecords() will return all Ids so this is the key indicator to enable properly interpret that the added ids were not from delta from prior state.

hollow/src/main/java/com/netflix/hollow/api/codegen/api/HollowDataAccessorGenerator.java
davesu: This is to address issue when HollowAPIGenerator.Builder().withClassPostfix("Hollow").

Usage of type of correct on other places but TYPE here is supposed to be matching schema type name.   We can consider renaming TYPE -> SCHEMA_TYPE if that help.   

Ping me and I can provide actual example on when existing code has issue.

Sunjeet: ACK, thanks

Sunjeet (approved):

PR #506, opened 2021-02-02
Hollow Diff Support Diffing Type that does not have primary key

davesu:
- Single Field Type : Use Single Field as match Path
- Multi Field Type : Just add all data as unmatched since there is no key to match on (diff basic information)

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/tools/diff/HollowDiff.java
+++ b/hollow/src/main/java/com/netflix/hollow/tools/diff/HollowDiff.java
@@ -19,11 +19,13 @@ package com.netflix.hollow.tools.diff;
 import com.netflix.hollow.core.index.key.PrimaryKey;
 import com.netflix.hollow.core.read.engine.HollowReadStateEngine;
 import com.netflix.hollow.core.schema.HollowObjectSchema;
+import com.netflix.hollow.core.schema.HollowObjectSchema.FieldType;
 import com.netflix.hollow.core.schema.HollowSchema;
 import com.netflix.hollow.core.util.SimultaneousExecutor;
 import com.netflix.hollow.tools.diff.exact.DiffEqualityMapping;
 
 import java.util.ArrayList;
+import java.util.EnumSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


davesu:
I just realized that I need to only support single field of basic type only.  Will update PR soon.

hollow/src/main/java/com/netflix/hollow/tools/diff/HollowDiff.java
```diff
@@ -83,10 +101,16 @@ public HollowDiff(HollowReadStateEngine from, HollowReadStateEngine to, boolean
...
                 if (schema instanceof HollowObjectSchema) {
-                    PrimaryKey pKey = ((HollowObjectSchema) schema).getPrimaryKey();
-                    if (pKey == null) continue;
+                    HollowObjectSchema objectSchema = ((HollowObjectSchema) schema);
+                    PrimaryKey pKey = objectSchema.getPrimaryKey();
+                    if (pKey==null && !isIncludeNonPrimaryKeyTypes) continue;
+
+                    // Support basic Single Field Types
+                    if (pKey==null && objectSchema.numFields()==1 && SINGLE_FIELD_SUPPORTED_TYPES.contains(objectSchema.getFieldType(0))) {
+                        pKey = new PrimaryKey(schema.getName(), objectSchema.getFieldName(0));
```
Sunjeet, line 110: Are the values in this field guaranteed to be unique?

hollow/src/main/java/com/netflix/hollow/tools/diff/HollowDiff.java
```diff
@@ -60,7 +64,7 @@
      * @param to the "to" state
      */
     public HollowDiff(HollowReadStateEngine from, HollowReadStateEngine to) {
-        this(from, to, true);
+        this(from, to, true, false);
```
Sunjeet, line 67: nit: passing a final static boolean for eg. `DONT_INCLUDE_NON_PRIMARY_KEY_TYPES = false` might be good for readability

ghost (approved):
👍

hollow/src/main/java/com/netflix/hollow/tools/diff/HollowDiff.java
davesu: It is a single Field ObjectType so it would be interesting if Hollow allows duplicate records, it should have de-duped them.

hollow/src/main/java/com/netflix/hollow/tools/diff/HollowDiff.java
davesu: Fair point but I was just going with the flow since existing code base does not do that :)

Also IDEs like Intellij seems to do a good job showing the param names.

Sunjeet (approved):

PR #507, opened 2021-02-08
DABP-738 Display custom header string and values passed from clients on Hollow explorer

nayanika-u:
… a deep link back to go/splashui for the topic.

### DIFF
--- a/hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/HollowExplorerUI.java
+++ b/hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/HollowExplorerUI.java
@@ -26,6 +26,7 @@ import com.netflix.hollow.explorer.ui.pages.ShowAllTypesPage;
 import com.netflix.hollow.ui.HollowUIRouter;
 import com.netflix.hollow.ui.HollowUISession;
 import java.io.IOException;
+import java.util.HashMap;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
@@ -35,9 +36,12 @@ public class HollowExplorerUI extends HollowUIRouter {
     private final HollowConsumer consumer;
     private final HollowClient client;
     private final HollowReadStateEngine stateEngine;
-    
-    private String headerDisplayString;
-    
+    /** General purpose map to store header string and corresponding values used in HollowExplorer. It also stores headerDisplayString
+     *  for backwards compatibility.
+     **/
+    private final HashMap<String, String> headerDisplayMap = new HashMap<>();
+    private final static String HEADER_DISPLAY_STRING = "headerDisplayString";
+
     private final ShowAllTypesPage showAllTypesPage;
     private final BrowseSelectedTypePage browseTypePage;
     private final BrowseSchemaPage browseSchemaPage;
@@ -105,13 +109,21 @@ public class HollowExplorerUI extends HollowUIRouter {
             return client.getStateEngine();
         return stateEngine;
     }
-    
-    public void setHeaderDisplayString(String str) {
-        this.headerDisplayString = str;
-    }
-    
+
     public String getHeaderDisplayString() {
-        return headerDisplayString;
+        return headerDisplayMap.get(HEADER_DISPLAY_STRING);
+    }
+
+    public void setHeaderDisplayString(String str) {
+        this.headerDisplayMap.put(HEADER_DISPLAY_STRING, str);
+    }
+
+    public void addToHeaderDisplayMap(String key, String value) {
+        headerDisplayMap.put(key, value);
+    }
+
+    public String getFromHeaderDisplayMap(String key) {
+        return headerDisplayMap.get(key);
     }
     
 }
--- a/hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/pages/HollowExplorerPage.java
+++ b/hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/pages/HollowExplorerPage.java
@@ -16,6 +16,7 @@
  */
 package com.netflix.hollow.explorer.ui.pages;
 
+import com.netflix.hollow.explorer.ui.HollowExplorerUI;
 import com.netflix.hollow.ui.EscapingTool;
 import com.netflix.hollow.ui.HollowUISession;
 import java.io.IOException;
--- a/hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/pages/HollowExplorerPage.java
+++ b/hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/pages/HollowExplorerPage.java
@@ -16,7 +16,6 @@
  */
 package com.netflix.hollow.explorer.ui.pages;
 
-import com.netflix.hollow.explorer.ui.HollowExplorerUI;
 import com.netflix.hollow.ui.EscapingTool;
 import com.netflix.hollow.ui.HollowUISession;
 import java.io.IOException;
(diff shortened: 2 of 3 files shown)


hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/HollowExplorerUI.java
```diff
@@ -105,13 +114,16 @@ public HollowReadStateEngine getStateEngine() {
             return client.getStateEngine();
         return stateEngine;
     }
-    
-    public void setHeaderDisplayString(String str) {
```
akhaku, line 109: is this going to break anyone else downstream?

hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/HollowExplorerUI.java
```diff
@@ -85,8 +88,14 @@ public boolean handle(String target, HttpServletRequest req, HttpServletResponse
         } else if("query".equals(pageName)) {
             queryPage.render(req, resp, session);
             return true;
+        } else if("go".equals(pageName)) {
+            // The target has / prefixed in namespaceURL, this is to escape it.
+            if(target.startsWith("/")) {
+                target = target.substring(1);
+            }
+            URL namespaceURL = new URL("http://"+target);
```
akhaku: I don't thing we should say anything about namespace here, right? Since hollow has no concept of namespace. This is really a generic url for the UI.
Also, not entirely sure what we're doing by stripping out the slash, are we trying to disallow absolute urls for some reason?

hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/HollowExplorerUI.java
```diff
@@ -105,13 +114,16 @@ public HollowReadStateEngine getStateEngine() {
             return client.getStateEngine();
         return stateEngine;
     }
-    
-    public void setHeaderDisplayString(String str) {
-        this.headerDisplayString = str;
+
+    public void addToHeaderStringMap(String var, String str) {
+        if (headerDisplayMap == null) {
+            headerDisplayMap = new HashMap<String, String>(5); //1.5 times the initial count of elements(2)
```
akhaku: Sorta feels like this is an unnecessary optimization - should we just initialize headerDisplayMap as an empty HashMap when we declare it?
Also, any need to have an addTo method or should we have users call getHeaderDisplayMap and then `.put`? Or it the getter only meant for the UI? If so, the getter should probably return an immutable view of the map.

hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/pages/HollowExplorerPage.java
```diff
@@ -43,12 +43,22 @@ public HollowExplorerPage(HollowExplorerUI ui) {
...
-        if(ui.getHeaderDisplayString() != null)
-            ctx.put("headerDisplayString", ui.getHeaderDisplayString());
-        
-        if(ui.getCurrentStateVersion() != Long.MIN_VALUE)
+        if (ui.getCurrentStateVersion() != Long.MIN_VALUE)
             ctx.put("stateVersion", ui.getCurrentStateVersion());
-        
+
+        if (ui.getHeaderDisplayMap() != null) {
+            String environment = ui.getHeaderDisplayMap().getOrDefault("environment", "");
```
akhaku: I don't think hollow should know anything about environment or namespace. We could use the keys to mean something special in the UI though, I can walk you through what I mean over a video call

akhaku (changes requested):

hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/HollowExplorerUI.java
nayanika-u: You're right. I was just trying to construct an absolute URL from the one being passed from the page. The target I receive from the explorer page is "/go/cinderdashboard...". It was a hacky way to get the link working. The namespaceURL part isn't final yet. Still figuring out a clean way to get it running from explorer link.

hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/HollowExplorerUI.java
nayanika-u: Will remove the size of hashmap.
This method gets called from ExploreCommand in cinder-cli. Added that method name for more readability. Correct, the getter method for this map is being called in HollowExplorerPage (UI).

hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/pages/HollowExplorerPage.java
nayanika-u: As discussed, will change the keys to generic names to avoid confusion in Hollow.

hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/HollowExplorerUI.java
nayanika-u: Yes. This will be a breaking change in Cinder-cli because we pass namespace to the explorer using this method.

nayanika-u: Another way to do it with minimal change is to just pass the namespace and env as space separated strings in headerDisplayString.

hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/HollowExplorerUI.java
```diff
@@ -26,18 +26,22 @@
...
     private final HollowClient client;
     private final HollowReadStateEngine stateEngine;
-    
-    private String headerDisplayString;
-    
+
+    /** General purpose map to store header string and corresponding values used in HollowExplorer. It also stores headerDisplayString
+     *  for backwards compatibility.
+     **/
+    private Map<String, String> headerDisplayMap = new HashMap<>();
```
akhaku: minor: `final`

hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/HollowExplorerUI.java
```diff
@@ -105,13 +108,21 @@ public HollowReadStateEngine getStateEngine() {
             return client.getStateEngine();
         return stateEngine;
     }
-    
+
+    public String getHeaderDisplayString() {
+        return headerDisplayMap.get("headerDisplayString");
```
akhaku: minor: might be good to move this to a private static final constant

hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/pages/HollowExplorerPage.java
```diff
@@ -43,12 +43,18 @@ public HollowExplorerPage(HollowExplorerUI ui) {
...
-        if(ui.getCurrentStateVersion() != Long.MIN_VALUE)
+        if (ui.getCurrentStateVersion() != Long.MIN_VALUE)
             ctx.put("stateVersion", ui.getCurrentStateVersion());
-        
+
+        String headerDisplayString = ui.getHeaderDisplayString();
+        ctx.put("headerDisplayString", headerDisplayString);
+
+        if (headerDisplayString != null) {
+            String headerDisplayURL = ui.getFromHeaderDisplayMap(headerDisplayString);
```
akhaku: might be good to null check or `containsKey` on this, otherwise we'll have a link that contains the word `null`

hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/pages/HollowExplorerPage.java
```diff
@@ -43,12 +43,18 @@ public HollowExplorerPage(HollowExplorerUI ui) {
...
+            headerDisplayURL = headerDisplayURL + "/" + headerDisplayString;
```
akhaku: is this intentional?

akhaku (approved):

hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/HollowExplorerUI.java
nayanika-u: done

hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/HollowExplorerUI.java
nayanika-u: done

hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/pages/HollowExplorerPage.java
nayanika-u: thanks for pointing out this one. I was earlier adding a getOrDefault in the getter method itself with default as "" but decided to against it for not having a common default values for all entries in the map. Added null check here.

hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/pages/HollowExplorerPage.java
nayanika-u: yes. headerDisplayURL is supposed to be a go link which auto expands to a URL where I have to explicitly add "/" to append the env and namespace values.

nayanika-u: Removed the URL creation part from Hollow. This decision now happens in the internal client.

PR #508, opened 2021-02-10
Optional blob parts: configure specific types into blob 'parts'

dkoszewnik:
Blob 'parts' are published and retrieved as separate files.  Clients can download and load only the 'parts' they are interested in.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright 2016-2019 Netflix, Inc.
+ *  Copyright 2016-2021 Netflix, Inc.
  *
  *     Licensed under the Apache License, Version 2.0 (the "License");
  *     you may not use this file except in compliance with the License.
@@ -32,6 +32,7 @@ import com.netflix.hollow.api.metrics.HollowConsumerMetrics;
 import com.netflix.hollow.api.metrics.HollowMetricsCollector;
 import com.netflix.hollow.core.HollowConstants;
 import com.netflix.hollow.core.memory.MemoryMode;
+import com.netflix.hollow.core.read.OptionalBlobPartInput;
 import com.netflix.hollow.core.read.engine.HollowReadStateEngine;
 import com.netflix.hollow.core.read.filter.HollowFilterConfig;
 import com.netflix.hollow.core.read.filter.TypeFilter;
@@ -44,6 +45,7 @@ import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
 import java.util.concurrent.RejectedExecutionException;
@@ -462,6 +464,11 @@ public class HollowConsumer {
          * @return the blob of the reverse delta
          */
         HollowConsumer.Blob retrieveReverseDeltaBlob(long currentVersion);
+
+        default Set<String> configuredOptionalBlobParts() {
+            return null;
+        }
+
     }
 
     /**
@@ -521,6 +528,19 @@ public class HollowConsumer {
          * @throws IOException if the input stream to the blob cannot be obtained
          */
         public abstract InputStream getInputStream() throws IOException;
+
+        /**
+         * Implementations may define how to retrieve the optional blob part data for this specific transition from a data store.
+         * <p>
+         * It is expected that none of the returned InputStreams will be interrupted.  For this reason, it is a good idea to
+         * retrieve the entire blob part data (e.g. to disk) from a remote datastore prior to returning these streams.
+         * 
+         * @return
+         * @throws IOException
+         */
+        public OptionalBlobPartInput getOptionalBlobPartInputs() throws IOException {
+            return null;
+        }
 
         public File getFile() throws IOException {
             throw new UnsupportedOperationException();
--- a/hollow/src/main/java/com/netflix/hollow/api/consumer/fs/HollowFilesystemBlobRetriever.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/consumer/fs/HollowFilesystemBlobRetriever.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright 2016-2019 Netflix, Inc.
+ *  Copyright 2016-2021 Netflix, Inc.
  *
  *     Licensed under the Apache License, Version 2.0 (the "License");
  *     you may not use this file except in compliance with the License.
@@ -19,7 +19,10 @@ package com.netflix.hollow.api.consumer.fs;
 import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
 
 import com.netflix.hollow.api.consumer.HollowConsumer;
+import com.netflix.hollow.api.consumer.HollowConsumer.Blob.BlobType;
 import com.netflix.hollow.core.HollowConstants;
+import com.netflix.hollow.core.read.HollowBlobInput;
+import com.netflix.hollow.core.read.OptionalBlobPartInput;
 import java.io.BufferedInputStream;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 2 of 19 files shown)


hollow/src/main/java/com/netflix/hollow/api/consumer/fs/HollowFilesystemBlobRetriever.java
```diff
@@ -138,7 +202,7 @@ public HollowFilesystemBlobRetriever(Path blobStorePath, HollowConsumer.BlobRetr
                 String filename = path.getFileName().toString();
                 if(filename.startsWith("delta-" + currentVersion)) {
```
Sunjeet, line 202: this won't match on `"delta_"+part+"-"+currentVersion+"-"+destinationVersion` is that intentional?

hollow/src/main/java/com/netflix/hollow/api/consumer/fs/HollowFilesystemBlobRetriever.java
```diff
@@ -260,5 +347,37 @@ public File getFile() throws IOException {
...
+        @Override
+        public OptionalBlobPartInput getOptionalBlobPartInputs() throws IOException {
+            OptionalBlobPartInput remoteOptionalParts = remoteBlob.getOptionalBlobPartInputs();
+            if(remoteOptionalParts == null)
+                return null;
+
+            OptionalBlobPartInput localOptionalParts = new OptionalBlobPartInput();
+
+
+            for(Map.Entry<String, HollowBlobInput> entry : remoteOptionalParts.getInputsByPartName().entrySet()) {
```
Sunjeet: wonder if there'd be much to gain by doing these in parallel

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
```diff
@@ -462,6 +464,11 @@ public HollowConsumerMetrics getMetrics() {
          * @return the blob of the reverse delta
          */
         HollowConsumer.Blob retrieveReverseDeltaBlob(long currentVersion);
+
+        default Set<String> configuredOptionalBlobParts() {
```
Sunjeet, line 468: nit: alternatively,
`default Optional<Set<String>> configuredBlobParts`
But given the number of touch points maybe not worth it

hollow/src/main/java/com/netflix/hollow/core/read/OptionalBlobPartInput.java
```diff
@@ -0,0 +1,44 @@
...
+
+public class OptionalBlobPartInput {
+
+    private final Map<String, HollowBlobInput> inputsByPartName;
+
+    public OptionalBlobPartInput() {
+        this.inputsByPartName = new HashMap<>();
+    }
+
+    public void addInput(String partName, InputStream in) {
```
Sunjeet, line 41: minor: I think you already considered it, but we could potentially drop this one given its in the new code path.

hollow/src/main/java/com/netflix/hollow/api/producer/ProducerOptionalBlobPartConfig.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Function;
+
+/**
+ * A blob may be configured to produce optional "parts".  Parts will contain the data one or more types.
+ * 
+ * A type may be assigned to one and only one section -- by default a type will be serialized to the main "part".
```
Sunjeet, line 33: Curious about your thoughts on configurable "dual write" for producer i.e. if writing in parts then also write a full snapshot - so not all consumers need to have upgraded for a producer to onboard parts

hollow/src/main/java/com/netflix/hollow/core/HollowBlobOptionalPartHeader.java
```diff
@@ -0,0 +1,64 @@
...
+ *     limitations under the License.
+ *
+ */
+package com.netflix.hollow.core;
+
+import com.netflix.hollow.core.schema.HollowSchema;
+import java.util.Collections;
+import java.util.List;
+
+public class HollowBlobOptionalPartHeader {
```
Sunjeet, line 23: I noticed you dropped `Map<String, String>  headerTags` from the part headers? Does that mean we envision consumers to always consume the "main" part?

Sunjeet:
IIUC if type A and type B both depend on String, and type A and B are configured in different parts by the producer, both part blobs will include all the data in the String type. So for now, a consumer of just type B will hold all String data in memory, even though some is not referenced by type B?

hollow/src/main/java/com/netflix/hollow/core/read/engine/HollowBlobReader.java
```diff
@@ -147,6 +165,17 @@ public void readSnapshot(HollowBlobInput in, TypeFilter filter) throws IOExcepti
             typeNames.add(typeName);
         }
 
+        if(optionalParts != null) {
+            for(Map.Entry<String, HollowBlobInput> optionalPartEntry : optionalParts.getInputsByPartName().entrySet()) {
+                numStates = VarInt.readVInt(optionalPartEntry.getValue());
+
+                for(int i=0;i<numStates;i++) {
+                    String typeName = readTypeStateSnapshot(optionalPartEntry.getValue(), filter);
```
Sunjeet, line 175: Just wanted to confirm, does the order in which `readTypeStateSnapshot` is called on types matter? I noticed we use a `TreeSet` above

Sunjeet:
The usage of `HollowBlobInput` looked good to me from reading the diff, and from what I can tell shared-memory mode seems supported in the new feature. Thank you! I'll be happy to test out shared memory mode with blob parts for you once this change is merged (or if you wanted to add in a test that uses HollowConsumer built with `.withMemoryMode` that should work too)

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
Sunjeet: also, came across https://justamonad.com/wrapping-collection-in-optional-is-a-big-no-no/

hollow/src/main/java/com/netflix/hollow/core/read/engine/HollowBlobReader.java
dkoszewnik: It shouldn't matter.  During an update references on some types may be incorrect *but* none of the consumer's indexing structures should point to any of the new data until the update is fully complete.

hollow/src/main/java/com/netflix/hollow/core/HollowBlobOptionalPartHeader.java
dkoszewnik: That is the idea.  The main part might end up having no types (if all types were optional), but the headers would still be consumed from the main part.  This keeps the APIs consistent and backwards-compatible.

hollow/src/main/java/com/netflix/hollow/api/producer/ProducerOptionalBlobPartConfig.java
dkoszewnik: I'm not opposed.  I think this can be accomplished in a subsequent iteration if we find a need.

hollow/src/main/java/com/netflix/hollow/core/read/OptionalBlobPartInput.java
dkoszewnik: I didn't add it originally, but I found myself wrapping the `InputStream` in multiple places so putting the convenience method here saved time.

hollow/src/main/java/com/netflix/hollow/api/consumer/fs/HollowFilesystemBlobRetriever.java
dkoszewnik: Yes, intentional, we only want to match the main blob part here.

hollow/src/main/java/com/netflix/hollow/api/consumer/fs/HollowFilesystemBlobRetriever.java
dkoszewnik: I think we don't want that -- we don't want to use up all the CPU on the consumer instances to do our delta updates -- taking a bit longer is preferable.

Sunjeet (approved):

PR #512, opened 2021-02-22
Data Corruption Fix: overflow when reading values from arrays

mikemuske:
My company runs many Hollow based apps in production and we noticed data corruption with a couple datasets that used a large number of array elements. After debugging we found that the failure occurs when the dataset consists of more than 2^29 array elements. It does not matter how many elements are in any individual array. Cumulatively there must be less than 2^29 array elements across all arrays.

The problem was happening because HollowListTypeReadStateShard tries to read 2 values at once, which results in a 58 bit read. However the 29 bit value can be offset by 7 bits from a byte boundary, causing the result of a 58 bit read to overflow the 64 bit space and an invalid value to be returned.

The fix was to modify HollowListTypeReadStateShard to read the values individually.

Unit tests are provided demonstrating the situations described above. In addition, we found that FixedLengthData values are actually limited to 58 bits in length, contrary to the comments in code which claim it can support 60 bit wide values. This is because a 59 bit wide value will overflow when offset by 6 bits from a byte boundary. There are also unit tests included to demonstrate this.

Issue: #513

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/memory/FixedLengthData.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/memory/FixedLengthData.java
@@ -16,7 +16,7 @@ import java.io.IOException;
  * the value 100100 in binary, or 36 in base 10, would be returned. </p>
  * <p>
  *     As a result there two ways to obtain an element value from the bit string at a given bit index.  The first,
- * using {@link #getElementValue} for values less than 61 bits in length and the second, using
+ * using {@link #getElementValue} for values less than 59 bits in length and the second, using
  * {@link #getLargeElementValue} that is recommended for values of upto 64 bits in length.
  * </p>
  */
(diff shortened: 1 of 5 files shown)


mikemuske:
@rpalcolea Hello from Target

erichw:
@toolbear @dkoszewnik If you have a moment to take a look at this, it would be much appreciated. We've hit this overflow problem with a couple production clients now, and it effectively makes them unable to process new updates. Causing an ArrayIndexOutOfBoundsException or DELTA CHECKSUM errors with all subsequent Producer cycles.

mikemuske:
#512

dkoszewnik (approved):
LGTM -- if we notice a performance regression we can add a check for more than 58 bits and pivot between the single-read and double-read logic.

zurdoron:
@mikemuske can you please elaborate on the diagnosis process for this issue? we are experiencing similar symptoms on HollowSetTypeReadStateShard in long running consumer instances

PR #518, opened 2021-02-27
Support ARM64/AWS Graviton Instances by not performing unaligned accesses

### LINKED ISSUE #517: hollow on ARM
Hollow isn't supported on ARM instances, unaligned memory access causes SIGBUS.

In doing some preliminary testing, we were able to localize the failure (atleast on the consumer side) to sun.misc.Unsafe.putOrderedLong, which is invoked in the Hollow delta application code path when transitioning from one data version to another. Although it is not definitive if that is the only failure site, because putOrderedLong and other non aligned access are called from several other code paths. But in our limited testing, we were able to pin the consumer to block any delta updates, then run some benchmarking that included non aligned reads. 

Results from running the Hollow microbenchmarks on m6g.4xlarge (arm) against m5.4xlarge (amd64) are attached for exploration. These files can be unzipped and be dropped into https://jmh.morethan.io/ et al for viz. The perf code was run twice at different times of day and showed consistent results for memory and CPU. On a high level, ARM instances showed 70-90% improvement in memory read intensive workload. 

[arm.json.zip](https://github.com/Netflix/hollow/files/6053524/arm.json.zip)
[amd64.json.zip](https://github.com/Netflix/hollow/files/6053525/amd64.json.zip)

Dillonb:
Resolves https://github.com/Netflix/hollow/issues/517 and allows Hollow to be used on ARM based machines such as AWS Graviton instances.

There are a few things left to finish up, but I wanted to get this in front of people sooner rather than later to get some feedback on the design here.

What I've done is swapped HollowUnsafeHandle from being simply a way to get an instance of sun.misc.Unsafe to a singleton class wrapping sun.misc.Unsafe. It seems the only classes that perform unaligned accesses are the FixedLengthElementArray and SegmentedLongArray families, so it might also make sense to have only those classes call an unaligned helper method on HollowUnsafeHandle, rather than having all unsafe calls go through this class. Let me know if this would be preferred.

I've tested this (but only the consumer side) on an r6g.4xlarge instance on AWS, and performance is very good. Also, the kernel doesn't kill it with a SIGBUS :slightly_smiling_face: 

I'm new to the sun.misc.Unsafe API, so there are a few things I haven't been able to figure out how to do yet:

- How do we detect whether the host platform supports unaligned accesses?

OpenJDK 8 does it like [this internally](https://github.com/openjdk/jdk/blob/9a9add8825a040565051a09010b29b099c2e7d49/jdk/src/share/classes/java/nio/Bits.java#L606-L615), but there seems to not be a public-facing API version of this, at least in JDK8. Maybe a configuration setting would be most appropriate if there's no reliable way to auto detect it.

- Improve efficiency by checking the alignment of the address

I'd like to check if the address is aligned, and use the faster putOrderedLong/getLong calls directly if possible. Since all the accesses use the object + offset form instead of raw addresses, there doesn't seem to be a way to do this unless we can retrieve the address of the base object, which doesn't seem to be possible through sun.misc.Unsafe.

Thanks for your time.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/memory/HollowUnsafeHandle.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/memory/HollowUnsafeHandle.java
@@ -51,19 +51,29 @@ public class HollowUnsafeHandle {
         return singleton;
     }
 
-    public void putOrderedLong(Object o, long l, long l1) {
-        if (UNALIGNED_ALLOWED) {
-            unsafe.putOrderedLong(o, l, l1);
+    public void putOrderedLong(Object o, long offset, long value) {
+        if (UNALIGNED_ALLOWED || (offset & 7) == 0) {
+            unsafe.putOrderedLong(o, offset, value);
+        } else if ((offset & 3) == 0) {
+            // 4 byte aligned, little endian
+            unsafe.putOrderedInt(o, offset, (int)(value));
+            unsafe.putOrderedInt(o, offset + 4, (int)(value >>> 32));
+        } else if ((offset & 1) == 0) {
+            // 2 byte aligned, little endian
+            unsafe.putShortVolatile(o, offset, (short)(value));
+            unsafe.putShortVolatile(o, offset + 2, (short)(value >>> 16));
+            unsafe.putShortVolatile(o, offset + 4, (short)(value >>> 32));
+            unsafe.putShortVolatile(o, offset + 6, (short)(value >>> 48));
         } else {
-            // Assume little endian
-            unsafe.putByteVolatile(o, l, (byte) (l1));
-            unsafe.putByteVolatile(o, l + 1, (byte) (l1 >>> 8));
-            unsafe.putByteVolatile(o, l + 2, (byte) (l1 >>> 16));
-            unsafe.putByteVolatile(o, l + 3, (byte) (l1 >>> 24));
-            unsafe.putByteVolatile(o, l + 4, (byte) (l1 >>> 32));
-            unsafe.putByteVolatile(o, l + 5, (byte) (l1 >>> 40));
-            unsafe.putByteVolatile(o, l + 6, (byte) (l1 >>> 48));
-            unsafe.putByteVolatile(o, l + 7, (byte) (l1 >>> 56));
+            // 1 byte aligned, little endian
+            unsafe.putByteVolatile(o, offset, (byte) (value));
+            unsafe.putByteVolatile(o, offset + 1, (byte) (value >>> 8));
+            unsafe.putByteVolatile(o, offset + 2, (byte) (value >>> 16));
+            unsafe.putByteVolatile(o, offset + 3, (byte) (value >>> 24));
+            unsafe.putByteVolatile(o, offset + 4, (byte) (value >>> 32));
+            unsafe.putByteVolatile(o, offset + 5, (byte) (value >>> 40));
+            unsafe.putByteVolatile(o, offset + 6, (byte) (value >>> 48));
+            unsafe.putByteVolatile(o, offset + 7, (byte) (value >>> 56));
         }
     }
 
--- a/hollow/src/main/java/com/netflix/hollow/core/memory/HollowUnsafeHandle.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/memory/HollowUnsafeHandle.java
@@ -26,6 +26,12 @@ import sun.misc.Unsafe;
 public class HollowUnsafeHandle {
     private static final Logger log = Logger.getLogger(HollowUnsafeHandle.class.getName());
     private static final Unsafe unsafe;
+    private static final HollowUnsafeHandle singleton;
+
+    // TODO: figure this out somehow
+    private static final boolean UNALIGNED_ALLOWED = false;
+
+    private HollowUnsafeHandle() {}
 
     static {
         Field theUnsafe;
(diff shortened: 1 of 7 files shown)


hollow/src/main/java/com/netflix/hollow/core/memory/HollowUnsafeHandle.java
```diff
@@ -38,10 +44,102 @@
             log.log(Level.SEVERE, "Unsafe access failed", e);
         }
         unsafe = u;
+        singleton = new HollowUnsafeHandle();
+    }
+
+    public static HollowUnsafeHandle getUnsafe() {
+        return singleton;
```
Sunjeet, line 51: We could return an implementation of `HollowUnsafeHandle` that always does unaligned, or an instance that always does aligned. Then we don't incur the cost of branching in the read path.

hollow/src/main/java/com/netflix/hollow/core/memory/HollowUnsafeHandle.java
```diff
@@ -38,10 +44,102 @@
...
+            unsafe.putByteVolatile(o, l + 3, (byte) (l1 >>> 24));
+            unsafe.putByteVolatile(o, l + 4, (byte) (l1 >>> 32));
+            unsafe.putByteVolatile(o, l + 5, (byte) (l1 >>> 40));
+            unsafe.putByteVolatile(o, l + 6, (byte) (l1 >>> 48));
+            unsafe.putByteVolatile(o, l + 7, (byte) (l1 >>> 56));
+        }
+    }
+
+    public long getLong(Object o, long l) {
+        if (UNALIGNED_ALLOWED) {
```
Sunjeet: I'm curious, if you change this to `if (true)` do you see the SIGBUS? In our testing the only place where we saw SIGBUS was when the delta application code calls `putOrderedLog`. Once we disabled deltas, everything else (initializing from a snapshot, creating indices, un aligned reads, etc.) seemed to work.

Sunjeet: Asking because of performance criticality here

hollow/src/main/java/com/netflix/hollow/core/memory/HollowUnsafeHandle.java
```diff
@@ -38,10 +44,102 @@
...
+            return unsafe.getLong(o, l);
+        } else {
+            // Assume little endian
+            return ((long) unsafe.getByte(o, l) & 0xFF)
```
Sunjeet: If we end up needing to do this, it will likely be faster to batch these up into- read 2 words when a read spans a word boundary then stitch up the long. 
If I am reading this right the `getByte`s translate into reading word or half word anyway https://developer.arm.com/documentation/ddi0360/f/unaligned-and-mixed-endian-data-access-support/unaligned-access-support/armv6-extensions

Sunjeet:
> It seems the only classes that perform unaligned accesses are the FixedLengthElementArray and SegmentedLongArray families

To double check, `EncodedByteBuffer` and `EncodedLongBuffer` also do this

hollow/src/main/java/com/netflix/hollow/core/memory/HollowUnsafeHandle.java
Dillonb: I'm fairly certain that, since the guard is a `final`, the JVM will optimize out the branch not-taken when it compiles this code to native, but I can swap to using two implementations to be more certain.

Sunjeet:
> It seems the only classes that perform unaligned accesses are the FixedLengthElementArray and SegmentedLongArray families, so it might also make sense to have only those classes call an unaligned helper method on HollowUnsafeHandle, rather than having all unsafe calls go through this class.

That sounds fine. Something else to consider is that we'll at some point need to replace all use of sun.misc.Unsafe with MemorySegment et. al. https://openjdk.java.net/jeps/370

hollow/src/main/java/com/netflix/hollow/core/memory/HollowUnsafeHandle.java
Dillonb: I'm fairly certain that, since the guard is a final, the JVM will optimize out the branch not-taken when it compiles this code to native, but I can swap to using two implementations to be more certain.

hollow/src/main/java/com/netflix/hollow/core/memory/HollowUnsafeHandle.java
Dillonb: I'll test this and get back to you.

Sunjeet:
Hi @Dillonb thanks for looking into this! To me your proposed approach to add ARM support is welcome, it allows us to sidestep the SIGBUS error and incrementally get to a performant and stable Hollow on ARM (note: thats low on the priority list for us at the moment). I think with a little tweak to your pr (make HollowUnsafeHandle polymorphic) we can achieve this without introducing performance regression in the x86/amd64 path.

Also adding @dkoszewnik and @toolbear for visibility, and if they have any concerns or additional perspective on going down this road.

hollow/src/main/java/com/netflix/hollow/core/memory/HollowUnsafeHandle.java
Dillonb: Yeah, one of the points I mentioned was detecting when accesses are 1, 2, 4, or 8 bytes aligned and performing the most efficient one (any performance penalty from the branches should be outweighed by the gains from larger reads/writes)

At the time, I wasn't sure how to determine this, since all accesses are done with "object plus offset" methods - but it seems that heap objects are always 8-byte aligned, so we should be able to figure this out based on the offset alone. I'll update the PR with this optimization.

Dillonb:
Thanks for the feedback!

Seems the last thing remaining to figure out is how to detect if unaligned accesses are allowed on the host. I'm really hoping to not have to go down the road of string matching on `os.arch`, but it's looking more and more like that's the way forward.

dkoszewnik:
This is awesome.  No concerns from me with the approach, and glad to see this being tackled as the need arises.  Thanks a lot for making this happen!

hollow/src/main/java/com/netflix/hollow/core/memory/HollowUnsafeHandle.java
Sunjeet: Ah I didn't consider that.

hollow/src/main/java/com/netflix/hollow/core/memory/HollowUnsafeHandle.java
Sunjeet: Heads up the ByteBuffers in `EncodedByteBuffer` and `EncodedLongBuffer` are off heap and won't be word-aligned, I think

Sunjeet:
@PaulSandoz brought up a couple of interesting points offline, 
* do we really need `putOrderedLong` here, or could we switch to plain store `putLong` that might just work for unaligned access on ARM? 
* this looks a bit cleaner-
```
 if ((offset & 7) == 0) {
        return getLong(o, offset);
    } else if ((offset & 3) == 0) {
        return makeLong(getInt(o, offset),
                        getInt(o, offset + 4));
    } else if ((offset & 1) == 0) {
        return makeLong(getShort(o, offset),
                        getShort(o, offset + 2),
                        getShort(o, offset + 4),
                        getShort(o, offset + 6));
    } else {
        return makeLong(getByte(o, offset),
                        getByte(o, offset + 1),
                        getByte(o, offset + 2),
                        getByte(o, offset + 3),
                        getByte(o, offset + 4),
                        getByte(o, offset + 5),
                        getByte(o, offset + 6),
                        getByte(o, offset + 7));
    }
```

Dillonb:
Ah, switching to straight `.putLong` from `.putOrderedLong` actually does resolve the SIGBUS issue. Must be that memory fences must be aligned, but stores can be unaligned. Thanks for the idea!

Dillonb:
I've created https://github.com/Netflix/hollow/pull/519 with this new approach.

PaulSandoz:
@Sunjeet @Dillonb thanks. I forgot to add that the proposed unaligned ordered put changes the memory order and atomicity guarantees, so it no longer conforms to the semantics of an ordered put (release store). A read can now observe a partial write.

I cannot recall the exact concurrency requirements of these array-like classes, but It's not clear to me why the ordered put (release store) is needed. It's never paired with an acquire load. If not needed i think it greatly simplify things if it can be replaced with a plain atomic store on x64 and ARM, thereby preserving the atomicity guarantees (which i think may be the more important property).

Dillonb:
@PaulSandoz just to be absolutely safe, in #519 I've added a storeFence() after every call that I changed from putOrderedLong to putLong, to keep the ordering and atomicity guarantees. Performance seemed to be fine, but I didn't test it on any absurdly large datasets 🙂

Sunjeet:
https://github.com/Netflix/hollow/pull/519 was accepted, thanks @Dillonb !

PR #519, opened 2021-03-02
Replace all calls to putOrderedLong with putLong

Dillonb:
An alternative approach to #518 

It seems like the culprit wasn't the unaligned write, but rather the memory fence that putOrderedLong creates. 

This is a much cleaner solution that still provides all the safety of putOrderedLong through the calls to storeFence().

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/memory/encoding/FixedLengthElementArray.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/memory/encoding/FixedLengthElementArray.java
@@ -192,7 +192,6 @@ public class FixedLengthElementArray extends SegmentedLongArray implements Fixed
         long l = unsafe.getLong(segment, elementByteOffset);
 
         unsafe.putLong(segment, elementByteOffset, l + (increment << whichBit));
-        unsafe.storeFence();
 
         /// update the fencepost longs
         if((whichByte & byteBitmask) > bitmask * 8 && (whichSegment + 1) < segments.length) {
--- a/hollow/src/main/java/com/netflix/hollow/core/memory/encoding/FixedLengthElementArray.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/memory/encoding/FixedLengthElementArray.java
@@ -197,11 +197,9 @@ public class FixedLengthElementArray extends SegmentedLongArray implements Fixed
         /// update the fencepost longs
         if((whichByte & byteBitmask) > bitmask * 8 && (whichSegment + 1) < segments.length) {
             unsafe.putLong(segments[whichSegment + 1], (long) Unsafe.ARRAY_LONG_BASE_OFFSET, segments[whichSegment][bitmask + 1]);
-            unsafe.storeFence();
         }
         if((whichByte & byteBitmask) < 8 && whichSegment > 0) {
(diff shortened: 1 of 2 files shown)


PaulSandoz:
I suspect the SIGBUS was caused by the store instruction associated with the `putOrdered` method (I am not familiar enough with ARM and HotSpot to know the exact sequence of instructions that are generated). 

The LoadStore|StoreStore fence (`s.m.Unsafe.storeFence`) is placed incorrectly. It should be placed *before* the store, such prior that loads and stores that occur before the fence are not reordered after it i.e. you want to make sure the store after the fence is not reordered (by the compiler or the CPU by some means). If you wanna get into the finer details see http://gee.cs.oswego.edu/dl/html/j9mm.html (and specifically section "RA Fences").

However, i don't think the  `storeFence` is required, since acquire reads are not paired with release (ordered) writes, thereby there are no ordering guarantees between writes and reads. 
It's not clear why `putOrdered` was used in the first place. Perhaps Hollow encountered a HotSpot compiler bug, and using `putOrdered` changed the set of applicable optimizations and worked around the bug?

Dillonb:
Interesting, thank you for the clarification. I had placed it after the store to ensure that loads after it were not reordered before it. I noticed some normal array accesses around some of the putOrderedLong calls, and thought it better to be cautious, because I'm new to the Java unsafe API.

All automated tests pass with the fences removed, and the consumer side works fine without them as well. 

It would be good to get some clarification from Netflix around why the ordered writes were used, as it looks like they were in place since before Hollow went open source.

To the maintainers: is this something you feel comfortable moving forward with? Anything I can help test, I'm willing to.

hollow/src/main/java/com/netflix/hollow/core/memory/encoding/FixedLengthElementArray.java
```diff
@@ -191,13 +191,16 @@ public void increment(long index, long increment) {
         long elementByteOffset = (long) Unsafe.ARRAY_LONG_BASE_OFFSET + (whichByte & byteBitmask);
         long l = unsafe.getLong(segment, elementByteOffset);
 
-        unsafe.putOrderedLong(segment, elementByteOffset, l + (increment << whichBit));
+        unsafe.putLong(segment, elementByteOffset, l + (increment << whichBit));
+        unsafe.storeFence();
```
Sunjeet: Is this one intentional?

Sunjeet:
Looping in @dkoszewnik to test his retention of why we might have intentionally gone with `putOrderedLong` here.

Re: comfort level of integrating this change, deferring to Drew's judgement. I'll just resurface (originally Tim's point) that the producer integrity check should protect us here but some producers also disable the integrity check for performance. I like the idea of introducing this change behind a flag that depends on integrity check being enabled, for now.

[An after-thought: delta application on the consumer side calls (before this change) calls putOrderedLong, corruption there won't be protected by the integrity check]

dkoszewnik:
I recall that there was a reason I used this API, but the potential issue I was considering was hypothetical -- IIRC, it had to do with an incomplete application of a delta being visible by a reading thread after the new state was swapped in.  I remember that at the time I ran some testing without the `putOrderedLong` attempting to observe the theoretical issue and was **not** able to observe it.

I left this in because it didn't hurt anything.  I think it's unlikely to cause an issue if you remove it.

Sunjeet:
Thanks everyone. Merging, and will bump major version for release

PR #520, opened 2021-03-02
Upgrade nebula.netflixoss to replace bintray publication and update TravisCI secrets

rpalcolea:
Hi folks,

As you might be aware, JFrog is sunsetting Bintray and JCenter: https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

This upgrades to Nebula NetflixOSS which will publish to Nebula's OSS repos and Maven Central.

Repositories based on status:

* https://netflixoss.jfrog.io/artifactory/maven-oss-snapshots
* https://netflixoss.jfrog.io/artifactory/maven-oss-candidates
* https://netflixoss.jfrog.io/artifactory/maven-oss-releases

Few notes in other changes:

* Upgraded to latest Gradle
* Your checkstyle `CustomImportOrder` rule is more strict now, that's why you will see empty lines being deleted
```
    <module name="CustomImportOrder">
        <property name="customImportOrderRules"
            value="STATIC###THIRD_PARTY_PACKAGE"/>
        <property name="standardPackageRegExp" value=""/> <!-- don't special case java/javax -->
        <property name="sortImportsInGroupAlphabetically" value="true"/>
        <property name="separateLineBetweenGroups" value="true"/>
    </module>
``` 
* Fixed some classpath issues by going from findbugs to spotbugs (`com.github.spotbugs:spotbugs:4.2.1`). `HollowCodeGenerationCompileUtil` still uses `Findbugs2` class but it is coming from spotbugs now. Gradle dropped findbugs a few releases ago because it is not maintained anymore and it became spotbugs
* The `javadoc` task in Gradle found warnings around invalid tags which prevented from upgrading, I fixed those


akhaku (approved):
Awesome, thank you very much for doing this! Especially pleased about the stricter checkstyle: my goal was to disallow those empty lines but checkstyle let them through

PR #522, opened 2021-03-12
Refresh successful after metrics updated

yonatang:
Right now, if someone will try to log the latest metrics using `RefreshListener`, they will get wrong results, because the event is emitted before the metrics are updated. If this is the first refresh, it will show 0 entities and 0 kb used. 

This is particularly annoying when trying to build a tool around Hollow that will integrate with existing corporate monitoring systems.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
@@ -175,14 +175,14 @@ public class HollowClientUpdater {
                 hollowDataHolderVolatile.update(updatePlan, localListeners, () -> {});
             }
 
-            for(HollowConsumer.RefreshListener refreshListener : localListeners)
-                refreshListener.refreshSuccessful(beforeVersion, getCurrentVersionId(), requestedVersion);
-
             metrics.updateTypeStateMetrics(getStateEngine(), requestedVersion);
             if(metricsCollector != null)
                 metricsCollector.collect(metrics);
 
             initialLoad.complete(getCurrentVersionId()); // only set the first time
+            for(HollowConsumer.RefreshListener refreshListener : localListeners)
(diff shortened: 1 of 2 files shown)


nayanika-u (commented):
Changes look good to me.

Edit : We might have to consider a workaround to solve your issue instead of going forward with these changes. 
It would not be safe to mark consumer initial load as completed before the refresh listeners are executed (i.e all indices are initialized) as the consumer will start accepting requests sooner than it should and will give erroneous results. Will update here with a workaround.

yonatang:
@nayanika-u, thanks for the comment. Is that true for the metric collectors as well? Would pushing event emitting loop (lines 183-185) above 
```java
initialLoad.complete(getCurrentVersionId()); 
``` 
but below 
```java
metrics.updateTypeStateMetrics(getStateEngine(), requestedVersion);
if(metricsCollector != null)
    metricsCollector.collect(metrics);
```
do the trick?

Sunjeet:
@yonatang your proposed tweak sounds good to me

[EDIT: I take my above approval back, thanks Nayanika for pointing out to me that if a refresh listener throws an exception then we would have already logged the success metric at that point, derp.]

Sunjeet:
Upon a closer look, @yonatang it might not be a good idea to log those metrics using a `RefreshListener`. Basically, regardless of whether we update those metrics before or after the refresh listeners are called, in both scenarios there are some undesirable outcomes. 

Have you considered specifying a metrics collector on the consumer? That seems to solve your problem of not seeing the latest metrics through a refresh listener.

yonatang:
@Sunjeet AFAIK metric collector can be specified during creation only. Since I'm thinking about kind of a middleware, I'd like to instrument existing consumers that were created externally (think about consumers that were created by `@Bean` in spring) - and in such case I cannot apply a metric collector in hindsight.

PR #527, opened 2021-04-14
OptionalBlobPartInput implements Closeable, and is closed after use

dkoszewnik:


hollow/src/main/java/com/netflix/hollow/core/read/OptionalBlobPartInput.java
```diff
@@ -18,27 +18,33 @@
...
         this.inputsByPartName = new HashMap<>();
+        this.streamsToClose = new ArrayList<>();
     }
     
     public void addInput(String partName, File file) {
         inputsByPartName.put(partName, file);
     }
 
     public void addInput(String partName, InputStream in) {
+        streamsToClose.add(in);
```
Sunjeet, line 47: Small preference for letting the layer that opens the streams also control whether they should be closed. So for eg. `getOptionalBlobPartInputs()` in `CinderBlobRetriever` could go from -
```
obpi.addInput(entry.getKey(), bis);
```
to 
```
obpi.addInput(entry.getKey(), bis);
obpi.streamsToClose.add(bis);
```
With the current implementation we'd only run into issues if we needed to close the `OptionalBlobPartInput` but reuse one of its input streams (because the stream was already closed by OptionalBlobPartInput). I don't think we'll have that issue given how we use these streams. But this little tweak might be a bit cleaner and consistent with the rest of the code base.

Sunjeet (approved):

PR #530, opened 2021-04-22
Delta transitions for TestHollowConsumer 

Sunjeet:


hollow-test/src/main/java/com/netflix/hollow/test/HollowWriteStateEngineBuilder.java
```diff
@@ -53,6 +54,17 @@ public HollowWriteStateEngineBuilder(Collection<Class<?>> types) {
         }
     }
 
+    /**
```
Sunjeet: credit: Tim's WIP implementation https://github.com/Netflix/hollow/blob/incubating/generic-object-data-access/hollow-test/src/main/java/com/netflix/hollow/test/HollowWriteStateEngineBuilder.java#L65-L74

hollow-test/src/test/java/com/netflix/hollow/test/consumer/TestHollowConsumerTest.java
```diff
@@ -98,4 +102,103 @@ public void testAddSnapshot_afterUpdate() throws Exception {
...
+        // delta transition to version 2
+        long v2 = 2L;
+        HollowWriteStateEngine deltaWriteStateEngine = new HollowWriteStateEngineBuilder(Collections.singletonList(Movie.class))
+                .restoreFrom(consumer.getStateEngine())
+                .add(new Movie(2, "bar"))
+                .build();
+        consumer.addDelta(consumer.getCurrentVersionId(), v2, snapshotWriteStateEngine, deltaWriteStateEngine);
+        consumer.triggerRefreshTo(v2);
+        data = consumer.getAPI().getDataAccess();
+        customAssertEquals(data, 0, new Movie(1, "foo"));   // ghost record
```
akhaku: I don't understand why this is a ghost record - we restored from v1 and added `Movie(2, "bar")` without removing `Movie(1,  "foo")`, right? I'm probably misunderstanding what `deltaWriteStateEngine` is and how `restoreFrom` works...

Sunjeet: `restoreFrom` is the same code path that a producer calls when restoring prior read state to continue the delta chain. 
`deltaWriteStateEngine` is similar to the workable write state for the producer cycle after restoring from prior read state.
Any records that aren't re-added to the write state will be considered removed.

akhaku: I see, thanks! It might be possible to simplify the API to something like this:
```
HollowStateEngine engine1 = new HollowWriteStateEngineBuilder(...);
HollowStateEngine engine2 = new HollowWriteStateEngineBuilder(...);
long v1 = 1L;
long v2 = 2L;
consumer.addSnapshot(v1, engine1);
consumer.addDelta(v1, v2, engine2);
```
Where you _must_ `addSnapshot(v1)` before `addDelta(v1, v2)`. You could also choose to `consumer.addSnapshot(v2, engine2)` after, if that's not too complex to implement.
We can handle the `withRestore` and `previousStateEngine` bits inside the addDelta call ourselves since we already have the state engine for the "from" version (which could be a snapshot state engine _or_ a delta state engine) and throw a helpful exception if the user added state engines in the wrong order.

Sunjeet: good suggestions, implemented

hollow-test/src/main/java/com/netflix/hollow/test/consumer/TestHollowConsumer.java
```diff
@@ -89,10 +95,38 @@ protected TestHollowConsumer(Builder builder) {
         this.blobRetriever = builder.blobRetriever();
     }
 
-    public TestHollowConsumer addSnapshot(long version, HollowWriteStateEngine stateEngine) throws IOException {
+    /**
+     * Apply a snapshot transition to {@code version} by applying {@code state}. This can be called on a {@code TestHollowConsumer}
```
akhaku, line 99: minor: javadoc param names don't match up

hollow-test/src/main/java/com/netflix/hollow/test/consumer/TestHollowConsumer.java
```diff
@@ -89,10 +95,38 @@ protected TestHollowConsumer(Builder builder) {
...
+     * that hasn't been initialized with a read state, or as a double-snapshot on a {@code TestHollowConsumer} with existing state.
+     */
+    public void applySnapshot(long toVersion, HollowWriteStateEngine writeState) throws IOException {
+        addSnapshot(toVersion, writeState);
+        triggerRefreshTo(toVersion);
+    }
+
+    /**
+     * Apply a delta transition from {@code fromVersion} to {@code toVersion} by applying {@code state}.
```
akhaku: minor: javadoc seems out of date
Also, is there a way to clarify here that the `writeState` is the "final" state?

hollow-test/src/main/java/com/netflix/hollow/test/HollowWriteStateEngineBuilder.java
```diff
@@ -53,6 +54,17 @@ public HollowWriteStateEngineBuilder(Collection<Class<?>> types) {
...
+     * Restore type states from the provided read state engine
+     */
+    public HollowWriteStateEngineBuilder restoreFrom(HollowReadStateEngine readStateEngine) {
```
akhaku: minor: does this really need to be exposed? if it's not needed perhaps we remove the method?

hollow-test/src/main/java/com/netflix/hollow/test/consumer/TestHollowConsumer.java
```diff
@@ -89,10 +95,38 @@ protected TestHollowConsumer(Builder builder) {
...
+     */
+    public void applyDelta(long toVersion, HollowWriteStateEngine writeState) throws IOException {
+        addDelta(getCurrentVersionId(), toVersion, writeState);
+        triggerRefreshTo(toVersion);
+    }
+
+    public TestHollowConsumer addSnapshot(long version, HollowWriteStateEngine state) throws IOException {
+        // if consumer has state then restore it
+        if (getStateEngine() != null) {
+            HollowWriteStateEngine snapshotState = HollowWriteStateCreator.createWithSchemas(getStateEngine().getSchemas());
```
akhaku, line 118: nice

hollow-test/src/main/java/com/netflix/hollow/test/consumer/TestHollowConsumer.java
```diff
@@ -101,6 +135,49 @@ public TestHollowConsumer addSnapshot(long version, HollowWriteStateEngine state
         return this;
     }
 
+    public TestHollowConsumer addDelta(long fromVersion, long toVersion, HollowWriteStateEngine state)
+            throws IOException {
+
+        if (getStateEngine() == null) {
+            throw new UnsupportedOperationException("Delta can not be applied without first applying a snapshot");
```
akhaku, line 142: Nice, although should we also say something like "you will probably want to call addSnapshot first"?

hollow-test/src/test/java/com/netflix/hollow/test/consumer/TestHollowConsumerTest.java
```diff
@@ -98,4 +108,112 @@ public void testAddSnapshot_afterUpdate() throws Exception {
...
+                .build();
+
+        HollowWriteStateEngine snapshotWriteStateEngine = new HollowWriteStateEngineBuilder().build();
+        consumer.addSnapshot(snapshotVersion, snapshotWriteStateEngine);
+        assertEquals("Should be no version", VERSION_NONE, consumer.getCurrentVersionId());
+        consumer.triggerRefreshTo(snapshotVersion);
+        assertEquals("Should be at snapshot version", snapshotVersion, consumer.getCurrentVersionId());
+
+        long deltaToVersion = 2L;
+        HollowWriteStateEngine deltaWriteStateEngine = new HollowWriteStateEngineBuilder().build();
```
akhaku: calling them snapshot and delta write state engine is confusing IMO, since they're really "complete" write state engines

akhaku (approved):
Looks great, nice!

hollow-test/src/main/java/com/netflix/hollow/test/consumer/TestHollowConsumer.java
Sunjeet: thanks, fixed these. Missing feature in javadocs https://stackoverflow.com/questions/1667212/how-to-add-reference-to-a-method-parameter-in-javadoc

hollow-test/src/main/java/com/netflix/hollow/test/consumer/TestHollowConsumer.java
Sunjeet: Leaving this as is, because there are addSnapshot and applySnapshot and maybe other ways to get there.

PR #532, opened 2021-05-04
Fix NPE in StaleHollowReferenceDetector

nayanika-u:


hollow/src/main/java/com/netflix/hollow/api/client/StaleHollowReferenceDetector.java
```diff
@@ -222,16 +222,17 @@ private boolean shouldDetach() {
...
 
-            api.detachCaches();
+                if (dataAccess instanceof HollowProxyDataAccess)
+                    ((HollowProxyDataAccess) dataAccess).disableDataAccess();
+                else if (dataAccess instanceof HollowReadStateEngine)
+                    ((HollowReadStateEngine) dataAccess).invalidate();
 
-            detached = true;
+                api.detachCaches();
+                detached = true;
```
Sunjeet: should `detach = true` be outside the if condition?

nayanika-u: ack. If inside, this would lead to redundant calls to shouldDetach().

PR #539, opened 2021-06-02
Enforce primary producer during publish and announce

Sunjeet:


hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -807,6 +807,13 @@ private void announce(ProducerListeners listeners, HollowProducer.ReadState read
         if (announcer != null) {
```
nayanika-u, line 828: Shouldn't this added if clause at L810 be checked before we even fire the announcement listeners at L808?

Sunjeet: If we moved the validation up, then between the validation happening and the announcement happening the leader lock could have been relinquished and a different leader might step in while this producer's announcement will be allowed to succeed based on a stale result.

But your comment helped me realize that the current implementation doesn't protect us entirely from the race condition, it just shortens the time window (now shortened to execution getting from line 810 to when the announcement is consistently available for reads) during which if the leader lock is released and a different instance acquires primary status then we'd have a bifurcated delta chain. More work is required.

Sunjeet: Added a revision that holds the leader lock for the time period between this check + until announcement is made. This should protect from the case where `producer.enablePrimaryProducer(false)` comes in after the primary check passes but before the announcement is made.

Sunjeet: Please take another look, and also inviting thoughts from @akhaku

akhaku: I think we additionally want to check before uploading blobs since that could cause a delta chain overwrite

hollow/src/main/java/com/netflix/hollow/api/producer/enforcer/AbstractSingleProducerEnforcer.java
```diff
@@ -27,6 +29,8 @@
     private boolean wasPrimary = false;
     private final Logger logger = Logger.getLogger(AbstractSingleProducerEnforcer.class.getName());
 
+    private volatile Lock lock = new ReentrantLock();
```
akhaku: I think this would be better as `final` and not `volatile` since it's never reassigned?

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -807,7 +807,21 @@ private void announce(ProducerListeners listeners, HollowProducer.ReadState read
...
+                // don't allow producer to enable/disable between validating primary status and then announcing
+                singleProducerEnforcer.lock();
+                try {
+                    if (!singleProducerEnforcer.isPrimary()) {
+                        // This situation can arise when the producer is asked to relinquish primary status mid-cycle.
+                        // If this non-primary producer is allowed to announce then a different producer instance could
+                        // have been running concurrently as primary and that would break the delta chain.
+                        log.log(Level.INFO,
+                                "Fail the announcement because current producer is not primary (aka leader)");
+                        throw new IllegalStateException("Announcement failed primary (aka leader) check");
```
akhaku, line 840: Oh good idea to fail the cycle: I was initially concerned about how we would handle cycle refresh listeners if a cycle was "canceled" but this is a clean way to handle that.

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -807,7 +807,21 @@ private void announce(ProducerListeners listeners, HollowProducer.ReadState read
...
-                announcer.announce(readState.getVersion(), readState.getStateEngine().getHeaderTags());
+                // don't allow producer to enable/disable between validating primary status and then announcing
+                singleProducerEnforcer.lock();
+                try {
+                    if (!singleProducerEnforcer.isPrimary()) {
+                        // This situation can arise when the producer is asked to relinquish primary status mid-cycle.
+                        // If this non-primary producer is allowed to announce then a different producer instance could
+                        // have been running concurrently as primary and that would break the delta chain.
+                        log.log(Level.INFO,
+                                "Fail the announcement because current producer is not primary (aka leader)");
```
akhaku, line 839: I _think_ internally we tend to say "lease holder" rather than "leader", not sure which one is more precise though

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -807,7 +807,21 @@ private void announce(ProducerListeners listeners, HollowProducer.ReadState read
         if (announcer != null) {
             Status.StageWithStateBuilder status = listeners.fireAnnouncementStart(readState);
             try {
-                announcer.announce(readState.getVersion(), readState.getStateEngine().getHeaderTags());
+                // don't allow producer to enable/disable between validating primary status and then announcing
+                singleProducerEnforcer.lock();
```
akhaku, line 832: nice, I think this lock works pretty well 👍

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
akhaku: I should probably elaborate so it's more clear: at least in our internal implementation of the delta chain, it's the act of uploading the delta and reverse delta blob that creates the delta chain (since we store to_version and from_version), not the announcement itself. The announcement only decides the "latest" version, the metadata on the blobs decides the delta chain, so if we wait until the announcement to check the "primary producer" status we're probably too late.
We should probably keep the check here too though, for completeness. We could extend this lock to start before uploading blobs? Or perhaps wrap blob publishing in a similar-looking check `¯\_(ツ)_/¯`

nayanika-u: The lock part looks good. I am just wondering why do we even care about firing the listeners if the announcement shouldn't be allowed to go through in case of a non-primary producer. That is why I was suggesting the check to be placed before we obtain the status of firing the announcement listeners.

Sunjeet: Sounds good @akhaku, I think a similar check at blob publish would be better as compared to extending the duration of the lock so that if someone's trying to disable primary they're not waiting forever (publish, listeners, etc.)

@nayanika-u this way the failure handling is a bit more clean. We call announcement start listeners, but we fail before announcing and we fire announcement complete with a failure status. It can be thought of as "the announcement failed because producer is not primary."

Failing before calling announce start would mean that we failed after validate but before announce. We'd need a way to represent to the user why we didn't get to announce stage i.e. we'd need a separate stage or maybe the check could be a part of the validation stage? But as I noted earlier its good to reduce the time frame between when this check is done and when the announcement occurs (so for eg. a long running onAnnouncementStart listener wouldn't prolog that interval).

hollow/src/main/java/com/netflix/hollow/api/producer/enforcer/AbstractSingleProducerEnforcer.java
Sunjeet: Indeed

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
Sunjeet: Leader, being more generic, seems more appropriate here from the point of view of the user

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
Sunjeet: Latest revision adds the check to delta and reverse delta publishing

akhaku (approved):

nayanika-u (approved):

PR #540, opened 2021-06-07
UniqueKeyIndex performance improvement

nayanika-u:


hollow/src/main/java/com/netflix/hollow/api/consumer/index/UniqueKeyIndex.java
```diff
@@ -157,13 +158,32 @@ public T apply(Q key) {
      * @return the unique object
      */
     public T findMatch(Q key) {
-        Object[] keyArray = matchFields.stream().map(mf -> mf.extract(key)).toArray();
 
-        int ordinal = hpki.getMatchingOrdinal(keyArray);
-        if (ordinal == -1) {
-            return null;
+        Object[] keyArray = new Object[matchFields.size()];
```
Sunjeet: Lets make this an instance variable

hollow/src/main/java/com/netflix/hollow/core/memory/encoding/HashCodes.java
```diff
@@ -162,7 +162,7 @@ public static int hashLong(long key) {
     }
 
     public static int hashInt(int key) {
-        key = ~key + (key << 15);
```
Sunjeet: nit: undo this

nayanika-u: done

hollow/src/main/java/com/netflix/hollow/api/consumer/index/UniqueKeyIndex.java
nayanika-u: done

Sunjeet: Thanks, we can also move the initialization up into the constructor so that the array is initialized once instead of per call to findMatch

nayanika-u: sorry. missed this. For some reason, I thought matchField size would vary for every findMatch call so I let it stay here.

Sunjeet (approved):

PR #541, opened 2021-06-12
Helpful log msg when NPE in HollowDiffMatcher

nayanika-u:


hollow/src/main/java/com/netflix/hollow/core/index/HollowPrimaryKeyIndex.java
```diff
@@ -276,13 +277,18 @@ public int getMatchingOrdinal(Object key1, Object key2, Object key3) {
      */
     public int getMatchingOrdinal(Object... keys) {
         PrimaryKeyIndexHashTable hashTable = hashTableVolatile;
-        if(fieldPathIndexes.length != keys.length || hashTable.bitsPerElement == 0)
+
+        // for special case of keys containing one null element
+        if (keys == null || (keys != null && fieldPathIndexes.length != keys.length) || hashTable.bitsPerElement == 0)
```
nayanika-u: This will throw NPE for cases like getMatchingOrdinal(null). 

getMatchingOrdinal((Object) null), getMatchingOrdinal(null, null), getMatchingOrdinal(new Object[]{null, null}) and so on will pass this check.

Sunjeet: The `keys != null` check is redundant, it can be dropped

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrimaryKeyIndex.java
```diff
@@ -276,13 +277,18 @@ public int getMatchingOrdinal(Object key1, Object key2, Object key3) {
...
             return -1;
 
         int hashCode = 0;
-        for(int i=0;i<keys.length;i++)
-            hashCode ^= keyHashCode(keys[i], i);
-
+        for (int i = 0;i < keys.length;i++) {
+            // null values can percolate from upstream in the method's Object array
+            if (Objects.nonNull(keys[i])) {
```
Sunjeet: Nit: use `!= null` here? Might read better and avoid a tiny performance overhead

Sunjeet: Do we need to add the null check to the non vararg flavors of `getMatchingOrdinal`?

Sunjeet: IIUC a behavior change we're introducing with this change that a component of the primary key is now nullable where earlier it wasn't. I'm not sure what the implication of the computed hashCode would be. Is this something we've already looked into, or do we need to rethink whether we want to support null for key[i] or fail gracefully instead?

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrimaryKeyIndex.java
nayanika-u: Done.

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrimaryKeyIndex.java
nayanika-u: As discussed, this code change might lead to incorrect hashCode computation for cases where it should have just failed (cases with PK passing null field values) and lead to incorrect results in the lookup.

nayanika-u:
Merging this after checking with David.

hollow/src/main/java/com/netflix/hollow/tools/diff/HollowDiffMatcher.java
```diff
@@ -21,7 +21,10 @@
 import com.netflix.hollow.core.read.engine.object.HollowObjectTypeReadState;
 import com.netflix.hollow.core.util.IntList;
 import com.netflix.hollow.core.util.LongList;
+import com.sun.org.slf4j.internal.Logger;
```
Sunjeet, line 24: Note: double check import

hollow/src/main/java/com/netflix/hollow/tools/diff/HollowDiffMatcher.java
```diff
@@ -91,7 +95,14 @@ public void calculateMatches() {
...
         while(candidateToMatchOrdinal != -1) {
             Object key[] = toIdx.getRecordKey(candidateToMatchOrdinal);
-            int matchedOrdinal = fromIdx.getMatchingOrdinal(key);
+            int matchedOrdinal = -1;
+            try {
+                matchedOrdinal = fromIdx.getMatchingOrdinal(key);
+            } catch(NullPointerException ex) {
+                LOGGER.error("Null values found for type " + toTypeState.getSchema().getName()
+                        + " with key field values " + Arrays.asList(key) + " at ordinal : " + candidateToMatchOrdinal
+                        + " with stacktrace : " + ex.getStackTrace());
```
Sunjeet, line 104: NOTE: double check that this is logging the stack trace.

nayanika-u: ack. have removed it from printing. As discussed, re throwing this as RuntimeException to preserve existing behavior for Hollow Tooling

hollow/src/main/java/com/netflix/hollow/tools/diff/HollowDiffMatcher.java
nayanika-u: changed to java.util.Logging

hollow/src/main/java/com/netflix/hollow/tools/diff/HollowDiffMatcher.java
```diff
@@ -91,7 +95,14 @@ public void calculateMatches() {
         int candidateToMatchOrdinal = toPopulatedOrdinals.nextSetBit(0);
         while(candidateToMatchOrdinal != -1) {
             Object key[] = toIdx.getRecordKey(candidateToMatchOrdinal);
-            int matchedOrdinal = fromIdx.getMatchingOrdinal(key);
+            int matchedOrdinal = -1;
```
akhaku, line 98: btw might be good to use `HollowConstants.ORDINAL_NONE` here

nayanika-u: ack. Kept it -1 like other ordinal checks in this class.

PR #543, opened 2021-06-22
Fix error logs for null values encountered in HollowDiffMatcher

nayanika-u:


hollow/src/main/java/com/netflix/hollow/tools/diff/HollowDiffMatcher.java
```diff
@@ -99,9 +99,11 @@ public void calculateMatches() {
             try {
                 matchedOrdinal = fromIdx.getMatchingOrdinal(key);
             } catch(NullPointerException ex) {
-                LOGGER.error("Null values found for type " + toTypeState.getSchema().getName()
+                LOGGER.log(Level.SEVERE, "Null values found for type " + toTypeState.getSchema().getName()
```
akhaku: So I actually agree with what you were saying earlier w.r.t there being no reason to log _and_ throw. Since we're going with rethrowing, how about we remove the log message entirely? As long as the message in the thrown RuntimeException has the details, we don't need to log here since the details will be included whenever the RuntimeException is logged or bubbles up.

hollow/src/main/java/com/netflix/hollow/tools/diff/HollowDiffMatcher.java
```diff
@@ -99,9 +99,11 @@ public void calculateMatches() {
...
+                        + " with key field values " + Arrays.asList(key) + " at ordinal : " + candidateToMatchOrdinal);
+                throw new RuntimeException("Error fetching matching ordinal for null type " + toTypeState.getSchema().getName()
                         + " with key field values " + Arrays.asList(key) + " at ordinal : " + candidateToMatchOrdinal
-                        + " with stacktrace : " + ex.getStackTrace());
+                        + "with stack trace " + ex.getCause());
```
akhaku: I'm not sure if `ex.getCause()` will do what you want - I think you probably want to pass `ex` _as_ the cause (as the second argument to `new RuntimeException`)

akhaku (approved):

hollow/src/main/java/com/netflix/hollow/tools/diff/HollowDiffMatcher.java
nayanika-u: fixed

hollow/src/main/java/com/netflix/hollow/tools/diff/HollowDiffMatcher.java
nayanika-u: Sounds good.

PR #544, opened 2021-06-28
Fix: Track current and previous cycle start times to set the right value in blob header

Sunjeet:
Cycle start time is set by the producer in the blob headers, using local time in milliseconds, corresponding to the write state creation. Consumers rely on this value to compute a metric for measuring data staleness- time elapsed since cycle started on the producer.

The earlier implementation always injected the start time of the current runCycle into the blob header. This was problematic for reverse deltas since the reverse delta from v1 to v0 is created in runcycle corresponding to v1 and thus contained the start time for v1 runcycle. Instead, the reverse detla header should contain the start time corresponding to v0 cycle, so that consumers can accurately compute staleness when they transition to v0 state by following reverse delta from v1 to v0.

In the reworked implementation, previous cycle start time is tracked in the write state engine and thats set on the reverse delta.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/util/HollowWriteStateCreator.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/util/HollowWriteStateCreator.java
@@ -193,10 +193,16 @@ public class HollowWriteStateCreator {
         
         writeEngine.addHeaderTags(readEngine.getHeaderTags());
         writeEngine.overrideNextStateRandomizedTag(readEngine.getCurrentRandomizedTag());
-        if (readEngine.getHeaderTag(HEADER_TAG_METRIC_CYCLE_START) != null) {
-            writeEngine.overrideCycleStartTs(Long.valueOf(
-                    readEngine.getHeaderTag(HEADER_TAG_METRIC_CYCLE_START)));
+
+        // override current cycle start time
+        writeEngine.overrideCycleStartTs(System.currentTimeMillis());
+
+        // set previousCycleStartTs to the restored read state's cycle time
+        String readStateCycleStartTs = readEngine.getHeaderTag(HEADER_TAG_METRIC_CYCLE_START);
+        if (readStateCycleStartTs != null) {
+            writeEngine.overridePreviousCycleStartTs(Long.valueOf(readStateCycleStartTs));
         }
+
         writeEngine.prepareForWrite();
     }
 }
--- a/hollow/src/main/java/com/netflix/hollow/core/util/HollowWriteStateCreator.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/util/HollowWriteStateCreator.java
@@ -16,6 +16,8 @@
  */
 package com.netflix.hollow.core.util;
 
+import static com.netflix.hollow.core.HollowStateEngine.HEADER_TAG_METRIC_CYCLE_START;
+
 import com.netflix.hollow.core.read.engine.HollowReadStateEngine;
 import com.netflix.hollow.core.read.engine.HollowTypeReadState;
 import com.netflix.hollow.core.read.engine.PopulatedOrdinalListener;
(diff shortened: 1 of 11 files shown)


hollow/src/main/java/com/netflix/hollow/core/write/HollowBlobWriter.java
```diff
@@ -280,13 +280,17 @@ private void writeHeaders(DataOutputStream os, ProducerOptionalBlobPartConfig.Op
 
         /// write main header
         HollowBlobHeader header = new HollowBlobHeader();
-        header.setHeaderTags(stateEngine.getHeaderTags());
+        header.getHeaderTags().putAll(stateEngine.getHeaderTags());
```
Sunjeet: subtle change here to make a deep copy

hollow/src/main/java/com/netflix/hollow/core/util/HollowWriteStateCreator.java
```diff
@@ -191,6 +193,10 @@ public void run() {
         
         writeEngine.addHeaderTags(readEngine.getHeaderTags());
         writeEngine.overrideNextStateRandomizedTag(readEngine.getCurrentRandomizedTag());
+        if (readEngine.getHeaderTag(HEADER_TAG_METRIC_CYCLE_START) != null) {
```
nayanika-u: Not very clear on why we use the header from read state and use that value to override the write state. Why is this needed?

hollow/src/main/java/com/netflix/hollow/core/write/HollowBlobWriter.java
nayanika-u: +1 since we are updating header tags in different places.

hollow/src/main/java/com/netflix/hollow/core/write/HollowBlobWriter.java
```diff
@@ -280,13 +280,17 @@ private void writeHeaders(DataOutputStream os, ProducerOptionalBlobPartConfig.Op
...
         if(isReverseDelta) {
             header.setOriginRandomizedTag(stateEngine.getNextStateRandomizedTag());
             header.setDestinationRandomizedTag(stateEngine.getPreviousStateRandomizedTag());
+            // cycle start time on reverse delta from v1 to v0 is the cycle start time of v0
+            stateEngine.getPreviousCycleStartTs().ifPresent(t -> header.setCycleStartTsTag(t));
         } else {
             header.setOriginRandomizedTag(stateEngine.getPreviousStateRandomizedTag());
             header.setDestinationRandomizedTag(stateEngine.getNextStateRandomizedTag());
+            // cycle start time on delta from v0 to v1 is the cycle start time of v1
```
nayanika-u, line 292: For the comment at 292 : Shouldn't this also be cycle start time of v0?

hollow/src/main/java/com/netflix/hollow/core/write/HollowWriteStateEngine.java
```diff
@@ -168,6 +171,11 @@ public void run() {
 
         previousStateRandomizedTag = readStateEngine.getCurrentRandomizedTag();
         nextStateRandomizedTag = mintNewRandomizedStateTag();
+        if (readStateEngine.getHeaderTag(HEADER_TAG_METRIC_CYCLE_START) != null) {
+            previousCycleStartTs = OptionalLong.of(Long.valueOf(
```
nayanika-u, line 175: I understand that we keep track of previous cycle TS. In restoreFrom(), If a producer is restarted then why do we need to capture the previous cycle TS? Isn't this value redundant in this scenario?

nayanika-u: Okay, this value is assigned during reverse delta computation..

hollow/src/main/java/com/netflix/hollow/core/write/HollowBlobWriter.java
nayanika-u: Are we validating the size of header tags map in the UT?

nayanika-u (approved):

hollow/src/main/java/com/netflix/hollow/core/write/HollowBlobWriter.java
nayanika-u: just bumping this one^

Sunjeet:
Abandoned in favor of [this PR](https://github.com/Netflix/hollow/pull/564) which fixes how header tags are populated on reverse deltas. The current implementation sets the cycle start time as a header tag within runCycle, so that fix resolves this

PR #545, opened 2021-06-30
Fix decimalformat

davesu:
Fix DecimalFormat not being ThreadSafe and could cause NPE

### DIFF
--- a/hollow-ui-tools/src/main/java/com/netflix/hollow/ui/HollowDiffUtil.java
+++ b/hollow-ui-tools/src/main/java/com/netflix/hollow/ui/HollowDiffUtil.java
@@ -3,7 +3,7 @@ package com.netflix.hollow.ui;
 import java.text.DecimalFormat;
 
 public class HollowDiffUtil {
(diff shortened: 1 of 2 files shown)


hollow-ui-tools/src/main/java/com/netflix/hollow/ui/HollowDiffUtil.java
```diff
@@ -13,6 +12,7 @@ public static String formatBytes(long sizeInBytes) {
         sizeInBytes = Math.abs(sizeInBytes);
 
         int digitGroups = (int) (Math.log10(sizeInBytes)/Math.log10(1024));
+        DecimalFormat BYTE_FORMATTER = new DecimalFormat("#,##0.##");
```
Sunjeet: Ship it! 
nit: variable name casing.

Sunjeet (approved):

hollow-ui-tools/src/main/java/com/netflix/hollow/ui/HollowDiffUtil.java
davesu: Good point, it used to be static final but it is renamed to just "formatter" now.

PR #546, opened 2021-08-06
Throw a more specific NotAPrimaryProducerException for failures due t…

nayanika-u:
…o producer failing primary producer check. Under majority of cases, primary producer check failures are transient and do not require any action by App teams. This will be helpful in setting more relevant alerts.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
@@ -607,7 +607,7 @@ abstract class AbstractHollowProducer {
             if (!blob.type.equals(HollowProducer.Blob.Type.SNAPSHOT)) {
                 // Don't allow producer to enable/disable between validating primary status and publishing for delta
                 // or reverse delta blobs. No need to check for primary status when publishing snapshots because a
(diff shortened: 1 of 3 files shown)


hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -837,7 +837,7 @@ private void announce(ProducerListeners listeners, HollowProducer.ReadState read
                         // have been running concurrently as primary and that would break the delta chain.
                         log.log(Level.INFO,
                                 "Fail the announcement because current producer is not primary (aka leader)");
-                        throw new IllegalStateException("Announcement failed primary (aka leader) check");
+                        throw new HollowProducer.NotAPrimaryProducerMidCycleException("Announcement failed primary (aka leader) check");
```
Sunjeet: nit: since it's scoped to HollowProducer, we could make the name a bit shorter-
`HollowProducer.NotPrimaryMidCycleException`
or
`HollowProducer.LostPrimaryStatusMidCycleException`

nayanika-u: changed it to HollowProducer.NotPrimaryMidCycleException.

Sunjeet (approved):

PR #555, opened 2021-12-02
Add support for Publishing Header alongside Blob (Snapshot/Delta)

mbrilnetflix:


hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -573,7 +576,7 @@ void publish(ProducerListeners listeners, long toVersion, Artifacts artifacts) t
                 artifacts.markSnapshotPublishComplete();
                 numStatesUntilNextSnapshot = numStatesBetweenSnapshots;
             }
-
+            publishHeader(listeners, artifacts.header);
```
Sunjeet: minor: maybe better to publish before snapshot etc. so that we always have the header artifact corresponding to a snapshot version

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -755,7 +796,17 @@ private ReadStateHelper checkIntegrity(
             listeners.fireIntegrityCheckComplete(status);
         }
     }
-    
+
+    private void verifyHeader(HollowProducer.HeaderBlob header) {
```
Sunjeet: minor: seems like test code

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -506,6 +506,15 @@ public void removeListener(HollowProducerEventListener listener) {
...
         HollowProducer.Blob openSnapshot(long version);
 
+        /**
+         * Returns a blob with which a {@code HollowProducer} will write a header for the version specified.
+         * <p>
+         * The producer will pass the returned blob back to this publisher when calling {@link Publisher#publish(HollowProducer.HeaderBlob)}.
+         * @param version the blob version
+         * @return a {@link HollowProducer.HeaderBlob} representing a header for the {@code version}
+         */
+        HollowProducer.HeaderBlob openHeader(long version);
```
Sunjeet, line 516: Re: backwards compatibility, I think it should fine to not provide a default implementation here, we can bump up the major version on the release just in case of external implementations of this interface

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -578,6 +587,46 @@ public InputStream decompress(InputStream is) {
...
+         * @param headerBlob the header blob to publish
+         */
+        void publish(HollowProducer.HeaderBlob headerBlob);
+    }
+
+    public static abstract class HeaderBlob {
+        protected final long toVersion;
+        protected final long fromVersion;
+
+        protected HeaderBlob(long fromVersion, long toVersion) {
```
Sunjeet: Something seems off- should `HeaderBlob` have both from and to versions?

Sunjeet:
I think we can simplify the implementation by not staging the header blob given it is pretty small in size. So-
1. `HollowBlobWriter` could have a method that returns a `HollowBlobHeader` given the writestateengine that it is already initialized with. Maybe this new method can be reused in how `HollowBlobWriter` currently writes the headers.
2. We could call this method in `AbstractHollowProducer`'s `publish` method to get back a `HollowBlobHeader` and pass that to a new `publishHeaders` method on the `HollowProducer.Publisher` interface. Note this wouldn't do any of the listener notifications stuff that you implemented for the headers published etc.
3. Both `HollowProducer.Blob` and `HollowBlobHeader` classes could implement a new `PublishArtifact` interface and we would update everywhere (listener events, metrics, etc.) that handled `HollowProducer.Blob` to also handle `HollowBlobHeader`. 
I think we could also stop at (2) and add (3) in future if we think there is value is notifying on header publish.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -578,6 +587,46 @@ public InputStream decompress(InputStream is) {
...
         void publish(HollowProducer.Blob blob);
+
+        /**
+         * Publish the headerBlob specified to this publisher's header blobstore.
+         * <p>
+         * {@link BlobStager#openHeader(long)} on this publisher.
+         *
+         * @param headerBlob the header blob to publish
+         */
+        void publish(HollowProducer.HeaderBlob headerBlob);
```
Sunjeet: We should provide a default implementation for backwards compatibility

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
mbrilnetflix: Actually followed common practices with snapshot. 
but, yeah I can see why this is confusing... I'll fix it.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
mbrilnetflix: ACK

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
mbrilnetflix: ACK

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
mbrilnetflix: ACK

hollow/src/main/java/com/netflix/hollow/api/producer/fs/HollowFilesystemBlobStager.java
```diff
@@ -110,6 +111,63 @@ public HollowFilesystemBlobStager(File stagingPath, BlobCompressor compressor) {
         return new FilesystemBlob(fromVersion, toVersion, REVERSE_DELTA, stagingPath, compressor, optionalPartConfig);
     }
 
+    @Override
+    public HollowProducer.HeaderBlob openHeader(long version) {
+        return new FilesystemHeaderBlob(version, stagingPath, compressor);
+    }
+
+    public static class FilesystemHeaderBlob extends HeaderBlob {
```
Sunjeet, line 119: From a very superficial view, it seems that a bunch of duplicate code could be avoided if `HeaderBlob` and `HollowProducer.Blob` implemented a common interface (say `PublishArtifact`). Is that something you've already considered? So handling header vs. snapshot et al could look something like-
```
if (publishArtifact instanceof HeaderBlob) {
    // do somethign with header
} else if (publishArtifact instanceof HollowProducer.Blob) {
    HollowProducer.Blob blob = (HollowProducer.Blob) publishArtifact;
    switch (blob.ype) {
        case SNAPSHOT: // do something with snapshot; break;
        ...
    }
}
```

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -578,6 +587,41 @@ public InputStream decompress(InputStream is) {
...
+
+        public abstract void cleanup();
+        protected abstract void write(HollowBlobWriter blobWriter) throws IOException;
+        public abstract InputStream newInputStream() throws IOException;
+
+        public long getVersion() {
+            return version;
+        }
+
+        @Deprecated
```
Sunjeet: minor: if this is already Deprecated we probably dont need this at all

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -555,6 +558,7 @@ void publish(ProducerListeners listeners, long toVersion, Artifacts artifacts) t
                 artifacts.reverseDelta = stageBlob(listeners,
                         blobStager.openReverseDelta(toVersion, readStates.current().getVersion()));
 
+                publishHeaderBlob(artifacts.header);
```
Sunjeet: nit: this call can be moved up before the if-else, since it is duplicated in both if and else

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -585,6 +589,13 @@ void publish(ProducerListeners listeners, long toVersion, Artifacts artifacts) t
         }
     }
 
+    private HollowProducer.HeaderBlob publishHeader(HollowProducer.HeaderBlob headerBlob)
```
Sunjeet: nit: dont forget to drop unused method before merging

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
Sunjeet: nit: dont forget to drop unused method before merging

hollow/src/main/java/com/netflix/hollow/api/producer/fs/HollowFilesystemBlobStager.java
mbrilnetflix: Did some refactoring following our discussion

hollow/src/main/java/com/netflix/hollow/api/client/HollowBlobRetriever.java
```diff
@@ -32,18 +33,23 @@
...
      * @return a reverse delta transition which can be applied to the currentVersion
      */
-    public HollowBlob retrieveReverseDeltaBlob(long currentVersion);
+    HollowBlob retrieveReverseDeltaBlob(long currentVersion);
 
+    /**
+     * @param desiredVersion the desired version
+     * @return the header for the state with an identifier equal to or less than the desired version
+     */
+    HollowBlobHeader retrieveHeaderBlob(long desiredVersion);
```
Sunjeet: Add a default implementation here for backwards compatibility

hollow/src/main/java/com/netflix/hollow/api/client/HollowClientConsumerBridge.java
```diff
@@ -29,7 +30,21 @@
...
+            @Override
+            public HollowConsumer.HeaderBlob retrieveHeaderBlob(long desiredVersion) {
+                final HollowBlobHeader blob = blobRetriever.retrieveHeaderBlob(desiredVersion);
+                if(blob == null)
+                    return null;
+
+                return new HollowConsumer.HeaderBlob(desiredVersion) {
+                    @Override
+                    public InputStream getInputStream() throws IOException {
+                        return null;
```
Sunjeet: maybe throw UnsupportedOperationException instead

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
```diff
@@ -471,6 +473,42 @@ public HollowConsumerMetrics getMetrics() {
             return null;
         }
 
+        HollowConsumer.HeaderBlob retrieveHeaderBlob(long desiredVersion);
+    }
+
+    protected static abstract class AbstractVersionedBlob {
+        protected final long fromVersion;
```
Sunjeet: Since both fromVersion and toVersion are specific to HollowProducer.Blob and HollowCOnsumer.Blob; and only toVersion is applicable to HollowProducer.HeaderBlob and HollowConsumer.HeaderBlob => I'd move the state fromVersion and toVersion into HollowProducer.Blob and HollowConsumer.Blob; and toVersion to HollowProducer.HederBlob and HollowConsumer.HeaderBlob.

Sunjeet: Once we do that, an interface would be better suited here instead of abstract class

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -574,13 +583,41 @@ public InputStream decompress(InputStream is) {
...
+         * {@link BlobStager#openHeader(long)} on this publisher.
          *
-         * @param blob the blob to publish
+         * @param publishArtifact the blob to publish
          */
-        void publish(HollowProducer.Blob blob);
+        void publish(HollowProducer.AbstractPublishArtifact publishArtifact);
+    }
+
+    public static abstract class AbstractPublishArtifact {
```
Sunjeet: An interface would be better suited here

hollow/src/main/java/com/netflix/hollow/api/producer/Status.java
```diff
@@ -146,6 +146,15 @@ PublishBuilder blob(HollowProducer.Blob blob) {
         }
     }
 
+    static final class PublishHeaderBuilder extends AbstractStatusBuilder<PublishHeaderBuilder> {
```
Sunjeet: minor: unused

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -574,13 +583,41 @@ public InputStream decompress(InputStream is) {
          * It is guaranteed that {@code blob} was created by calling one of
          * {@link BlobStager#openSnapshot(long)}, {@link BlobStager#openDelta(long, long)}, or
          * {@link BlobStager#openReverseDelta(long, long)} on this publisher.
+         * {@link BlobStager#openHeader(long)} on this publisher.
          *
-         * @param blob the blob to publish
+         * @param publishArtifact the blob to publish
          */
-        void publish(HollowProducer.Blob blob);
```
Sunjeet, line 580: We need to keep `void publish(HollowProducer.Blob blob);` around for backwards compatibility

hollow/src/main/java/com/netflix/hollow/api/producer/fs/HollowFilesystemPublisher.java
```diff
@@ -44,7 +44,30 @@ public HollowFilesystemPublisher(Path blobStorePath) {
...
+        }
+    }
+
+    private void publishHeader(HollowProducer.HeaderBlob headerBlob) {
+        Path destination = blobStorePath.resolve(String.format("header-%d", headerBlob.getVersion()));
+        try (
+                InputStream is = headerBlob.newInputStream();
+                OutputStream os = Files.newOutputStream(destination)
+        ) {
+            byte buf[] = new byte[4096];
```
Sunjeet, line 66: Minor: the buffered copy code is now at 3 places in this file, consider moving it into a different method

hollow/src/main/java/com/netflix/hollow/api/client/HollowBlobRetriever.java
mbrilnetflix: This is deprecated but with only a single impl... 
I've added the default, just in case.

hollow/src/main/java/com/netflix/hollow/api/client/HollowClientConsumerBridge.java
mbrilnetflix: Ack

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
mbrilnetflix: ack

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
mbrilnetflix: Ack

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
mbrilnetflix: Done, marked it as Deprecated as well, with a link to the *new* publish in docs

hollow/src/main/java/com/netflix/hollow/api/producer/Status.java
mbrilnetflix: Correct, left over from previous refactoring

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
mbrilnetflix: Done

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
mbrilnetflix: Done

hollow/src/main/java/com/netflix/hollow/api/client/HollowBlobRetriever.java
```diff
@@ -32,18 +33,25 @@
...
     /**
      * @param currentVersion the current version
      * @return a reverse delta transition which can be applied to the currentVersion
      */
-    public HollowBlob retrieveReverseDeltaBlob(long currentVersion);
+    HollowBlob retrieveReverseDeltaBlob(long currentVersion);
 
+    /**
+     * @param desiredVersion the desired version
+     * @return the header for the state with an identifier equal to or less than the desired version
```
Sunjeet: Hmm, wonder if this should be a strict match on version similar to deltas

hollow/src/main/java/com/netflix/hollow/api/client/HollowClientConsumerBridge.java
```diff
@@ -29,7 +30,21 @@
...
+            @Override
+            public HollowConsumer.HeaderBlob retrieveHeaderBlob(long desiredVersion) {
+                final HollowBlobHeader blob = blobRetriever.retrieveHeaderBlob(desiredVersion);
+                if(blob == null)
+                    return null;
+
+                return new HollowConsumer.HeaderBlob(desiredVersion) {
+                    @Override
+                    public InputStream getInputStream() throws IOException {
+                        throw new UnsupportedOperationException();
```
Sunjeet: Is this just not implemented yet? Should we be throwing NotImplementedException?

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -577,10 +587,50 @@ public InputStream decompress(InputStream is) {
...
          */
+        @Deprecated
         void publish(HollowProducer.Blob blob);
+
+        /**
+         * Publish the blob specified to this publisher's blobstore.
+         * <p>
+         * It is guaranteed that {@code blob} was created by calling one of
+         * {@link BlobStager#openSnapshot(long)}, {@link BlobStager#openDelta(long, long)}, or
+         * {@link BlobStager#openReverseDelta(long, long)} on this publisher.
```
Sunjeet: nit: typo

hollow/src/test/java/com/netflix/hollow/api/producer/validation/HollowProducerValidationListenerTest.java
```diff
@@ -39,7 +39,15 @@
 
     @Before
     public void setup() {
-        publisher = blob -> {};
+        publisher = new Publisher() {
```
Sunjeet, line 42: nit: unused

hollow/src/main/java/com/netflix/hollow/api/client/HollowClientConsumerBridge.java
mbrilnetflix: This was added following your previous comment :)

mbrilnetflix: Will add comment as to why this is not supported

hollow-test/src/main/java/com/netflix/hollow/test/consumer/TestHeaderBlob.java
```diff
@@ -0,0 +1,24 @@
...
+
+    public TestHeaderBlob(long version, InputStream inputStream) {
+        super(version);
+        this.inputStream = inputStream;
+    }
+
+
+    @Override
+    public InputStream getInputStream() throws IOException {
+        return inputStream;
```
Sunjeet: Consider throwing unsupported operation or not implemented exception here

hollow/src/main/java/com/netflix/hollow/api/metrics/HollowProducerMetrics.java
```diff
@@ -34,6 +34,8 @@
     private int deltasFailed = 0;
     private int reverseDeltasCompleted = 0;
     private int reverseDeltasFailed = 0;
+    private int headersCompleted = 0;
```
Sunjeet: nit: we dont currently do anything with these, they can be added later if necessary

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -577,10 +587,50 @@ public InputStream decompress(InputStream is) {
...
+         * {@link BlobStager#openHeader(long)} on this publisher.
+         *
+         * @param publishArtifact the blob to publish
+         */
+        void publish(HollowProducer.PublishArtifact publishArtifact);
```
Sunjeet: This one would be good to double check for code references outside hollow, we may need a default implementation (apologies if we already ruled this out in an earlier PR iteration).

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
Sunjeet: We could also provide a default implementation for this in the interface-
```
@Deprecated
default void publish(HollowProducer.Blob blob) {
    PublishArtifact publishArtifact = (PublishArtifact) blob;
    publish(publishArtifact);
}
```
and drop the overridden method from the implementations within hollow

hollow/src/main/java/com/netflix/hollow/core/write/HollowBlobWriter.java
```diff
@@ -289,9 +302,12 @@ private void writeHeaders(DataOutputStream os, ProducerOptionalBlobPartConfig.Op
...
+        return new HollowBlobHeaderWrapper(header, schemasByPartName);
+    }
 
-        VarInt.writeVInt(os, mainSchemas.size());
+    private void writeHeaders(DataOutputStream os, ProducerOptionalBlobPartConfig.OptionalBlobPartOutputStreams partStreams, boolean isReverseDelta, HollowBlobHeaderWrapper hollowBlobHeaderWrapper) throws IOException {
+        headerWriter.writeHeader(hollowBlobHeaderWrapper.header, os);
+        VarInt.writeVInt(os, hollowBlobHeaderWrapper.header.getSchemas().size());
 
         if(partStreams != null) {
             /// write part headers
```
Sunjeet, line 313: It would be good to test out the optional parts case

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
```diff
@@ -471,6 +473,29 @@ public HollowConsumerMetrics getMetrics() {
             return null;
         }
 
+        HollowConsumer.HeaderBlob retrieveHeaderBlob(long desiredVersion);
```
Sunjeet: We need a default implementation here too

hollow/src/main/java/com/netflix/hollow/api/client/HollowBlobRetriever.java
dkoszewnik: +1 to strict match.

Also can the default `throw new UnsupportedOperationException()`?

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
dkoszewnik: default == `throw new UnsupportedOperationException()`?

New features like this should be entirely optional.

hollow/src/main/java/com/netflix/hollow/api/metrics/HollowProducerMetrics.java
dkoszewnik: +1

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
dkoszewnik: I would go as far as to say all modified interfaces should get default interfaces.  I know the `BlobStager` is exposed as a public API -- nobody implementing it should have to modify their code to rev the library.

dkoszewnik (changes requested):
Overall looks good, but some changes needed for backwards compatibility, I think.  It should be 100% optional to use the feature.

hollow/src/test/java/com/netflix/hollow/api/producer/validation/HollowProducerValidationListenerTest.java
mbrilnetflix: It is used...

hollow-test/src/main/java/com/netflix/hollow/test/consumer/TestHeaderBlob.java
mbrilnetflix: Actually deleted the entire class.. it's not used

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
mbrilnetflix: Added a default impl

hollow/src/main/java/com/netflix/hollow/api/client/HollowBlobRetriever.java
Sunjeet: As you mentioned this is likely not used much, but just in case it would be better to throw an UnsupportedOperationException

hollow/src/main/java/com/netflix/hollow/api/client/HollowClientConsumerBridge.java
```diff
@@ -29,7 +30,22 @@
...
+                if(headerBlob == null)
+                    return null;
+
+                return new HollowConsumer.HeaderBlob(currentVersion) {
+                    @Override
+                    public InputStream getInputStream() throws IOException {
+                        // Getting input stream from HeaderBlob is not supported for ClientConsumerBridge.
+                        throw new UnsupportedOperationException();
+                    }
+                };
```
Sunjeet: Would it make sense to replace this whole block with an UnsupportedOperationException?

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -577,10 +587,55 @@ public InputStream decompress(InputStream is) {
...
+         * <p>
+         * It is guaranteed that {@code blob} was created by calling one of
+         * {@link BlobStager#openSnapshot(long)}, {@link BlobStager#openDelta(long, long)}, or
+         * {@link BlobStager#openDelta(long, long)} on this publisher, or
+         * {@link BlobStager#openReverseDelta(long, long)} on this publisher, or
+         * {@link BlobStager#openHeader(long)} on this publisher.
+         *
+         * @param publishArtifact the blob to publish
+         */
+        default void publish(HollowProducer.PublishArtifact publishArtifact) {
```
Sunjeet, line 606: Nit: Maybe a comment here that headers are a no-op, and 

since we marked publish(Blob) above deprecated, it'd be good to provide a default implementation (could throw an exception) so that consumers don't have to implement the deprecated method.

hollow/src/main/java/com/netflix/hollow/api/producer/fs/HollowFilesystemPublisher.java
```diff
@@ -45,6 +45,38 @@ public HollowFilesystemPublisher(Path blobStorePath) {
 
     @Override
     public void publish(HollowProducer.Blob blob) {
```
Sunjeet, line 48: nit: we can probably clean up calls to this deprecated method

hollow/src/test/java/com/netflix/hollow/api/producer/HollowProducerTest.java
```diff
@@ -519,5 +559,31 @@ public InputStream getInputStream() throws IOException {
...
         }
+
+        @Override
+        public HollowConsumer.HeaderBlob retrieveHeaderBlob(long desiredVersion) {
+            long blobVersion = desiredVersion;
+            File blobFile = headerFileMap.get(desiredVersion);
+            if (blobFile == null) {
+                // find the closest one
+                blobVersion = headerFileMap.keySet().stream()
+                        .filter(l -> l < desiredVersion)
```
Sunjeet: Double check this, looks like it needs to be updated for exact match

hollow/src/test/java/com/netflix/hollow/api/producer/HollowProducerTest.java
```diff
@@ -519,5 +559,31 @@ public InputStream getInputStream() throws IOException {
         public HollowConsumer.Blob retrieveReverseDeltaBlob(long currentVersion) {
             throw new UnsupportedOperationException();
         }
+
+        @Override
+        public HollowConsumer.HeaderBlob retrieveHeaderBlob(long desiredVersion) {
+            long blobVersion = desiredVersion;
+            File blobFile = headerFileMap.get(desiredVersion);
```
Sunjeet: Nit: do we need a `headerFileMap`, can we just use `headerBlobMap` instead? Btw is that being used anywhere?

hollow/src/test/java/com/netflix/hollow/api/consumer/InMemoryBlobStore.java
```diff
@@ -73,10 +80,37 @@ public Blob retrieveReverseDeltaBlob(long currentVersion) {
         return reverseDeltas.get(currentVersion);
     }
 
-    
-    
     @Override
-    public void publish(final HollowProducer.Blob blob) {
+    public HeaderBlob retrieveHeaderBlob(long desiredVersion) {
+        return (HeaderBlob) getDesiredVersion(desiredVersion, headers);
```
Sunjeet: Update to return exact version

hollow/src/test/java/com/netflix/hollow/api/producer/HollowProducerTest.java
mbrilnetflix: Done

hollow/src/test/java/com/netflix/hollow/api/producer/HollowProducerTest.java
mbrilnetflix: We do, as we will now have 2 files per version
Blob (Snapshot in this case) and HeaderBlob

Sunjeet (approved):

PR #556, opened 2021-12-10
Focus shard updates to reduce delta resource requirements.

dkoszewnik:
Currently, the effort required to process a type's delta is directly correlated with the size of the dataset, not the size of the delta.  For small updates on large types, this is suboptimal.  With this change, the effort required to process a delta of sharded types can be brought closer to being correlated with the size of the delta.

There are two changes here:
* The `HollowProducer` can be instructed to focus changes within a type to as few shards as possible where enough ordinal holes are available to fill.
* The `HollowConsumer` can be instructed to defer processing updates of shards where no additions are present in a delta.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
@@ -55,6 +55,8 @@ public class HollowClientUpdater {
     private final HollowConsumerMetrics metrics;
     private final HollowMetricsCollector<HollowConsumerMetrics> metricsCollector;
 
+    private boolean skipTypeShardUpdateWithNoAdditions;
+
     private TypeFilter filter;
 
     public HollowClientUpdater(HollowConsumer.BlobRetriever transitionCreator,
@@ -82,6 +84,13 @@ public class HollowClientUpdater {
         this.metrics = metrics;
         this.metricsCollector = metricsCollector;
         this.initialLoad = new CompletableFuture<>();
+    }
+
+    public void setSkipShardUpdateWithNoAdditions(boolean skipTypeShardUpdateWithNoAdditions) {
+        this.skipTypeShardUpdateWithNoAdditions = skipTypeShardUpdateWithNoAdditions;
+        HollowDataHolder dataHolder = hollowDataHolderVolatile;
+        if(dataHolder != null)
+            dataHolder.getStateEngine().setSkipTypeShardUpdateWithNoAdditions(skipTypeShardUpdateWithNoAdditions);
     }
 
     /**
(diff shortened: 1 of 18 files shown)


hollow/src/main/java/com/netflix/hollow/core/read/engine/list/HollowListTypeReadState.java
```diff
@@ -96,14 +97,35 @@ public void applyDelta(HollowBlobInput in, HollowSchema schema, ArraySegmentRecy
...
+                    oldRemovals.destroy();
+                } else {
+                    if(!deltaData.encodedRemovals.isEmpty()) {
+                        currentData.encodedRemovals = GapEncodedVariableLengthIntegerReader.combine(oldRemovals, deltaData.encodedRemovals, memoryRecycler);
+                        oldRemovals.destroy();
+                    }
+                    deltaData.encodedRemovals.destroy();
+                }
+
+                deltaData.encodedAdditions.destroy();
```
Sunjeet, line 119: Are we missing a `deltaData.destory()` here?

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
```diff
@@ -398,7 +398,11 @@ public void compact(ThreadSafeBitSet usedOrdinals) {
         }
 
         byteData.setPosition(currentCopyPointer);
-        freeOrdinalTracker.sort();
+
```
Sunjeet, line 401: minor: we could handle `numShards == 1` in the implementation of `sort(int)`

hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
```diff
@@ -84,6 +86,13 @@ public HollowClientUpdater(HollowConsumer.BlobRetriever transitionCreator,
         this.initialLoad = new CompletableFuture<>();
     }
 
+    public void setSkipShardUpdateWithNoAdditions(boolean skipTypeShardUpdateWithNoAdditions) {
```
Sunjeet, line 89: Do you think it would be good to have a similar method on the `HollowConsumer` as well so a consumer can toggle this behavior on the fly?

hollow/src/main/java/com/netflix/hollow/core/read/engine/list/HollowListTypeReadState.java
dkoszewnik: Yes!  Thank you.  Fixing now.

hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
dkoszewnik: Hmm, yes maybe.  That way we could potentially flip back with FP if we catch something in the integrity check.

PR #557, opened 2021-12-14
Fix null value for 64-bit wide fields.

dkoszewnik:
Bugfix: When a field requires 64 bits, the designated null value was incorrectly being calculated as 0 due to bit shift wrapping.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectTypeDataElements.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectTypeDataElements.java
@@ -139,7 +139,7 @@ public class HollowObjectTypeDataElements {
 
             if(unfilteredFieldIsIncluded[i]) {
                 bitsPerField[filteredFieldIdx] = readBitsPerField;
-                nullValueForField[filteredFieldIdx] = (1L << bitsPerField[filteredFieldIdx]) - 1;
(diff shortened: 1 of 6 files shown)


hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectTypeDataElements.java
```diff
@@ -139,13 +139,17 @@ private void readFieldStatistics(HollowBlobInput in, HollowObjectSchema unfilter
...
-                nullValueForField[filteredFieldIdx] = (1L << bitsPerField[filteredFieldIdx]) - 1;
+                nullValueForField[filteredFieldIdx] = bitsPerField[filteredFieldIdx] == 64 ? -1L : (1L << bitsPerField[filteredFieldIdx]) - 1;
                 bitOffsetPerField[filteredFieldIdx] = bitsPerRecord;
                 bitsPerRecord += bitsPerField[filteredFieldIdx];
                 filteredFieldIdx++;
             }
         }
     }
+    
+    public static void main(String args[]) {
```
Sunjeet: minor: cleanup

dkoszewnik: Fixed, thank you.

Sunjeet (approved):

PR #560, opened 2022-01-25
[DO NOT MERGE] Build history using reverse deltas

Sunjeet:
Building on @shoc2020 's work, this PR contians minimal changes to allow history to initialize to a version V, then build history by applying reverse deltas to versions V-1, V-2, etc. Did some rudimentary manual testing for display parity with building history the traditional way using deltas and it seems to be fine. 

Note that it doesn't currently handle double snapshots or backwards-compatible schema changes, and there may be some ancillary issues like header tags but in the simple case the history seems to be rendering fine. 

There is a unit test in there `startServerOnPorts7777And7778` that starts history servers initialized using fwd and reverse deltas on different ports for comparison (should look equivalent).

More to come...


Sunjeet:
Discontinued in favor of https://github.com/Netflix/hollow/pull/565

PR #564, opened 2022-03-22
Fix reverse delta header tags

Sunjeet:
Header tags set on reverse delta should be the same as those on the snapshot corresponding to the destination version (not those of source version). The old behavior caused issues when-
* pinning back to an older version, the header tags at the destination version weren't what you'd expect. Once place this caused an issue was in reporting the cycle start time metric which relies on a header tag- after traversing a reverse delta the value of the metric would reflect the origin state instead of the destination
* building history using reverse deltas- going from v2 to v1 we wouldn't have header tags set on the snapshot/delta to arrive at v1

### DIFF
--- a/hollow-test/src/main/java/com/netflix/hollow/test/consumer/TestHollowConsumer.java
+++ b/hollow-test/src/main/java/com/netflix/hollow/test/consumer/TestHollowConsumer.java
@@ -145,7 +145,6 @@ public class TestHollowConsumer extends HollowConsumer {
         // create a new write state for delta application, restore from current state
         HollowWriteStateEngine deltaState = HollowWriteStateCreator.createWithSchemas(getStateEngine().getSchemas());
         deltaState.restoreFrom(getStateEngine());
-        deltaState.overridePreviousStateRandomizedTag(this.getStateEngine().getCurrentRandomizedTag());
 
         // add all records from passed in {@code state} to delta write state
         HollowCombiner combiner = new HollowCombiner(HollowCombinerCopyDirector.DEFAULT_DIRECTOR,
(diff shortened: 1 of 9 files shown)


hollow-test/src/main/java/com/netflix/hollow/test/consumer/TestHollowConsumer.java
```diff
@@ -145,7 +145,6 @@ public TestHollowConsumer addDelta(long fromVersion, long toVersion, HollowWrite
         // create a new write state for delta application, restore from current state
         HollowWriteStateEngine deltaState = HollowWriteStateCreator.createWithSchemas(getStateEngine().getSchemas());
         deltaState.restoreFrom(getStateEngine());
-        deltaState.overridePreviousStateRandomizedTag(this.getStateEngine().getCurrentRandomizedTag());
```
Sunjeet, line 148: `restoreFrom` in the above line sets this, doesn't need to be called again

hollow/src/main/java/com/netflix/hollow/tools/patch/delta/HollowStateDeltaPatcher.java
```diff
@@ -114,6 +115,7 @@ public void prepareInitialTransition() {
     public void prepareFinalTransition() {
         writeEngine.prepareForNextCycle();
         writeEngine.overrideNextStateRandomizedTag(to.getCurrentRandomizedTag());
+        writeEngine.addHeaderTags(to.getHeaderTags());
```
Sunjeet, line 118: The second transition now also populates header tags corresponding to the final state

hollow/src/test/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListenerTest.java
```diff
@@ -209,4 +214,77 @@ public void refreshEndMetricsReporting(ConsumerRefreshMetrics refreshMetrics) {
...
+                        break;
+                    case 3:
+                        assertEquals(1L, refreshMetrics.getCycleStartTimestamp().getAsLong());
+                        break;
+                    case 4:
+                        assertEquals(3L, refreshMetrics.getCycleStartTimestamp().getAsLong());
+                        break;
+                }
+            }
+        }
```
mbrilnetflix, line 258: this is intersting, I don't have a better way to do it :), but, took me a sec to wrap my head around it.

mbrilnetflix (approved):

nayanika-u (approved):
Changes look good.

dkoszewnik:
LGTM

PR #565, opened 2022-04-08
feature: HollowHistory can build history using reverse deltas to allow building history in both directions simultaneously

Sunjeet:
Peering credit to @shoc2020 

HollowHistory maintains two read states- one that always moves in the forward direction and one that always moves in the reverse direction. Once max retention has been reached the read state corresponding to the reverse direction is dropped and (as before) each future forward delta application results in eviction of the oldest historical state.

Building history in reverse allows recent data versions to be available in the history tooling sooner at the time of HollowHistory initialization, and also has the nice property that history initialization uses the latest schema at the time of initialization so recently introduced backwards compatible schema changes are visible without the need to apply double snapshots.

The UI view of History built using reverse deltas is at parity with when building using forward deltas. 

NOTE: if computing history using reverse deltas, for header tags to be displayed accurately the producer needs to be on Hollow v7.0.2 - that release fixed a bug in how headers tags were set on reverse deltas.

Example usage:
(continue existing usage for fwd only)
```
consumerFwd.triggerRefreshTo(v);
HollowHistoryUIServer historyUIServer = new HollowHistoryUIServer(consumerFwd, PORT);
historyUIServer.start();
consumerFwd.triggerRefreshTo(v+1);
consumerFwd.triggerRefreshTo(v+2);
...
```
(new for building history in both directions)
```
consumerFwd.triggerRefreshTo(v);
  consumerRev.triggerRefreshTo(v);
HollowHistoryUIServer historyUIServer = new HollowHistoryUIServer(consumerFwd, consumerRev, PORT);
historyUIServer.start();
consumerFwd.triggerRefreshTo(v+1);
  consumerRev.triggerRefreshTo(v-1);
  consumerRev.triggerRefreshTo(v-2);
  consumerRev.triggerRefreshTo(v-3);
consumerFwd.triggerRefreshTo(v+2);
...

```

### DIFF
--- a/hollow-diff-ui/src/main/java/com/netflix/hollow/history/ui/HollowHistoryRefreshListener.java
+++ b/hollow-diff-ui/src/main/java/com/netflix/hollow/history/ui/HollowHistoryRefreshListener.java
@@ -22,22 +22,34 @@ import com.netflix.hollow.api.custom.HollowAPI;
 import com.netflix.hollow.core.read.engine.HollowReadStateEngine;
 import com.netflix.hollow.tools.history.HollowHistory;
 
+/**
+ * This can listener can be attached to either (a) a single consumer that always transitions in the forward direction of
+ * increasing versions (v1, v2, v3, etc.), or (b) to two consumers that start at the same version but one always transitions
+ * in the forward direction and the other always transitions in the reverse direction.
+ *
+ * This class synchronizes modifications to HollowHistory to prevent the two consumers (if applicable) from concurrently
+ * invoking modifications on the underlying HollowHistory object.
+ */
 public class HollowHistoryRefreshListener extends HollowConsumer.AbstractRefreshListener {
 
     private final HollowHistory history;
-    
+
     public HollowHistoryRefreshListener(HollowHistory history) {
         this.history = history;
     }
 
     @Override
-	public void snapshotUpdateOccurred(HollowAPI api, HollowReadStateEngine stateEngine, long version) throws Exception { 
+	public synchronized void snapshotUpdateOccurred(HollowAPI api, HollowReadStateEngine stateEngine, long version) throws Exception {
 		history.doubleSnapshotOccurred(stateEngine, version);
 	}
 	
 	@Override
-	public void deltaUpdateOccurred(HollowAPI api, HollowReadStateEngine stateEngine, long version) throws Exception {
-		history.deltaOccurred(version);
+	public synchronized void deltaUpdateOccurred(HollowAPI api, HollowReadStateEngine stateEngine, long version) throws Exception {
+		if (stateEngine == history.getOldestState()) {
+			history.reverseDeltaOccurred(version);
+		} else {
+			history.deltaOccurred(version);
+		}
 	}
 	
 	@Override public void refreshStarted(long currentVersion, long requestedVersion) { }
--- a/hollow-diff-ui/src/main/java/com/netflix/hollow/history/ui/HollowHistoryUI.java
+++ b/hollow-diff-ui/src/main/java/com/netflix/hollow/history/ui/HollowHistoryUI.java
@@ -64,6 +64,13 @@ public class HollowHistoryUI extends HollowUIRouter implements HollowRecordDiffU
 
     private String[] overviewDisplayHeaders;
 
+    /**
+     * HollowHistoryUI constructor that builds history for a consumer that transitions forwards i.e. in increasing
+     * version order (v1, v2, v3...). This constructor defaults max states to 1024 and time zone to PST.
+     *
+     * @param baseUrlPath url path for history UI endpoint
+     * @param consumer HollowConsumer (already initialized with data) that will be traversing forward deltas or double snapshots
+     */
     public HollowHistoryUI(String baseUrlPath, HollowConsumer consumer) {
         this(baseUrlPath, consumer, 1024, VersionTimestampConverter.PACIFIC_TIMEZONE);
     }
@@ -76,15 +83,21 @@ public class HollowHistoryUI extends HollowUIRouter implements HollowRecordDiffU
         this(baseUrlPath, createHistory(consumer, numStatesToTrack), timeZone);
     }
     
-    private static HollowHistory createHistory(HollowConsumer consumer, int numStatesToTrack) {
-        consumer.getRefreshLock().lock();
-        try {
-            HollowHistory history = new HollowHistory(consumer.getStateEngine(), consumer.getCurrentVersionId(), numStatesToTrack);
-            consumer.addRefreshListener(new HollowHistoryRefreshListener(history));
-            return history;
-        } finally {
-            consumer.getRefreshLock().unlock();
-        }
+    /**
+     * HollowHistoryUI that supports building history in both directions simultaneously.
+     * Fwd and rev consumers should be initialized to the same version before calling this constructor.
+     * This constructor defaults max states to 1024 and time zone to PST.
+     *
+     * @param baseUrlPath url path for history UI endpoint
+     * @param consumerFwd HollowConsumer (already initialized with data) that will be traversing forward deltas or double snapshots
+     * @param consumerRev HollowConsumer (also initialized to the same version as consumerFwd) that will be traversing reverse deltas
+     */
+    public HollowHistoryUI(String baseUrlPath, HollowConsumer consumerFwd, HollowConsumer consumerRev) {
+        this(baseUrlPath, consumerFwd, consumerRev, 1024, VersionTimestampConverter.PACIFIC_TIMEZONE);
+    }
+
+    public HollowHistoryUI(String baseUrlPath, HollowConsumer consumerFwd, HollowConsumer consumerRev, int numStatesToTrack, TimeZone timeZone) {
+        this(baseUrlPath, createHistory(consumerFwd, consumerRev, numStatesToTrack), timeZone);
     }
 
     public HollowHistoryUI(String baseUrlPath, HollowHistory history) {
@@ -105,13 +118,44 @@ public class HollowHistoryUI extends HollowUIRouter implements HollowRecordDiffU
         this.viewProvider = new HollowHistoryViewProvider(this);
         this.diffViewOutputGenerator = new DiffViewOutputGenerator(viewProvider);
 
-        this.customHollowEffigyFactories = new HashMap<String, CustomHollowEffigyFactory>();
-        this.customHollowRecordNamers = new HashMap<String, HollowHistoryRecordNamer>();
-        this.matchHints = new HashMap<String, PrimaryKey>();
+        this.customHollowEffigyFactories = new HashMap<>();
+        this.customHollowRecordNamers = new HashMap<>();
+        this.matchHints = new HashMap<>();
         this.overviewDisplayHeaders = new String[0];
         this.timeZone = timeZone;
     }
-    
+
+    private static HollowHistory createHistory(HollowConsumer consumer, int numStatesToTrack) {
+        return createHistory(consumer, null, numStatesToTrack);
+    }
+
+    private static HollowHistory createHistory(HollowConsumer consumerFwd, HollowConsumer consumerRev, int numStatesToTrack) {
+        if (consumerRev == null) {
+            consumerFwd.getRefreshLock().lock();
+            try {
+                HollowHistory history = new HollowHistory(consumerFwd.getStateEngine(), consumerFwd.getCurrentVersionId(), numStatesToTrack);
+                consumerFwd.addRefreshListener(new HollowHistoryRefreshListener(history));
+                return history;
+            } finally {
+                consumerFwd.getRefreshLock().unlock();
+            }
+        } else {
+            consumerFwd.getRefreshLock().lock();
+            consumerRev.getRefreshLock().lock();
+            try {
+                HollowHistory history = new HollowHistory(consumerFwd.getStateEngine(), consumerRev.getStateEngine(),
+                        consumerFwd.getCurrentVersionId(), consumerRev.getCurrentVersionId(), numStatesToTrack);
+                consumerFwd.addRefreshListener(new HollowHistoryRefreshListener(history));
+                consumerRev.addRefreshListener(new HollowHistoryRefreshListener(history));
+                return history;
+            } finally {
+                consumerFwd.getRefreshLock().unlock();
+                consumerRev.getRefreshLock().unlock();
+            }
+
+        }
+    }
+
     public HollowHistory getHistory() {
         return history;
     }
--- a/hollow-diff-ui/src/test/java/com/netflix/hollow/diffview/HollowHistoryUITest.java
+++ b/hollow-diff-ui/src/test/java/com/netflix/hollow/diffview/HollowHistoryUITest.java
@@ -1,0 +1,269 @@
+package com.netflix.hollow.diffview;
+
+import static com.netflix.hollow.diffview.FakeHollowHistoryUtil.assertUiParity;
+import static org.junit.Assert.assertNotNull;
+
+import com.netflix.hollow.history.ui.HollowHistoryUI;
+import com.netflix.hollow.history.ui.jetty.HollowHistoryUIServer;
+import com.netflix.hollow.test.consumer.TestBlobRetriever;
+import com.netflix.hollow.test.consumer.TestHollowConsumer;
+import org.junit.Before;
+import org.junit.Test;
+
+public class HollowHistoryUITest {
+
+    private final int PORT_EXPECTED = 7777;
+    private final int PORT_ACTUAL = 7778;
+
+    private TestBlobRetriever testBlobRetriever;
+    private TestHollowConsumer consumerExpected;    // builds history using only deltas
+    private TestHollowConsumer consumerFwd;
+    private TestHollowConsumer consumerRev;
+    private HollowHistoryUIServer historyUIServerExpected;
+    private HollowHistoryUIServer historyUIServerActual;
+    private HollowHistoryUI historyUiExpected;  // built using fwd deltas application only
+
+    public HollowHistoryUITest() throws Exception {
+        testBlobRetriever = new TestBlobRetriever();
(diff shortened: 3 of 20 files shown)


hollow-diff-ui/src/main/java/com/netflix/hollow/history/ui/HollowHistoryRefreshListener.java
```diff
@@ -22,22 +22,38 @@
 import com.netflix.hollow.core.read.engine.HollowReadStateEngine;
 import com.netflix.hollow.tools.history.HollowHistory;
 
+/**
+ * This can listener can be attached to a consumer for building HollowHistory in the forward direction (v1, v2, v3, etc.),
+ * or to 2 consumers for building history in both directions simultaneously.
+ *
```
mbrilnetflix, line 29: The grammar of this sentence seems off

Sunjeet: reworded

hollow-diff-ui/src/main/java/com/netflix/hollow/history/ui/HollowHistoryUI.java
```diff
@@ -77,16 +77,53 @@ public HollowHistoryUI(String baseUrlPath, HollowConsumer consumer, int numState
     }
     
     private static HollowHistory createHistory(HollowConsumer consumer, int numStatesToTrack) {
```
mbrilnetflix: nit: Please add comment to clarify this will be a Forward only

hollow-diff-ui/src/main/java/com/netflix/hollow/history/ui/HollowHistoryUI.java
```diff
@@ -77,16 +77,53 @@ public HollowHistoryUI(String baseUrlPath, HollowConsumer consumer, int numState
...
-        try {
-            HollowHistory history = new HollowHistory(consumer.getStateEngine(), consumer.getCurrentVersionId(), numStatesToTrack);
-            consumer.addRefreshListener(new HollowHistoryRefreshListener(history));
-            return history;
-        } finally {
-            consumer.getRefreshLock().unlock();
+        return createHistory(consumer, null, numStatesToTrack);
+    }
+
+    private static HollowHistory createHistory(HollowConsumer consumerFwd, HollowConsumer consumerRev, int numStatesToTrack) {
```
mbrilnetflix: we don't support a rev only consumer? in that case, consumerFwd will be null and throw a NPE
If we do not, it's fine, we can add a null check.

hollow-diff-ui/src/main/java/com/netflix/hollow/history/ui/HollowHistoryUI.java
```diff
@@ -77,16 +77,53 @@ public HollowHistoryUI(String baseUrlPath, HollowConsumer consumer, int numState
...
+            consumerRev.getRefreshLock().lock();
+            try {
+                HollowHistory history = new HollowHistory(consumerFwd.getStateEngine(), consumerRev.getStateEngine(),
+                        consumerFwd.getCurrentVersionId(), consumerRev.getCurrentVersionId(), numStatesToTrack);
+                consumerFwd.addRefreshListener(new HollowHistoryRefreshListener(history));
+                consumerRev.addRefreshListener(new HollowHistoryRefreshListener(history));
+                return history;
+            } finally {
+                consumerFwd.getRefreshLock().unlock();
+                consumerRev.getRefreshLock().unlock();
```
mbrilnetflix: NPE if Rev is null

hollow/src/main/java/com/netflix/hollow/tools/history/HollowHistory.java
```diff
@@ -83,17 +106,75 @@ public HollowHistory(HollowReadStateEngine initialHollowStateEngine, long initia
...
-        this.latestVersion = initialVersion;
+
+        // validate fwd moving state initialization
+        requireNonNull(fwdMovingHollowReadStateEngine, "Fwd direction read state engine should be initialized");
+        if (fwdInitialVersion == VERSION_NONE) {
+            throw new IllegalArgumentException("Valid version corresponding to fwdMovingHollowReadStateEngine should be specified" +
+                    "during HollowHistory initialization");
+        }
+        this.latestHollowReadStateEngine = fwdMovingHollowReadStateEngine;
+        this.fwdInitialVersion=  fwdInitialVersion;
```
mbrilnetflix: checkstyle didn't complain about this?

hollow/src/main/java/com/netflix/hollow/tools/history/HollowHistory.java
```diff
@@ -83,17 +106,75 @@ public HollowHistory(HollowReadStateEngine initialHollowStateEngine, long initia
...
+        this.latestVersion = fwdInitialVersion;
+        this.latestHeaderEntries = latestHollowReadStateEngine.getHeaderTags();
+
+        // rev moving state, may or may not be specified at initialization
+        if (revMovingHollowReadStateEngine != null || revInitialVersion != VERSION_NONE) {
```
mbrilnetflix, line 172: this can be confusing IMHO, 
We should throw exception if either is set and other is not.

hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowHistoryTypeKeyIndex.java
```diff
@@ -93,14 +93,19 @@ public void updateReadStateEngine(HollowReadStateEngine readEngine) {
...
 
-        if (isDelta)
-            populateNewCurrentRecordKeys(latestTypeState);
-        else
-            populateAllCurrentRecordKeys(latestTypeState);
+        if (isDeltaAndIndexInitialized) {
+            // record all newly seen keys when going from v1->v2 or v2->v1 to the history type key index
+            populateNewCurrentRecordKeysIntoIndex(latestTypeState);
+        }
+        else {
```
mbrilnetflix: nit: this seems odd fomating

hollow/src/test/java/com/netflix/hollow/core/util/RemovedOrdinalIteratorTest.java
```diff
@@ -22,11 +22,11 @@
 
 public class RemovedOrdinalIteratorTest {
 
+    private BitSet previousOrdinals = bitSet(1, 2, 3, 4, 6, 7, 9, 10);
+    private BitSet currentOrdinals = bitSet(1, 3, 4, 5, 7, 8, 9);
```
mbrilnetflix: these are not reset on each test (when multiple test cases are ran)

mbrilnetflix (commented):
LGTM if assumptions mentioned in my comments are confirmed, 
I'll defer to Drew for this one, as I don't have full context on this.

hollow/src/main/java/com/netflix/hollow/tools/history/HollowHistory.java
```diff
@@ -354,8 +552,29 @@ private void addHistoricalState(HollowHistoricalState historicalState) {
...
+        } else { // if reverse delta occurs before any fwd deltas
+            historicalState.getDataAccess().setNextState(latestHollowReadStateEngine);
+        }
+
+        historicalStates.add(historicalState);
+        historicalStateLookupMap.put(historicalState.getVersion(), historicalState);
+
+        if (historicalStates.size() >= maxHistoricalStatesToKeep) {
+            // drop old read state because we won't be building history in reverse after we get here
+            oldestHollowReadStateEngine = null;
```
Sunjeet, line 572: Ah, just learned that `HollowHistoricalState`s hold some references into the typestates contained within `oldestHollowHistoryState` so dereferencing it here won't actually free up much memory. I think we can drop those references once the `HollowHistoricalState` has been populated. Adding that...

Sunjeet: Added code to dereference read state engine once HollowHistoricalState has been populated so it can be GC'ed.

hollow-diff-ui/src/main/java/com/netflix/hollow/history/ui/HollowHistoryUI.java
Sunjeet: The underlying `HollowHistory` initialization has a 
```
requireNonNull(fwdMovingHollowReadStateEngine, "Fwd direction read state engine should be initialized");
```
and yes, we don't support rev only because it isn't really useful and it would add complexity- api that accepts a single state engine would need to also accept/determine which direction it will be traversing in

hollow-diff-ui/src/main/java/com/netflix/hollow/history/ui/HollowHistoryUI.java
Sunjeet: consumerRev is null checked before getting into this `else` block. ConsumerFwd is null checked within HollowHistory init

hollow/src/main/java/com/netflix/hollow/tools/history/HollowHistory.java
Sunjeet: yes! the call to `initializeReverseStateEngine` in the next line validates the inputs

hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowHistoryTypeKeyIndex.java
Sunjeet: fixed

hollow/src/test/java/com/netflix/hollow/core/util/RemovedOrdinalIteratorTest.java
Sunjeet: They weren't intended to be. I failed to make them static final. Now addressed.

hollow-diff-ui/src/main/java/com/netflix/hollow/history/ui/HollowHistoryUI.java
dkoszewnik: +1

Sunjeet: added

hollow/src/main/java/com/netflix/hollow/tools/history/HollowHistory.java
Sunjeet: Sounds like something python linter would do :) 
Anyway, fixed

hollow/src/main/java/com/netflix/hollow/tools/history/HollowHistoricalStateCreator.java
```diff
@@ -97,27 +101,31 @@ public HollowHistoricalStateDataAccess createBasedOnNewDelta(long version, Hollo
...
 
-    private void createDeltaHistoricalTypeState(IntMapOrdinalRemapper typeRemovedOrdinalMapping, List<HollowTypeReadState> historicalTypeStates, HollowTypeReadState typeState) {
+    private void createDeltaHistoricalTypeState(IntMapOrdinalRemapper typeRemovedOrdinalMapping, List<HollowTypeReadState> historicalTypeStates, HollowTypeReadState typeState, boolean reverse) {
         if(typeState instanceof HollowObjectTypeReadState) {
-            HollowObjectDeltaHistoricalStateCreator deltaHistoryCreator = new HollowObjectDeltaHistoricalStateCreator((HollowObjectTypeReadState)typeState);
+            HollowObjectDeltaHistoricalStateCreator deltaHistoryCreator = new HollowObjectDeltaHistoricalStateCreator((HollowObjectTypeReadState)typeState, reverse);
             deltaHistoryCreator.populateHistory();
             typeRemovedOrdinalMapping.addOrdinalRemapping(typeState.getSchema().getName(), deltaHistoryCreator.getOrdinalMapping());
             historicalTypeStates.add(deltaHistoryCreator.createHistoricalTypeReadState());
+            deltaHistoryCreator.dereferenceTypeState();
```
dkoszewnik, line 111: Can you please help me understand why we need this?

dkoszewnik (approved):
It looks fantastic and solves an important problem.  I'm very excited that our history/explorer UI will soon detect schema changes, then re-initialize the history in reverse, so new fields will be automatically visible while still also providing visibility into a long history of changes.

hollow/src/main/java/com/netflix/hollow/core/read/engine/list/HollowListDeltaHistoricalStateCreator.java
```diff
@@ -70,6 +70,12 @@ public void populateHistory() {
         }
     }
 
+    public void dereferenceTypeState() {
```
nayanika-u, line 73: A comment or java doc would be helpful here to understand why/where this method is required.

hollow-diff-ui/src/main/java/com/netflix/hollow/history/ui/HollowHistoryUI.java
```diff
@@ -76,15 +83,21 @@ public HollowHistoryUI(String baseUrlPath, HollowConsumer consumer, int numState
...
+     * HollowHistoryUI that supports building history in both directions simultaneously.
+     * Fwd and rev consumers should be initialized to the same version before calling this constructor.
+     * This constructor defaults max states to 1024 and time zone to PST.
+     *
+     * @param baseUrlPath url path for history UI endpoint
+     * @param consumerFwd HollowConsumer (already initialized with data) that will be traversing forward deltas or double snapshots
+     * @param consumerRev HollowConsumer (also initialized to the same version as consumerFwd) that will be traversing reverse deltas
+     */
+    public HollowHistoryUI(String baseUrlPath, HollowConsumer consumerFwd, HollowConsumer consumerRev) {
+        this(baseUrlPath, consumerFwd, consumerRev, 1024, VersionTimestampConverter.PACIFIC_TIMEZONE);
```
nayanika-u, line 96: nit : could we make the max states configurable?

hollow-diff-ui/src/test/java/com/netflix/hollow/diffview/HollowHistoryUITest.java
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    @Test(expected=NullPointerException.class)
+    public void historyUsingFwdOnly_noPastVersionsAvailableAtInit()  {
+        // consumerFwd hasn't incurred snapshot load yet
+        historyUIServerActual = new HollowHistoryUIServer(consumerFwd, PORT_ACTUAL);
+    }
+
+    @Test(expected=UnsupportedOperationException.class)
+    public void historyUsingFwdAndRevConsumer_revConsumerMustBeInitialized() throws Exception  {
```
nayanika-u, line 223: How does the HistoryUIServer recover in the 2 scenarios - a) revConsumer and fwdConsumer are on different versions at init - When would that happen? b) revConsumer not initialized. Do we restart the history server with less or maybe 1 state?

hollow-diff-ui/src/main/java/com/netflix/hollow/history/ui/HollowHistoryUI.java
Sunjeet: (on the constructor)

hollow/src/main/java/com/netflix/hollow/tools/history/HollowHistoricalStateCreator.java
Sunjeet: Added code comment

hollow/src/main/java/com/netflix/hollow/core/read/engine/list/HollowListDeltaHistoricalStateCreator.java
Sunjeet: Added

hollow-diff-ui/src/main/java/com/netflix/hollow/history/ui/HollowHistoryUI.java
Sunjeet: Yup they are, this constructor is just plugging in a default to the other constructor that accepts it as a param

hollow-diff-ui/src/test/java/com/netflix/hollow/diffview/HollowHistoryUITest.java
Sunjeet: (a) The history server fails hard with an IllegalStateException at initialization when it detects that reverse consumer didnt start at the same version that the fwd consumer started with. There is also a unit test `historyUsingFwdAndRevConsumer_revAndFwdConsumersMustBeOnSameVersionAtInit` in `HollowHistoryUITest.java` that exercises this scenario. We could potentially make improvements to handle the various initialization scenarios like there is a gap in the start versions that gets stitched together,  or the fwd and rev initialize so that they will pass each other at some point. There isnt a compelling reason to handle those scenarios at the moment.

Sunjeet: (b) revConsumer can be initialized at any point (for e.g. after >1 historical states have been created using the fwd consumer). So if fwd and rev start at v0, at that point there are no historical states. When fwd transitions to v1 we have one historical state (historical state v1 representing the diff between v0 and v1). More fwd transition applications would lead to more historical states v2, v3, v4, etc. When the rev consumer comes along its first transition (v0 to v-1) will result in the creation of the historical state v0 (giving the same effect as if the fwd consumer had transitioned from v-1 to v0).

hollow-diff-ui/src/main/java/com/netflix/hollow/history/ui/HollowHistoryUI.java
Sunjeet: Hah, we were missing the max states configurable flavor constructor in the `HollowHistoryUIServer` class. Not sure if thats the one you may have been thinking of here. Now added- https://github.com/Netflix/hollow/commit/d4d3dbed944a8885ba128843a61e569ddbcce08e

PR #567, opened 2022-07-05
Add a new HollowUniqueKeyIndex that supports object longevity. Update HollowHashIndex to support object longevity.

Jorel-Master-Of-Scheduling:
Add an alternative to HollowPrimaryKeyIndex called HollowUniqueKeyIndex. It uses HollowDataAccess instead of HollowObjectTypeReadState. This allows HollowUniqueKeyIndex to opt-in to Object Longevity by using the HollowDataAccess from a Consumer rather than the HollowObjectTypeReadState.

HollowObjectTypeReadState is mutable and gets updated with each delta.  If a PrimaryKeyIndex does NOT listen to updates, then it will become stale over time. The internal index structure will remain unchanged, but the HollowPrimaryKeyValueDeriver will start returning incorrect keys as the ordinals get re-used and re-assigned to new records.

By migrating to HollowDataAccess, users can choose to use the HollowObjectTypeReadState (which stays up to date) or use HollowApi.getDataAccess() which may be object-longevity-aware.


hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
```diff
@@ -216,8 +217,8 @@ public void endUpdate() {
        reindexHashIndex();
     }
 
-    public HollowReadStateEngine getStateEngine() {
-        return stateEngine;
+    public HollowDataAccess getStateEngine() {
```
dkoszewnik: I don't think we should change the public API here.

hollow/src/main/java/com/netflix/hollow/core/index/TestableUniqueKeyIndex.java
```diff
@@ -0,0 +1,24 @@
+package com.netflix.hollow.core.index;
+
+import java.util.Collection;
+
+/**
+ * This package is for internal use. Do not depend on it.
```
dkoszewnik, line 6: Should we mark it deprecated?

dkoszewnik (approved):
Looks good to me, thank you James.

hollow/src/main/java/com/netflix/hollow/core/index/key/PrimaryKey.java
```diff
@@ -31,6 +33,30 @@
...
+     * @param type             hollow type
+     * @param fieldPaths       field paths for fields that make up the primary key. If no fields are passed in, then create using schema definition of primary key
+     * @return populated primary key
+     */
+    public static PrimaryKey create(HollowDataAccess hollowDataAccess, String type, String... fieldPaths) {
+        if (fieldPaths != null && fieldPaths.length != 0) {
+            return new PrimaryKey(type, fieldPaths);
+        }
+
+        HollowSchema schema = hollowDataAccess.getSchema(type);
```
mbrilnetflix, line 53: nit: can `hollowDataAccess` be null? it's ok if we have fieldPaths

hollow/src/main/java/com/netflix/hollow/core/index/TestableUniqueKeyIndex.java
mbrilnetflix: +1

mbrilnetflix (approved):
LGTM

mbrilnetflix (approved):

hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
Jorel-Master-Of-Scheduling: Oops! Let me fix that.

hollow/src/main/java/com/netflix/hollow/core/index/key/PrimaryKey.java
Jorel-Master-Of-Scheduling: Yes, I think so. Added a comment to the javadoc.

hollow/src/main/java/com/netflix/hollow/core/index/TestableUniqueKeyIndex.java
Jorel-Master-Of-Scheduling: Done.

PR #573, opened 2022-11-14
Explorer Null Issue in text mode

mailsanchu:
The problem is there is an additional null check for the text stringifier and looks like there is a bug in there.
The following are the debugging values I extracted from debugging and you can see the issue can be easily reproducible with these values.
```    
       long fixedLengthValue = 8189729;
        int numBits = 33;
        long result = (fixedLengthValue & (1 << (numBits - 1)));
        System.out.println("result = " + result); 


        long endByte = 8189729;
        int numBitsForField = 33;
        result = (endByte & (1L << (numBitsForField - 1)));
        System.out.println("result = " + result);


dkoszewnik:
Awesome!  Thank you very much @mailsanchu

PR #574, opened 2022-11-17
HollowObjectTypeReadStateShard iterate readString ints in a single pass

jasonk000:
This change introduces an uplift to `readString` performance for longer strings, in three parts
1. Introduce a new benchmark
2. Introduce `readVIntsInto` method to read a block of varints into an `int[]` or `char[]` as a batch
3. Switch `readString` to use the new approach.

Also I threw in a slight tidy-up of threadlocal usage for simplicity.
```
public static int readVIntsInto(ByteData arr, long position, int length, char[] output)
public static int readVIntsInto(ByteData arr, long position, int length, int[] output)
```

```
INITIAL

Benchmark                                               (countStrings)  (countStringsDb)  (maxStringLength)  (probabilityUnicode)  Mode  Cnt  Score   Error  Units
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                  5                    10  avgt   10  0.044 ± 0.001  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                 25                    10  avgt   10  0.059 ± 0.003  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                 50                    10  avgt   10  0.095 ± 0.003  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                150                    10  avgt   10  0.237 ± 0.004  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000               1000                    10  avgt   10  1.505 ± 0.056  ms/op

WITH THREADLOCAL NIT

Benchmark                                               (countStrings)  (countStringsDb)  (maxStringLength)  (probabilityUnicode)  Mode  Cnt  Score   Error  Units
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                  5                    10  avgt   10  0.043 ± 0.001  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                 25                    10  avgt   10  0.058 ± 0.002  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                 50                    10  avgt   10  0.089 ± 0.003  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                150                    10  avgt   10  0.275 ± 0.025  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000               1000                    10  avgt   10  1.541 ± 0.182  ms/op

WITH CHANGES

Benchmark                                               (countStrings)  (countStringsDb)  (maxStringLength)  (probabilityUnicode)  Mode  Cnt  Score   Error  Units
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                  5                    10  avgt   10  0.044 ± 0.001  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                 25                    10  avgt   10  0.058 ± 0.001  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                 50                    10  avgt   10  0.074 ± 0.002  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                150                    10  avgt   10  0.191 ± 0.005  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000               1000                    10  avgt   10  0.936 ± 0.120  ms/op
```


Sunjeet:
Wow, the perf numbers look great!

hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectTypeReadStateShard.java
```diff
@@ -314,24 +315,18 @@ private long fieldOffset(HollowObjectTypeDataElements currentData, int ordinal,
...
 
     private String readString(ByteData data, long position, int length) {
-        long endPosition = position + length;
-
-        char chararr[] = getCharArray();
-
-        if(length > chararr.length)
+        char[] chararr = HollowObjectTypeReadStateShard.chararr.get();
+        if (length > chararr.length) {
             chararr = new char[length];
```
Sunjeet, line 323: Is there a reason to not set the threadlocal `chararr` to this new bigger array?

hollow/src/main/java/com/netflix/hollow/core/memory/encoding/VarInt.java
```diff
@@ -167,6 +167,46 @@ public static int readVInt(ByteData arr, long position) {
...
+     * @param length the number of bytes that should be read
+     * @param output an array where outputs will be placed, which should be at least {@code length} long and zero.
+     * @return the number of values written
+     */
+    public static int readVIntsInto(ByteData arr, long position, int length, char[] output) {
+        int count = 0;
+        for(int i = 0; i < length; i++) {
+            int b = arr.get(position + i);
+
+            output[count] = (char) ((output[count] << 7) | (b & 0x7f));
```
Sunjeet, line 221: I don't know much about Hotspot but do you think [this optimization](https://github.com/protocolbuffers/protobuf/blob/main/java/core/src/main/java/com/google/protobuf/Utf8.java#L660-L669) (from protobuf decoding) could be something to try?

jasonk000: Hard to explore with the error bars, however it seems that ascii-only strings gives a further notable improvement for what _might_ be a very slight penalty - very likely a net benefit.

```
BEFORE FASTPATH

Benchmark                                               (countStrings)  (countStringsDb)  (maxStringLength)  (probabilityUnicode)  Mode  Cnt  Score   Error  Units
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                  5                     0  avgt   20  0.048 ± 0.001  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                  5                    10  avgt   20  0.049 ± 0.001  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                 25                     0  avgt   20  0.059 ± 0.001  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                 25                    10  avgt   20  0.059 ± 0.003  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                 50                     0  avgt   20  0.080 ± 0.003  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                 50                    10  avgt   20  0.076 ± 0.001  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                150                     0  avgt   20  0.165 ± 0.006  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                150                    10  avgt   20  0.179 ± 0.009  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000               1000                     0  avgt   20  0.907 ± 0.040  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000               1000                    10  avgt   20  0.921 ± 0.043  ms/op


AFTER FASTPATH

Benchmark                                               (countStrings)  (countStringsDb)  (maxStringLength)  (probabilityUnicode)  Mode  Cnt  Score   Error  Units
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                  5                     0  avgt   20  0.045 ± 0.004  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                  5                    10  avgt   20  0.044 ± 0.001  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                 25                     0  avgt   20  0.049 ± 0.001  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                 25                    10  avgt   20  0.047 ± 0.002  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                 50                     0  avgt   20  0.068 ± 0.001  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                 50                    10  avgt   20  0.072 ± 0.001  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                150                     0  avgt   20  0.157 ± 0.003  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000                150                    10  avgt   20  0.171 ± 0.007  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000               1000                     0  avgt   20  0.602 ± 0.048  ms/op
HollowObjectTypeReadStateShardBenchmark.testReadString             500            100000               1000                    10  avgt   20  0.938 ± 0.027  ms/op
```

hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectTypeReadStateShard.java
jasonk000: Only to retain existing behaviour, I'm not sure what the original rationale was.

hollow/src/main/java/com/netflix/hollow/core/memory/encoding/VarInt.java
Sunjeet: Oh, wow

hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectTypeReadStateShard.java
Sunjeet: May have been an oversight, what do you think @dkoszewnik ?

dkoszewnik: Not an oversight -- the rationale is that we don't want an occasional extremely long string to cause the ThreadLocal to hang on to an extremely large buffer.

hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectTypeReadStateShard.java
```diff
@@ -314,24 +315,18 @@ private long fieldOffset(HollowObjectTypeDataElements currentData, int ordinal,
...
-
-        int count = 0;
-
-        while(position < endPosition) {
-            int c = VarInt.readVInt(data, position);
-            chararr[count++] = (char)c;
-            position += VarInt.sizeOfVInt(c);
+        } else {
+            Arrays.fill(chararr, 0, length, '\0');
```
dkoszewnik, line 325: Is this necessary?  Why can't we just leave it dirty?

hollow/src/main/java/com/netflix/hollow/core/memory/encoding/VarInt.java
```diff
@@ -167,6 +167,64 @@ public static int readVInt(ByteData arr, long position) {
...
+                break;
+            output[i] = b;
+        }
+
+        int count = i;
+        for(; i < length; i++) {
+            int b = arr.get(position + i);
+
+            output[count] = (output[count] << 7) | (b & 0x7f);
+            count += (~b >> 7) & 0x1;
```
dkoszewnik, line 193: Neat :)

dkoszewnik (approved):
Awesome, thank you.

hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectTypeReadStateShard.java
jasonk000: It needs to start as 0's because the loop in `VarInt` requires an empty array (as it accumulates from a starting point of 0). This code specifically requires the initial value to be a zero'd array. 
```
             output[count] = (output[count] << 7) | (b & 0x7f);
```

Alternative options might be:
- do the zero'ing inside `VarInt::readVIntsInto`
- perform `Arrays.fill()` _after_ the `new String` construction, so that we only fill exactly the length we used
- add some more comments explaining ; I added some comments to the javadoc on `::readVIntsInto` but it seems potentially error-prone.

LMK thoughts.

dkoszewnik: I see it now!  Thank you.

hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectTypeReadStateShard.java
Sunjeet: Ok, makes sense. Thankyou!

PR #575, opened 2022-11-18
In code generator retrofit support for custom type names that start with a lowercase character

Sunjeet:
5 years ago when *DataAccessor classes were added to the generated API we lost support for custom type names that start with a lower case character (for e.g. "profileId").  This PR retro-fits that support to unblock one such known use case.

In general it would be good to not allow type names that start with a lower case character but that would be a breaking change for clients.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/codegen/HollowConsumerJavaFileGenerator.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/codegen/HollowConsumerJavaFileGenerator.java
@@ -118,7 +118,7 @@ public abstract class HollowConsumerJavaFileGenerator implements HollowJavaFileG
                     }
                 }
                 for (String schemaName : schemaNameSet) {
-                    appendImportFromBasePackage(builder, schemaName + config.getClassPostfix());
+                    appendImportFromBasePackage(builder, HollowCodeGenerationUtils.upperFirstChar(schemaName) + config.getClassPostfix());
                 }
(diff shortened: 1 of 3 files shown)


hollow/src/main/java/com/netflix/hollow/api/codegen/HollowConsumerJavaFileGenerator.java
```diff
@@ -118,7 +118,7 @@ protected void appendPackageAndCommonImports(StringBuilder builder,
                     }
                 }
                 for (String schemaName : schemaNameSet) {
-                    appendImportFromBasePackage(builder, schemaName + config.getClassPostfix());
```
akhaku, line 121: for the actual generated class, is it always uppercase?

Sunjeet: Yes we do that here https://github.com/Netflix/hollow/blob/45e19a67d3db4432085874af03809c3cd5c88e7e/hollow/src/main/java/com/netflix/hollow/api/codegen/HollowCodeGenerationUtils.java#L164
https://github.com/Netflix/hollow/blob/45e19a67d3db4432085874af03809c3cd5c88e7e/hollow/src/main/java/com/netflix/hollow/api/codegen/HollowCodeGenerationUtils.java#L177

Accessors seem the only place where we don't

akhaku (approved):

PR #576, opened 2022-11-21
Tag announcement metadata with readstate size on heap

Sunjeet:


hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -62,6 +64,8 @@
 abstract class AbstractHollowProducer {
 
     static final long DEFAULT_TARGET_MAX_TYPE_SHARD_SIZE = 16L * 1024L * 1024L;
+    // An announcement metadata tag indicating the approx heap footprint of the corresponding read state engine
+    private static final String ANNOUNCE_TAG_HEAP_FOOTPRINT = "hollow.data.size.heap.approx";
```
nayanika-u: LGTM. Minor - indicating the data size is in bytes could be helpful. Can do either in the comment or the tag itself.

nayanika-u (approved):

PR #552, opened 2021-11-02
Replace Jetty server with JDK httpserver

shoc2020:
httpserver port changes

PR #581, opened 2023-01-26
Replace use of jetty-server with com.sun.net.httpserver included in jdk

Sunjeet:
Credit to Shobhan Chinnam for most of the heavy lifting in this [PR#552](https://github.com/Netflix/hollow/pull/552)

This PR replaces the use of `jetty-server` as a `compileOnly` dependency (earlier clients were required to provide jetty-server implementation at runtime or else they would get an `OptionalDependencyException`) with the jdk-included `HttpServer` and `HttpServlet`. 

Often times developers need to write a unit test involving hollow ui tooling for e.g. spin up an explorer locally and they would forget to declare dependency on jetty-server only to incur an exception at runtime and then go back and add the dep. Promoting the existing dependency on `jetty-server` from `compileOnly` to `implementation` is less desirable given past dependency conflicts and jetty api version compatibility issues.

This change introduces the`javax.servlet:javax.servlet-api` JAR on the runtime for clients that depend on `hollow-diff-ui`, `hollow-explorer-ui`, or `hollow-tools-ui`.

The `HttpServer` lives under `com.sun.net.*` under jdk versions 8 through 17.
The `HttpServlet` lives under `javax.servlet.http.*` and comes from Oracle Java EE 7 in jdk8, but that changes in jdk11. In 11 through 77 Oracle Java EE is dropped in favor of Jakarta EE maintained by the Eclipse foundation, so the implication is that `javax.*` packages are moved to `jakarta.*`. [[source](https://blogs.oracle.com/javamagazine/post/its-time-to-move-your-applications-to-java-17-heres-why-and-heres-how), [jakarta.servlet.http.HttpServlet](https://jakarta.ee/specifications/platform/9/apidocs/jakarta/servlet/http/httpservlet)]


dkoszewnik (approved):
Awesome, thank you Sunjeet!

PR #587, opened 2023-02-14
Bug fixes for UI tools

Sunjeet:
This PR includes 2 bugfixes- 
* Propagate sever exceptions back to client browser. This regression was introduced when jetty was swapped for jdk webserver. Also refactored `handle` to return a boolean to be fully backwards compatibility with old behavior although no known usages of that return value.
* Explorer UI had a bug where ordinal value 0 was processed the same as null leading to loss of some information on the UI pages corresponding to ordinal zero. This was potentially introduced when upgrading velocity major version. Velocity #if directive documentation can be found [here](https://velocity.apache.org/engine/devel/user-guide.html#if-elseif-else)

### DIFF
--- a/hollow-explorer-ui/src/main/resources/browse-selected-type-top.vm
+++ b/hollow-explorer-ui/src/main/resources/browse-selected-type-top.vm
@@ -37,7 +37,7 @@ NOTE: Results are filtered by query <span style="font-family: monospace;">$esc.h
 				<a href="$basePath/type?type=$esc.url($type)&key=$esc.url($key)&ordinal=$ordinal&display=$display&page=$nextPage&pageSize=$pageSize">NEXT</a>
 			#end
 		</td>
-		#if($ordinal >= 0)
+		#if($ordinal != $null)
 			<td class="nav">
 				FORMAT: 
 				<a href="$basePath/type?type=$esc.url($type)&key=$esc.url($key)&ordinal=$ordinal&page=$page&pageSize=$pageSize&display=text">text</a> 
@@ -89,12 +89,12 @@ NOTE: Results are filtered by query <span style="font-family: monospace;">$esc.h
 						</form>
 					</td>
 				</tr>
-				#if($ordinal >= 0)
+				#if($ordinal != $null)
 					<tr valign=top>
 						<th align=right nowrap>Showing Record: </th>
 						<td id="recLink">
 							ordinal: $ordinal
-							#if($key && !$key.isEmpty())
+							#if($key)	<!-- not null and not empty -->
 								<br/>key: $esc.html($key)
 							#end
 						</td>
@@ -106,4 +106,4 @@ NOTE: Results are filtered by query <span style="font-family: monospace;">$esc.h
 					</th>
 					<td>
 						<div style="min-height:400px">
-							#if($ordinal >= 0)<pre>#end
+							#if($ordinal != $null)<pre>#end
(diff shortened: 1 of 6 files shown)


hollow-explorer-ui/src/main/resources/browse-selected-type-top.vm
```diff
@@ -37,7 +37,7 @@ NOTE: Results are filtered by query <span style="font-family: monospace;">$esc.h
 				<a href="$basePath/type?type=$esc.url($type)&key=$esc.url($key)&ordinal=$ordinal&display=$display&page=$nextPage&pageSize=$pageSize">NEXT</a>
 			#end
 		</td>
-		#if($ordinal)
+		#if($ordinal != $null)
```
nayanika-u, line 40: Have you verified if this null check works? I think checking against $null might cause some issues. 
Existing checks should suffice. The value null is tested as the boolean value false by Velocity. You could either add this - 
#if (!$variable)
 ... do stuff (nothing) here if the variable is null
#end
#if( $variable == true )
	...Variable is NOT null
#end
Ref - https://itecnote.com/tecnote/java-null-check-in-velocity/ 
http://oliviertech.com/java/velocity/how-to-check-if-a-value-is-null-with-velocity/

hollow-ui-tools/src/main/java/com/netflix/hollow/ui/HollowUIRouter.java
```diff
@@ -32,8 +34,18 @@ public abstract class HollowUIRouter extends HttpServlet {
     protected final VelocityEngine velocityEngine;
 
     @Override
-    public abstract void doGet(HttpServletRequest request,
-        HttpServletResponse response) throws IOException;
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        try {
+            handle(request.getPathInfo(), request, response);
```
nayanika-u, line 39: The return type boolean isn't being checked. Why was the method signature changed?

nayanika-u: The commit message mentions the reason for refactoring. Did you see any issue which led to this refactoring?

hollow-explorer-ui/src/main/resources/browse-selected-type-top.vm
Sunjeet: Yes this is tested and its the most direct way to check null, relies on the variable $null not being assigned to anything. The other ways as you noted are more roundabout. Unfortunately simply doing `#if ($ordinal)` doesnt seem to distinguish between null and 0 value when ordinal is an int.

hollow-ui-tools/src/main/java/com/netflix/hollow/ui/HollowUIRouter.java
Sunjeet: Did not discover any issues due to this, and did not see any usage of the return value within our code bases (but someone externally could be depending on it), but retained the return value for backwards compatibility with versions upto 7.1 (it was an oversight to drop it in 7.2)

hollow-explorer-ui/src/main/resources/browse-selected-type-top.vm
Sunjeet: from velocity 2x documentation- https://velocity.apache.org/engine/2.2/user-guide.html#if-elseif-else
refer `#if ($ref == $null) to specifically test for the null value (provided you didn't put anything in $null)`

PR #588, opened 2023-02-24
bugfix: tooling NPE with unexpected string in cookie

Sunjeet:


nayanika-u (approved):

Sunjeet:
Fixed in https://github.com/Netflix/hollow/pull/595 instead

PR #592, opened 2023-03-13
Add announcement timestamp in the header tags. 

nayanika-u:
This will be useful in measuring the duration between an announcement and a successful consumer refresh for a given version.


hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
```diff
@@ -163,29 +173,36 @@ public void refreshFailed(long beforeVersion, long afterVersion, long requestedV
...
+        trackTimestampsFromHeaders(version, stateEngine.getHeaderTags());
     }
 
     /**
-     * If the blob header contains producer cycle start time tag then save its value in version to cycle start time map
+     * If the blob header contains the timestamps like producer cycle start and announcement then save those values in
+     * the maps tracking version to cycle start time and version to announcement respectively.
      */
-    private void trackCycleStartTime(long version, Map<String, String> headers) {
+    private void trackTimestampsFromHeaders(long version, Map<String, String> headers) {
```
Sunjeet: Refactoring with this method signature should simplify the implementation
```
trackTimestampsFromHeaders(long version, Map<String, String> headers, String headerTag, Map<String, Long> timestampsMap)
```

nayanika-u: Thanks. Done.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
```diff
@@ -118,6 +124,21 @@ private final void noFailRefreshEndMetricsReporting(ConsumerRefreshMetrics refre
         }
     }
 
+    @Override
+    public void announcementDetected(long newVersion, Map<String, String> metadata, boolean isPinned) {
+        /** Track the timestamp for HEADER_TAG_METRIC_ANNOUNCEMENT only when the namespace is not pinned.
+        * Store the pinned status of previous cycle to detect the scenario where a version got unpinned in the next cycle
+        * then don't record this metric as this will record the duration of current version from the last announced version.
+        * namespacePinnedPreviously == true && isPinned == true then dont record
```
nayanika-u: just added this comment bit on the condition for PR review discussion.

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
```diff
@@ -860,6 +865,8 @@ public interface RefreshListener {
          */
         void refreshFailed(long beforeVersion, long afterVersion, long requestedVersion, Throwable failureCause);
 
+        default void announcementDetected(long newVersion, Map<String, String> metadata, boolean isPinned) {}
```
Sunjeet: nit: move method definition up to be before refresh started, making it the first method in the interface. Bring order in the sequence that they are invoked in during refresh helps with readability

Sunjeet: add javadoc

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
```diff
@@ -136,6 +157,10 @@ public void refreshSuccessful(long beforeVersion, long afterVersion, long reques
             refreshMetricsBuilder.setCycleStartTimestamp(cycleVersionStartTimes.get(afterVersion));
         }
 
+        if (afterVersion == requestedVersion && announcementTimestamps.containsKey(afterVersion)) {
+            refreshMetricsBuilder.setAnnouncementTimestamp(announcementTimestamps.get(afterVersion));
```
Sunjeet, line 165: This will NPE if key afterVersion doesn't exist in map

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
```diff
@@ -118,6 +124,21 @@ private final void noFailRefreshEndMetricsReporting(ConsumerRefreshMetrics refre
...
+    @Override
+    public void announcementDetected(long newVersion, Map<String, String> metadata, boolean isPinned) {
+        /** Track the timestamp for HEADER_TAG_METRIC_ANNOUNCEMENT only when the namespace is not pinned.
+        * Store the pinned status of previous cycle to detect the scenario where a version got unpinned in the next cycle
+        * then don't record this metric as this will record the duration of current version from the last announced version.
+        * namespacePinnedPreviously == true && isPinned == true then dont record
+        * namespacePinnedPreviously == true && isPinned == false then dont record
+        * namespacePinnedPreviously == false && isPinned == true then dont record
+        * namespacePinnedPreviously == false && isPinned == false then record */
+        if (!(namespacePinnedPreviously || isPinned)) {
```
Sunjeet: consider slightly refactoring to- always record version to announce time in this map, but only report metric if the pin/unpin criteria is validated

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
Sunjeet: Ignore, didn't see the containsKey above

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
Sunjeet: Ah I see, isPinned is only available here.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
```diff
@@ -55,11 +56,15 @@ public abstract class AbstractRefreshMetricsListener extends AbstractRefreshList
     ConsumerRefreshMetrics.Builder refreshMetricsBuilder;
 
     private final Map<Long, Long> cycleVersionStartTimes;
+    private final Map<Long, Long> announcementTimestamps;
```
Sunjeet, line 59: There is possibility of concurrent modification to this map from the refresh thread and the thread that calls `fireAnnouncementDetected` because of the way we've wired up the announcement metadata/ispinned propagtation. Lets peer on fixing this- and try to maintain the norm of all listeners invoked from only the refresh thread.

nayanika-u: Thanks for pointing this out. As discussed in our peering sessions, this concern is now addressed by change in HollowClientUpdater's logic. Now the announcement timestamps are recorded using the same thread which also triggers refresh started and completion.

hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
```diff
@@ -130,6 +134,13 @@ public synchronized boolean updateTo(long requestedVersion) throws Throwable {
         for (HollowConsumer.RefreshListener listener : localListeners)
             listener.refreshStarted(beforeVersion, requestedVersion);
 
+        for(HollowConsumer.RefreshListener listener : localListeners) {
+            // both pinning status of the version and headers map is required further
+            if (versionInfo.isPinned().isPresent() && versionInfo.getAnnouncementMetadata().isPresent()) {
+                listener.announcementDetected(requestedVersion, versionInfo.getAnnouncementMetadata().get(), versionInfo.isPinned().get());
```
Sunjeet: Sorry I realize we were going back and forth on whether this should be its own event or whether to collapse it into refreshStarted, I think given how it fires right around when refreshStarted fires, IMO it makes sense to add this as another flavor of refreshStarted, so
```
listener.refreshStarted(versionInfo, requestedVersion);
```
while also keeping the old one around as-is. Note we're also passing the versionInfo pojo instead of its constituent fields (client updater doesn't  care if versionInfo contains pin status and announcement metadata, the specific metrics listener does) .

Sunjeet (commented):
A minor suggestion around the listener interface, rest looks good to go

hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
nayanika-u: That makes sense based on how the events get fired in the listener. My concern is distinguishing between the two flavors of refreshStarted. Right now, refreshStarted instantiates metric builders and clears any existing state (header value maps in this case). just to clarify - is this more about renaming announcementDetected() method?

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
```diff
@@ -822,14 +822,13 @@ public void staleReferenceExistenceDetected(int count) {
      */
     public interface RefreshListener {
         /**
-         * Indicates that announcement information for new version has been detected. This information can be either be
-         * recorded or used in computing different metrics within Hollow.
+         * Indicates that a new version has been detected. This method requires the requested version's information
```
Sunjeet: javadoc  is intended as reference for users (as opposed to implementers) so it should read like "this method surfaces the requested version's info ..." (as opposed to this method "requires".. which would be more suitable style for code comment for implementers). Suggestion-

Called when a new version has been detected in consumer refresh (and just before refreshStarted). Surfaces metadata and pinning status pertaining to requested version if available from AnnouncementWatcher. Generally useful for logging/metrics.

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
```diff
@@ -67,15 +67,30 @@ public AbstractRefreshMetricsListener() {
...
         refreshMetricsBuilder = new ConsumerRefreshMetrics.Builder();
         refreshMetricsBuilder.setIsInitialLoad(currentVersion == VERSION_NONE);
         refreshMetricsBuilder.setUpdatePlanDetails(updatePlanDetails);
         cycleVersionStartTimes.clear(); // clear map to avoid accumulation over time
+    }
+
+    @Override
+    public void versionDetected(HollowConsumer.VersionInfo requestedVersionInfo) {
         announcementTimestamps.clear(); // clear map to avoid accumulation over time
+        if (!(requestedVersionInfo.isPinned().isPresent() && requestedVersionInfo.getAnnouncementMetadata().isPresent())) {
```
Sunjeet, line 85: would be good to add null checks here for future proofing (current code paths for constructing VersionInfo don't pass null, but VersionInfo allows null)

hollow/src/main/java/com/netflix/hollow/api/consumer/metrics/AbstractRefreshMetricsListener.java
```diff
@@ -124,16 +139,22 @@ private final void noFailRefreshEndMetricsReporting(ConsumerRefreshMetrics refre
...
-    public void announcementDetected(long newVersion, Map<String, String> metadata, boolean isPinned) {
-        // Track the version to announcement timestamp only when the namespace is not pinned (either in previous cycle
-        // or for the newVersion). Don't record this metric when a namespace was pinned previously and gets unpinned
-        // in the next cycle because this metric will record the refresh duration from the latest announced version.
-        if (!(namespacePinnedPreviously || isPinned)) {
-            trackTimestampsFromHeaders(newVersion, metadata, HEADER_TAG_METRIC_ANNOUNCEMENT, announcementTimestamps);
-        }
-        namespacePinnedPreviously = isPinned;
-    }
+//    @Override
```
Sunjeet: nit: cleanup

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
nayanika-u: Thanks for the note and suggestion. Have updated the javadoc. 
I refrained from adding the note regarding its usage before refreshStarted as, generically speaking, this method could be used at any point during a consumer refresh depending on its implementation. Thought this specification would limit the method's usage and interpretation.

Sunjeet (approved):

PR #595, opened 2023-03-20
Fix #594 - expolrer-ui can handle more than one cookie

### LINKED ISSUE #594: The new explorer-ui will break if there are more than 1 cookie
The `HttpHandlerWithServletSupport.RequestWrapper#getCookies()` will generate an array of `Cookie[]`, and will allocate the number of `Cookie` headers in the request (the `req.get("Cookie")` will return all headers named "Cookie", not all cookies pairs - should be always 1), but will try to populate it with number of tokens in that headers (might be more than 1).

image

yonatang:
https://github.com/Netflix/hollow/issues/594
Fixing the Explorer UI which fails if the browser have more than one cookie token in the cookie header.


dkoszewnik:
Thanks very much @yonatang, I just ran into this too.

PR #596, opened 2023-03-23
enable pojo caching in HollowObjectCacheProvider

workeatsleep:
1. enable pojo caching in HollowObjectCacheProvider
2. modify the detach logic so the right exception is thrown when object is accessed after detach


hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
```diff
@@ -231,6 +233,10 @@ public void triggerRefresh() {
         }
     }
 
+    public void fireAnnouncementDetected(long newVersion, Map<String, String> metadata, boolean isPinned) {
```
nayanika-u, line 236: Just FYI - You might want to clean up this PR as it has an additional unrelated commit.

PR #600, opened 2023-03-25
expose pojo caching in consumer

workeatsleep:
enable pojo caching in HollowObjectCacheProvider
modify the detach logic so the right exception is thrown when object is accessed after detach


hollow/src/main/java/com/netflix/hollow/api/codegen/HollowAPIClassJavaGenerator.java
```diff
@@ -175,6 +175,7 @@ public String generate() {
 
         builder.append("    }\n\n");
 
+        builder.append("/** Objects are no longer accessible after this method is called, do not call this method unless the object will not be used anymore. */ \n");
```
Sunjeet: nit: missing couple of `\n`s to make it a multi line comment

Sunjeet: Nit: rephrase to "Cached objects are no longer accessible after this method is called and an attempt to access them will throw an IllegalStateException".

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
```diff
@@ -1102,13 +1102,20 @@ public B withRefreshListeners(HollowConsumer.RefreshListener... refreshListeners
          * @return this builder
          * @throws IllegalArgumentException if provided API class is {@code HollowAPI} instead of a subclass
          */
-        public B withGeneratedAPIClass(Class<? extends HollowAPI> generatedAPIClass) {
+        public B withGeneratedAPIClass(Class<? extends HollowAPI> generatedAPIClass, String cachedType, String... additionalCachedTypes) {
             if (HollowAPI.class.equals(generatedAPIClass))
                 throw new IllegalArgumentException("must provide a code generated API class");
-            this.apiFactory = new HollowAPIFactory.ForGeneratedAPI<>(generatedAPIClass);
+            String[] types = new String[additionalCachedTypes.length + 1];
```
Sunjeet: nit: can guard against `additionalCachedTypes` being null if method is called like `.withGeneratedAPIClass(MovieAPI.class, "Movie", null)`

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
```diff
@@ -1102,13 +1102,20 @@ public B withRefreshListeners(HollowConsumer.RefreshListener... refreshListeners
          * @return this builder
          * @throws IllegalArgumentException if provided API class is {@code HollowAPI} instead of a subclass
          */
-        public B withGeneratedAPIClass(Class<? extends HollowAPI> generatedAPIClass) {
+        public B withGeneratedAPIClass(Class<? extends HollowAPI> generatedAPIClass, String cachedType, String... additionalCachedTypes) {
```
Sunjeet: The javadoc for the old flavor of this method got tacked on to this one. Need to move that below, and we can add some javadoc to this one. @toolbear  had a nice writeup in his PR https://github.com/Netflix/hollow/pull/379/files 
Also, thanks Tim for the code contribution in that PR that we reused here!

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
Sunjeet: And Tim's PR also had some code for handling null values passed in the additionalCachedTypes array. We could retain that too.

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
```diff
@@ -1102,13 +1102,20 @@ public B withRefreshListeners(HollowConsumer.RefreshListener... refreshListeners
...
-            this.apiFactory = new HollowAPIFactory.ForGeneratedAPI<>(generatedAPIClass);
+            String[] types = new String[additionalCachedTypes.length + 1];
+            types[0] = cachedType;
+            System.arraycopy(additionalCachedTypes, 0, types, 1, additionalCachedTypes.length);
+            this.apiFactory = new HollowAPIFactory.ForGeneratedAPI<>(generatedAPIClass, types);
             return (B)this;
         }
 
+        public B withGeneratedAPIClass(Class<? extends HollowAPI> generatedAPIClass) {
+            return withGeneratedAPIClass(generatedAPIClass, null);
```
Sunjeet: This is in turn leading to an array of one element equal to `null` being passed to the underlying `HollowAPIFactory`. I think the other comment regarding handling of null values in the cached types will fix this, or we could also just retain the original implementation of this method

hollow/src/main/java/com/netflix/hollow/api/objects/provider/HollowObjectCacheProvider.java
```diff
@@ -84,11 +84,18 @@ public HollowObjectCacheProvider(HollowTypeDataAccess typeDataAccess, HollowType
 
     @Override
     public T getHollowObject(int ordinal) {
-        return cachedItems.get(ordinal);
+        List<T> refCachedItems = cachedItems;   // SNAP: object cache
```
Sunjeet: This methods seems to be WIP

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
workeatsleep: Thanks @toolbear !

Sunjeet:
Since I have a commit on the PR, I've lost the Approve button. But its a Ship It! from me

hollow/src/main/java/com/netflix/hollow/api/objects/provider/HollowObjectCacheProvider.java
```diff
@@ -84,11 +84,18 @@ public HollowObjectCacheProvider(HollowTypeDataAccess typeDataAccess, HollowType
...
     public T getHollowObject(int ordinal) {
-        return cachedItems.get(ordinal);
+        List<T> refCachedItems = cachedItems;
+        if (refCachedItems == null) {
+            throw new IllegalStateException(String.format("HollowObjectCacheProvider for type %s has been detached or was not initialized", typeReadState == null ? null : typeReadState.getSchema().getName()));
+        }
+        if (refCachedItems.size() <= ordinal) {
+            throw new IllegalStateException(String.format("Ordinal %s is out of bounds for pojo cache array of size %s.", ordinal, refCachedItems.size()));
+        }
+        return refCachedItems.get(ordinal);
```
nayanika-u, line 94: Is the return type for getHollowObject nullable here? Is that expected? When does it become null?

Sunjeet: yes if for e.g.`api.getMovie(ordinal)` is called with an ordinal that isn't populated that will return a null- that seems reasonable. This behavior from `HollowObjectCacheProvider` (for cached types) is different from behavior in `HollowObjectFactoryProvider` (non cached types) where `HollowObjectFactoryProvider` will return garbage results if called with ordinals that aren't populated - i dont think thats ideal but it likely doesn't matter- its basically saying responsibility is on user of low level api (when passing ordinal) to ensure that it is called with ordinal that is populated current state.

nayanika-u (approved):

PR #608, opened 2023-04-28
FlatRecords can be returned to POJOs

dkoszewnik:
The `HollowObjectMapper` can read a `FlatRecord` back into the original POJO.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/HollowListTypeMapper.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/HollowListTypeMapper.java
@@ -17,16 +17,19 @@
 package com.netflix.hollow.core.write.objectmapper;
 
 import com.netflix.hollow.core.schema.HollowListSchema;
+import com.netflix.hollow.core.schema.HollowSchema;
 import com.netflix.hollow.core.util.IntList;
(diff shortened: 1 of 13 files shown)


eduardoramirez:
@Sunjeet 

A few notes/questions about the implementation

- for `Map`, `Set` and `List` schema types, i create a `HashMap`, `HashSet` and `ArrayList` for its parsed representation object. the assumption i made here is that the POJOs define fields using interfaces, is that a safe assumption? or, do POJOs use concrete classes in their definitions?

- in this implementation, the POJO is required to have a default constructor

Sunjeet (approved):

Sunjeet:
@eduardoramirez nice work in the PR. 
Re: your above comment (realized after merging PR that I can't comment reply to your comment anymore), 
* Most if not all data models I have seen use concrete POJOs. I haven't looked at many FlatRecord emitter apps but our internal Incremental Producer guide seems to suggest adding POJOs so I think that might be prevalent there too. 
In the past @PaulSandoz had some ideas around decoupling concrete implementations on producer and consumer side with interfaces and moving Hollow to a more schema-first model. There are some docs and sample code available internally, I can share those with you for context. 
* The default constructor requirement seems fairly easy to impose at write time or code generation time

PR #610, opened 2023-05-04
bugfix: HollowCombiner should call shouldCopy only for populated ordinals

Sunjeet:


dkoszewnik:
I suspect this behavior should be added to the `HollowCombinerCopyDirector.DEFAULT_DIRECTOR` rather than directly in the `HollowCombiner` -- if we do that instead, we get the same fix but we also won't preclude the possibility of e.g. using the `HollowCombiner` to copy the previous state of a `HollowReadStateEngine` that just transitioned through a delta.

PR #613, opened 2023-05-17
Consumer auto double snapshot on schema change

Sunjeet:
To use the feature, a consumer must 
(a) be built with double snapshot config that elects to double snapshot on schema change
(b) be able to lookup a schema hash attribute in version metadata (so producer must update code to start publishing this, and announcement watcher or the call to triggerRefreshTo must provide metadata)

Note no support for optional parts atm.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
@@ -16,6 +16,8 @@
  */
 package com.netflix.hollow.api.client;
 
+import static com.netflix.hollow.core.HollowStateEngine.HEADER_TAG_SCHEMA_HASH;
+
 import com.netflix.hollow.api.consumer.HollowConsumer;
 import com.netflix.hollow.api.custom.HollowAPI;
 import com.netflix.hollow.api.metrics.HollowConsumerMetrics;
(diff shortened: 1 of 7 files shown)


hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
```diff
@@ -251,9 +255,44 @@ public void forceDoubleSnapshotNextUpdate() {
...
+            }
+            Map<String, String> metadata = incomingVersionInfo.getAnnouncementMetadata().get();
+            if (metadata.get(HEADER_TAG_SCHEMA_HASH) == null) {
+                LOG.warning("Double snapshots on schema change are enabled but version metadata for incoming " +
+                        "version " + incomingVersionInfo.getVersion() + " did not contain the required attribute (" +
+                        HEADER_TAG_SCHEMA_HASH + "). Check that the producer supports setting this attribute. This " +
+                        "refresh will not be able to reflect any schema changes.");
+                return false;
+            }
+            if (!doubleSnapshotConfig.allowDoubleSnapshot()) {
```
nayanika-u, line 283: This can be moved to line 263 before the other checks are performed.

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -372,6 +373,7 @@ long runCycle(
 
             // 3. Produce a new state if there's work to do
             if (writeEngine.hasChangedSinceLastCycle()) {
+                writeEngine.addHeaderTag(HollowStateEngine.HEADER_TAG_SCHEMA_HASH, new HollowSchemaHash(writeEngine.getSchemas()).getHash());
                 boolean schemaChangedFromPriorVersion = readStates.hasCurrent() &&
```
nayanika-u, line 377: Wondering if `schemaChangedFromPriorVersion` should also be computed by comparing the SchemaHash values to ensure there is a sync between the 2 header values - HEADER_TAG_SCHEMA_HASH and HEADER_TAG_SCHEMA_CHANGE.

PR #617, opened 2023-05-25
Add a method to determine restore version for a producer in HollowCon…

nayanika-u:
…sumer.AnnouncementWatcher

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
@@ -694,6 +694,13 @@ public class HollowConsumer {
         default VersionInfo getLatestVersionInfo() {
             return new VersionInfo(getLatestVersion(), Optional.empty(), Optional.empty());
         }
+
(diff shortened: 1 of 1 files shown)


hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
```diff
@@ -694,6 +694,13 @@ public interface AnnouncementWatcher {
         default VersionInfo getLatestVersionInfo() {
             return new VersionInfo(getLatestVersion(), Optional.empty(), Optional.empty());
         }
+
+        /**
+         * @return versionInfo - the version determined for a HollowProducer to restore to while commencing a new cycle, its pinned status and announcement metadata.
+         * */
+        default VersionInfo determineVersionInfoForProducerRestore(long currentVersion) {
+            throw new UnsupportedOperationException("This feature is limited and is currently supported by only a specific implementation of HollowAnnouncementWatcher.");
```
Sunjeet, line 702: chatted offline, we dont need to do this

PR #618, opened 2023-06-05
Add a field to track cycle count of a producer with primary status

nayanika-u:


hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -319,6 +322,7 @@ public HollowObjectMapper getObjectMapper() {
     public boolean enablePrimaryProducer(boolean doEnable) {
         if (doEnable) {
             singleProducerEnforcer.enable();
+            cycleCountWithPrimaryStatus = 0;
```
Sunjeet: This resets the value even if this producer instance was already primary, so I think we need this from [here](https://github.com/Netflix/hollow/commit/8e37a2b787886bbe398395968dbfc725d08b0942)-
```
if (!singleProducerEnforcer.isPrimary()) {
            cycleNumSincePrimary = 0;
        }
```

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -86,6 +86,9 @@ abstract class AbstractHollowProducer {
     long lastSuccessfulCycle = 0;
     final HollowObjectHashCodeFinder hashCodeFinder;
     final boolean doIntegrityCheck;
+    // Count to track number of cycles run by a primary producer. In the future, this can be useful in determining stickiness of a
+    // producer instance.
+    int cycleCountWithPrimaryStatus = 0;
```
Sunjeet: nit: cycleCountSincePrimaryStatus

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -419,6 +424,8 @@ long runCycle(
 
                 log.info("Populate stage completed with no delta in output state; skipping publish, announce, etc.");
             }
+            // increment the cycle count only for success case i.e with or without any detected changes
+            cycleCountWithPrimaryStatus ++;
```
Sunjeet: Wondering if this should happen in the `finally` block i.e. we should increment it even for failed cycles. Is there a reason to not do that?

nayanika-u: I think we discussed this offline about incrementing it only for successful cycles so that we don't have to account for different failure modes - esp publish blob or announce failure where a producer may lose primary status. We have guardrails in those methods for checking primary status but logically it would make more sense to just count successful cycles. Less probable but by having this here, we don't have to worry about wrong restore leading to validation failures and exception cases in populate stage? If cycle failed, we do a consistency check again on restore. Also, this would be applicable to future reporting use cases.

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
nayanika-u: Ack. Changed

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
nayanika-u: Thanks for the catch.I missed the part that enablePrimaryProducer() is designed to be called before _every_ runCycle() invocation. Will change it.

Sunjeet: Yeah with the current single producer enforcer api we could see both these usages-

```
producer.enablePrimaryProducer(true)
while(true) runCycle(...);
```

and
``` 
while(true) {
    producer.enablePrimaryProducer(true);
    runCycle(...);
}
```

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
Sunjeet: But the benefit of counting both success and failed cycles i.e. moving the increment to the finallly block is that for the current feature i.e. account for eventual consistency on the first cycle since primary, we dont care if the cycle was success or fail- and if we only count success cycles then we potentially account for eventual consistency on multiple cycles before the first success cycle - seems unnecessary. 

Also thinking that we can drop these as they are redundant
```
// Logically, no new producer will resume from a mid-cycle point of a previous cycle but resetting
// this value here so that the caller code reads the correct value at all times
cycleCountSincePrimaryStatus = 0;
```
as the cycle count would get reset at the top of the next cycle anyway. The thing about 
>the caller code reads the correct value at all times

I dont know if thats entirely true, because in the success case the producer can give up primary status willingly and there will be a duration before the next invocation where the count has not been reset.

nayanika-u: Okay. Yes, waiting on restore due to cycle stage failures would be redundant in few scenarios. 

Re - resets in publish and announce where we check primary status seem "extra" but are useful. The read call in NFHollowProducer happens before runCycle. I was trying to reset the count wherever the primary status is lost so that anytime after this a read call would be able to see the latest value but, as you mentioned, this doesn't cover the case of graceful release of primary status. 

Yes, you're right. RN we don't have a way to determine when a graceful give up of primary status happens (would need some tweaking on NfCuratorSingleProducer level.). Actually, we shouldn't even have to worry about how the release is happening. We just keep checking primary status at different stages of a cycle and reset(not needed now; reset at start of the cycle should suffice). Will drop these.

nayanika-u: So could i keep the condition but change the comment to account for the 'graceful give up of primary status' case?

nayanika-u: Discussed offline. We wouldn't need to care about resetting the value in publish/announce. Worst case, a producer loses primary status, waits for consistency check and runs cycle to eventually result in a no-op. Another instance (if primary) will start from clean slate.

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -434,6 +441,8 @@ long runCycle(
             throw new RuntimeException(th);
         } finally {
             artifacts.cleanup();
+            // increment the cycle count for a producer with primary status for both cases of cycle runs - success or failure.
```
Sunjeet: Just a very minor note on code style- comments like this one and the one on line 341 are a bit superfluous, so the added verbosity feels a little less clean. For e.g. here it is pretty clear from a quick glance that if an action is performed in a finally block it is invoked regardless of success or failure, and for line 341 again a quick glance reveals that if instance is not primary then we are doing that action.

Sunjeet (approved):

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
nayanika-u: Ack. Noted.

nayanika-u: I think I was going after explaining the increment condition and it would have made more sense in the previous commit when increment was only being done for successful cycles.

PR #623, opened 2023-06-20
Significantly improved efficiency of HollowHistoryKeyIndex

stevenewald:
This PR optimizes `HollowHistoryTypeKeyIndex` by eliminating the need for both readStateEngine and writeStateEngine, effectively removing a round trip for each state update. This significantly enhances performance in terms of CPU usage and memory management.

In the newly proposed implementation, `HollowHistoryTypeKeyIndex` maintains the following associations in Java HashMaps:

1. Hashed ordinals mapped to assigned ordinals
2. Hashed fields mapped to assigned ordinals
3. Assigned ordinals mapping to the corresponding record's field string representations

This approach has a few key advantages:

1. **Efficient Updates**: Round trips are completely eliminated for key updates, instead, updates to mappings occur based on the latest typeState.
2. **Immediate Storing and Hashing**: Keys are stored and hashed immediately upon update and within the same function, removing the need for added indirection present previously.
3. **Enhanced Querying Performance**: Querying is now significantly faster due to the direct lookup in a hashmap instead of going through the readStateEngine.

The modified HollowHistoryTypeKeyIndex is also more concise, with about 300 lines of code as compared to the previous 500 lines, contributing to improved readability.

The performance impact of these changes is considerable. Preliminary profiling indicates that history processing is 3-4 times faster. Additionally, memory usage is likely reduced as it's no longer necessary to store objects in conjunction with their hashes and mappings.

Todos (potentially after initial CR, will discuss with Drew)

 - [x] ~~Implement support for multi-field key delimiters during queries~~
 - [x] Store field values by object, rather than by string
 - [x] Fix potential hash collision edge case where records are not equal but hashes are
 - [x] Implement memoization for fields
     - [x] Strings: intern() builtin
     - [x] Other (bytearray, float, etc): implement intern equivalent

Unit/integration tests for potential sources of bugs (or make sure they already exist):
- [x] Unit tests for ObjectInternPool
- [ ] Record hash collision test? inherently hard to find two records that collide though

## Profiler screenshots to illustrate the performance improvements
Note the stack percentages in the bottom left
### Old implementation
Old implementation
### New implementation
New implementation

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowHistoryKeyIndex.java
+++ b/hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowHistoryKeyIndex.java
@@ -18,20 +18,10 @@ package com.netflix.hollow.tools.history.keyindex;
 
 import com.netflix.hollow.core.HollowDataset;
 import com.netflix.hollow.core.index.key.PrimaryKey;
-import com.netflix.hollow.core.read.HollowBlobInput;
-import com.netflix.hollow.core.read.engine.HollowBlobReader;
 import com.netflix.hollow.core.read.engine.HollowReadStateEngine;
 import com.netflix.hollow.core.read.engine.object.HollowObjectTypeReadState;
 import com.netflix.hollow.core.util.SimultaneousExecutor;
-import com.netflix.hollow.core.write.HollowBlobWriter;
-import com.netflix.hollow.core.write.HollowWriteStateEngine;
 import com.netflix.hollow.tools.history.HollowHistory;
-import java.io.BufferedInputStream;
-import java.io.BufferedOutputStream;
-import java.io.Closeable;
-import java.io.IOException;
-import java.io.PipedInputStream;
-import java.io.PipedOutputStream;
 import java.util.HashMap;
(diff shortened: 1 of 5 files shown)


stevenewald:
The added memoization functionality (ObjectInternPool) works by checking if the same boxed primitive already exists in a HashMap. If so, it returns a reference to the existing one (and the garbage collector will deallocate the additional one).

In effect, this makes it so if many records have the same values (1f, 10d, "United States", etc), it will only have to store one copy of each. If we have 1000 records with equal primary keys, we only need to store the primary key values once.

The downside of this is that the storage of each object involves an additional hash. However, this should be fairly inexpensive and is well worth it for the decreased memory usage.

stevenewald:
Still brainstorming method of fixing hash collisions without a ton of added complexity/indirection

Current idea: open addressing with linear probing, maintain 4 mappings:

Hashed ordinal -> assigned ordinal (OA, LP)
Hashed field -> IntList of ordinals (Java HashMap)
Hashed ordinal -> memoized field objects (Java HashMap)
Ordinal -> hashed ordinal (Java Hashmap, designed for ordinals to be overwritten on new single/double snapshot)

PR #624, opened 2023-06-22
Enhanced HollowHistoryTypeKeyIndex with Optimized Mappings and Memoization

stevenewald:
This PR introduces a major optimization to the HollowHistoryTypeKeyIndex to significantly reduce unnecessary state update round trips. Consequently, performance and memory management are substantially enhanced.

**Architectural Overview:**

Leveraging the **`HollowOrdinalMapper`**, the architecture now utilizes open-addressed hash tables with linear probing to manage associations such as:
1. Mappings from hashed records to assigned ordinals.
2. Associations of hashed records with int[] of field object ordinals.
3. Mapping hashed fields to their assigned ordinals.
4. Assigned ordinals to their corresponding locations in arrays from points 1 & 2.

**Key Benefits:**

1. **Streamlined Updates:** State mappings now instantly reflect the latest typeState, eliminating the need for round trips.
2. **Immediate Storage & Hashing:** Keys are hashed and stored directly upon updates, removing previous layers of indirection.
3. **Reduced Memory Footprint:** Instead of individual POJOs for each record in every state, we now leverage a **`ByteArrayOrdinalMap`** to store distinct field objects. We then refer to these objects using their object ordinals.
4. **Enhanced Query Speed:** Direct hash table lookups now speed up querying, making readStateEngine traversals obsolete.

**Supporting Classes Introduced:**

- **HollowOrdinalMapper:** This class is dedicated to hashing and storage tasks for records and their fields. By decoupling from **`HollowHistoryTypeKeyIndex`**, the latter now solely focuses on state update functions.
- **ObjectInternPool:** This utility manages the storage of unique field objects across states. Each distinct object is stored once and assigned an integer ordinal, implementing extremely memory efficient memoization via a variable-length packed ByteOrdinalArray. This process facilitates HollowOrdinalMapper in storing just an integer per field in the hash table, optimizing memory use.

**Highlights & Impact:**

- **Code Conciseness:** The **`HollowHistoryTypeKeyIndex`** has been streamlined from ~550 lines to roughly 225 lines, improving readability.
- **Performance Boost:** Preliminary tests suggest that history processing speed has increased 3-4 times. However, when considering dataset processing as a whole (which includes several other functions), the speed improvement is about 30-50%.
- **Memory Efficiency:** Compared to the previous approach, which kept a full read/write state, ordinal mapping, hash mapping, and did not utilize memoization, the new implementation showcased a 12% memory usage reduction in tests.

## Profiler screenshots to illustrate the performance improvements
Note the stack percentages in the bottom left
### Old implementation
Old implementation
### New implementation
New implementation

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowHistoryTypeKeyIndex.java
+++ b/hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowHistoryTypeKeyIndex.java
@@ -18,485 +18,284 @@ package com.netflix.hollow.tools.history.keyindex;
 
 import static com.netflix.hollow.core.HollowConstants.ORDINAL_NONE;
 import static com.netflix.hollow.tools.util.SearchUtils.MULTI_FIELD_KEY_DELIMITER;
-import static com.netflix.hollow.tools.util.SearchUtils.getFieldPathIndexes;
-import static com.netflix.hollow.tools.util.SearchUtils.getOrdinalToDisplay;
-import static com.netflix.hollow.tools.util.SearchUtils.parseKey;
 
 import com.netflix.hollow.core.HollowDataset;
 import com.netflix.hollow.core.index.key.PrimaryKey;
 import com.netflix.hollow.core.memory.encoding.HashCodes;
 import com.netflix.hollow.core.read.HollowReadFieldUtils;
-import com.netflix.hollow.core.read.engine.HollowReadStateEngine;
 import com.netflix.hollow.core.read.engine.PopulatedOrdinalListener;
 import com.netflix.hollow.core.read.engine.object.HollowObjectTypeReadState;
 import com.netflix.hollow.core.schema.HollowObjectSchema;
+import com.netflix.hollow.core.schema.HollowObjectSchema.FieldType;
 import com.netflix.hollow.core.util.IntList;
-import com.netflix.hollow.core.util.LongList;
 import com.netflix.hollow.core.util.RemovedOrdinalIterator;
-import com.netflix.hollow.core.write.HollowObjectTypeWriteState;
-import com.netflix.hollow.core.write.HollowObjectWriteRecord;
-import com.netflix.hollow.core.write.HollowTypeWriteState;
-import com.netflix.hollow.core.write.HollowWriteStateEngine;
+import com.netflix.hollow.tools.util.ObjectInternPool;
 import java.util.Arrays;
 import java.util.BitSet;
+import java.util.HashMap;
+
 
 public class HollowHistoryTypeKeyIndex {
-
     private final PrimaryKey primaryKey;
-    private final String[][] keyFieldParts;
+    private final FieldType[] fieldTypes;
+
     private final boolean[] keyFieldIsIndexed;
-    private HollowObjectSchema keySchema;
+    private final String[][] keyFieldParts;
+    private final int[][] keyFieldIndices;
 
-    private int[] hashedRecordKeys;
-    private int[][] hashedFieldKeys;
-    private LongList hashedFieldKeyChains;
-
-    private int maxIndexedKeyOrdinal = 0;
-
-    private final HollowWriteStateEngine writeStateEngine;
-    private HollowReadStateEngine readStateEngine;
     private boolean isInitialized = false;
+    private int maxIndexedOrdinal = 0;
 
-    public HollowHistoryTypeKeyIndex(PrimaryKey primaryKey, HollowDataset dataModel, HollowWriteStateEngine writeEngine, HollowReadStateEngine readEngine) {
+
+    // Maps a hashed ordinal from within the original read state to an assigned ordinal
+    private final HashMap<Integer, Integer> ordinalMapping;
+    private final HashMap<Integer, IntList> ordinalFieldHashMapping;
+    private final HashMap<Integer, Object[]> ordinalFieldObjectMapping;
+
+    public HollowHistoryTypeKeyIndex(PrimaryKey primaryKey, HollowDataset dataModel) {
         this.primaryKey = primaryKey;
-        this.writeStateEngine = writeEngine;
-        this.readStateEngine = readEngine;
-        this.keyFieldParts = getKeyFieldParts(dataModel);
         this.keyFieldIsIndexed = new boolean[primaryKey.numFields()];
-    }
-
-    public void addFieldIndex(String fieldName, HollowDataset dataModel) {
-        String[] fieldPathParts = PrimaryKey.getCompleteFieldPathParts(dataModel, primaryKey.getType(), fieldName);
-        
+        this.keyFieldParts = getKeyFieldParts(dataModel);
+        this.keyFieldIndices = new int[primaryKey.numFields()][];
+        this.fieldTypes = new FieldType[primaryKey.numFields()];
         for(int i=0;i<primaryKey.numFields();i++) {
-            String[] pkFieldPathParts = PrimaryKey.getCompleteFieldPathParts(dataModel, primaryKey.getType(), primaryKey.getFieldPath(i));
-            if(Arrays.equals(pkFieldPathParts, fieldPathParts)) {
-                keyFieldIsIndexed[i] = true;
-                break;
-            }
+            this.keyFieldIndices[i] = PrimaryKey.getFieldPathIndex(dataModel, primaryKey.getType(), primaryKey.getFieldPath(i));
         }
+
+        this.ordinalMapping = new HashMap<>();
+        this.ordinalFieldHashMapping = new HashMap<>();
+        this.ordinalFieldObjectMapping = new HashMap<>();
     }
 
     public boolean isInitialized() {
         return isInitialized;
     }
 
-    public void initialize(HollowObjectTypeReadState initialTypeState) {
-        if (isInitialized) return;
+    public int findKeyIndexOrdinal(HollowObjectTypeReadState typeState, int ordinal) {
+        int hashedRecord = hashKeyRecord(typeState, ordinal);
+        if(!ordinalMapping.containsKey(hashedRecord)) {
+            return ORDINAL_NONE;
+        }
+        return ordinalMapping.get(hashedRecord);
+    }
 
-        initKeySchema(initialTypeState.getSchema());
-        initializeTypeWriteState();
+    public void initializeKeySchema(HollowObjectTypeReadState initialTypeState) {
+        if (isInitialized) return;
+        HollowObjectSchema schema = initialTypeState.getSchema();
+
+        for (String[] keyFieldPart : keyFieldParts) addSchemaField(schema, keyFieldPart, 0);
         isInitialized = true;
     }
 
-    public void updateReadStateEngine(HollowReadStateEngine readEngine) {
-        readStateEngine=readEngine;
+    private void addSchemaField(HollowObjectSchema schema, String[] keyFieldParts, int keyFieldPartPosition) {
+        int schemaPosition = schema.getPosition(keyFieldParts[keyFieldPartPosition]);
+        if (keyFieldPartPosition < keyFieldParts.length - 1) {
+            HollowObjectSchema nextPartSchema = (HollowObjectSchema) schema.getReferencedTypeState(schemaPosition).getSchema();
+            addSchemaField(nextPartSchema, keyFieldParts, keyFieldPartPosition + 1);
+        } else {
+            fieldTypes[keyFieldPartPosition] = schema.getFieldType(schemaPosition);
+        }
+    }
+
+    private String[][] getKeyFieldParts(HollowDataset dataModel) {
+        String[][] keyFieldParts = new String[primaryKey.numFields()][];
+        for (int i = 0; i < primaryKey.numFields(); i++)
+            keyFieldParts[i] = PrimaryKey.getCompleteFieldPathParts(dataModel, primaryKey.getType(), primaryKey.getFieldPath(i));
+        return keyFieldParts;
+    }
+
+    public void addFieldIndex(String fieldName, HollowDataset dataModel) {
+        String[] fieldPathParts = PrimaryKey.getCompleteFieldPathParts(dataModel, primaryKey.getType(), fieldName);
+        for (int i = 0; i < primaryKey.numFields(); i++) {
+            String[] pkFieldPathParts = PrimaryKey.getCompleteFieldPathParts(dataModel, primaryKey.getType(), primaryKey.getFieldPath(i));
+            if (Arrays.equals(pkFieldPathParts, fieldPathParts)) {
+                keyFieldIsIndexed[i] = true;
+                break;
+            }
+        }
     }
 
     public void update(HollowObjectTypeReadState latestTypeState, boolean isDeltaAndIndexInitialized) {
-        // copies over keys corresponding to previously populated ordinals AND currently populated ordinals
-        copyExistingKeys(); // only populates the ordianl map (to be the OR of previous and current ordinals)
         if (latestTypeState == null) return;
 
         if (isDeltaAndIndexInitialized) {
-            // record all newly seen keys when going from v1->v2 or v2->v1 to the history type key index
             populateNewCurrentRecordKeysIntoIndex(latestTypeState);
         } else {
-            // usually on the first fwd delta transition, or on  double snapshot
+            maxIndexedOrdinal = 0;
             populateAllCurrentRecordKeysIntoIndex(latestTypeState);
         }
     }
 
-    public void hashRecordKeys() {
-        HollowObjectTypeReadState keyTypeState = (HollowObjectTypeReadState) readStateEngine.getTypeState(primaryKey.getType());
-        if (keyTypeState == null) return;
-
-        int hashTableSize = HashCodes.hashTableSize(keyTypeState.maxOrdinal() + 1);
-
-        if(hashedRecordKeys == null || hashedRecordKeys.length < hashTableSize) {
-            rehashAllRecordKeys(keyTypeState, hashTableSize);
-        } else {
-            hashNewRecordKeys(keyTypeState);
-        }
-    }
-
-    private void hashNewRecordKeys(HollowObjectTypeReadState keyTypeState) {
-        for(int i=maxIndexedKeyOrdinal+1;i<=keyTypeState.maxOrdinal();i++)
-            indexOrdinal(keyTypeState, i, hashedRecordKeys, hashedFieldKeys, hashedFieldKeyChains);
-        maxIndexedKeyOrdinal = keyTypeState.maxOrdinal();
-    }
-
-    private void rehashAllRecordKeys(HollowObjectTypeReadState keyTypeState, int hashTableSize) {
-        int[] hashedRecordKeys = initializeHashedKeyArray(hashTableSize);
-        int[][] hashedFieldKeys = new int[primaryKey.numFields()][];
-        LongList hashedFieldKeyChains = new LongList();
-
-        for(int i=0;i<primaryKey.numFields();i++)
-            if(keyFieldIsIndexed[i])
-                hashedFieldKeys[i] = initializeHashedKeyArray(hashTableSize);
-
-        for(int i=0;i<=keyTypeState.maxOrdinal();i++)
-            indexOrdinal(keyTypeState, i, hashedRecordKeys, hashedFieldKeys, hashedFieldKeyChains);
-
-        this.hashedRecordKeys = hashedRecordKeys;
-        this.hashedFieldKeys = hashedFieldKeys;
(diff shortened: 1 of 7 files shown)


Sunjeet:
Hi Steve, thanks for putting this together. I'll review today.

hollow/src/main/java/com/netflix/hollow/tools/util/ObjectInternPool.java
```diff
@@ -0,0 +1,68 @@
...
+
+class GenericInternPool<T> {
+    private final Map<T, T> pool = new HashMap<>();
+
+    public T intern(T object) {
+        if (object == null) {
+            throw new IllegalArgumentException("Cannot intern null objects");
+        }
+
+        synchronized (pool) {
```
Sunjeet: I tried to sample the results with one of our ~1G datasets.. I didn't get a performance improvement in that, I suspect it may be due to this synchronization. This could be optimized e.g. do an optimistic get, lock stripping. Or just use ConcurrentHashMap? I dont think there would be much more memory overhead.

There was a little bug that came up in the primary key field parsing. The primary key referenced one field, like "id.value", this patch should fix it-
```
--- a/hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowHistoryTypeKeyIndex.java
+++ b/hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowHistoryTypeKeyIndex.java
@@ -93,18 +93,20 @@ public class HollowHistoryTypeKeyIndex {
         if (isInitialized) return;
         HollowObjectSchema schema = initialTypeState.getSchema();
 
-        for (String[] keyFieldPart : keyFieldNames) addSchemaField(schema, keyFieldPart, 0);
+        for (int i= 0; i < keyFieldNames.length; i ++) {
+            String[] keyFieldPart = keyFieldNames[i];
+            fieldTypes[i] = addSchemaField(schema, keyFieldPart, 0);
+        }
         isInitialized = true;
     }
 
-    private void addSchemaField(HollowObjectSchema schema, String[] keyFieldNames, int keyFieldPartPosition) {
+    private FieldType addSchemaField(HollowObjectSchema schema, String[] keyFieldNames, int keyFieldPartPosition) {
         int schemaPosition = schema.getPosition(keyFieldNames[keyFieldPartPosition]);
         if (keyFieldPartPosition < keyFieldNames.length - 1) {
             HollowObjectSchema nextPartSchema = (HollowObjectSchema) schema.getReferencedTypeState(schemaPosition).getSchema();
-            addSchemaField(nextPartSchema, keyFieldNames, keyFieldPartPosition + 1);
-        } else {
-            fieldTypes[keyFieldPartPosition] = schema.getFieldType(schemaPosition);
+            return addSchemaField(nextPartSchema, keyFieldNames, keyFieldPartPosition + 1);
         }
+        return schema.getFieldType(schemaPosition);
     }

```

The memory consumption optimization from interning everything has an interesting aspect: for datasets that don't have much repetition in Integer/Long/Float/Double values the overhead of interning adds up.. I think 16-32 bytes per object that isn't repeated. For that reason I wonder if it's possible to reuse Hollow's interning specifics without actually incurring the whole write state to read state round trip, i.e. a zero-copy population of the read state from write state. I haven't looked into how feasible that is, but I wonder if it's something you've already looked into and passed over?

stevenewald: > I tried to sample the results with one of our ~1G datasets.. I didn't get a performance improvement in that, I suspect it may be due to this synchronization. This could be optimized e.g. do an optimistic get, lock stripping. Or just use ConcurrentHashMap? I dont think there would be much more memory overhead.

I'm not sure if synchronization is the cause of the lower performance as each type index has its own pool, and I believe there is no more than one thread operating on a type index at a time. With that said, I didn't run a profiler after the addition of the intern pool, so it may still be the cause.

Can you elaborate on how you measured the performance improvement, as well as if the dataset is publicly available?

> There was a little bug that came up in the primary key field parsing. The primary key referenced one field, like "id.value", this patch should fix it-

Good catch! I'll add that to the PR. 


> The memory consumption optimization from interning everything has an interesting aspect: for datasets that don't have much repetition in Integer/Long/Float/Double values the overhead of interning adds up.. I think 16-32 bytes per object that isn't repeated.

Can you explain why this is the case? My idea was that Java will store only one instance of an object, as the hashmap in `HollowOrdinalMapper` will contain a reference to the object stored in the memoized pool. Is the 16-32 bytes due to the memory overhead of the hashmap per-object? If so, this may be hard to avoid assuming we want to memoize all objects (which is what Drew suggested on our call). If the previous Hollow memoization implementation avoids this per-object overhead, please let me know.

Thanks for taking a look!

Sunjeet: Very rudimentary testing. Just spinning up history server for an internal dataset that has a fair amount of complexity in data model and size, on my laptop using pre-downloaded hollow blobs. I was looking at how long it took to process deltas and it was comparable across the old and new code (around 15s to process 7 deltas), but let me actually setup a more legit test and confirm that.. hopefully a bit later today.  Thanks for clarifying the bit about synchronization.

Sunjeet: Yes re: the 16-32 bytes overhead, I was thinking of the references in HashMap. HashMap.Node looks like 
```
        final int hash;
        final K key;
        V value;
        Node<K,V> next;
```
This cost might be bearable and like you said unavoidable if we memoize objects.. We may need to add reference counting and drop objects from the intern pool when old history states are removed to avoid memory creep for long running history servers. 

>If the previous Hollow memoization implementation avoids this per-object overhead, please let me know.

I guess it's not really interning/memoizing(sorry for my poor use of that term), it's deduping field values within the specific historic state by adding records to a write state engine- the  memory overhead for storing those values is expected to be lower because of use of ordinals and variable length encoding into byte arrays, but we dont dedupe values across versions so the overall memory utilization may be higher if there is repetition in values across states. So theres that tradeoff there.. I wasn't aware you already chatted with Drew- if he suggested memoization lets go ahead and do that- I can also compare the memory footprint when comparing performance to validate with real world data.

stevenewald: Thanks for the explanation. Comparing the memory usage between the old and new implementation with real-world data would definitely be helpful. For your testing, I noticed the biggest performance(cpu) benefit when using many deltas/snapshots. Drew said the biggest pain point with history processing is when there's many deltas/snapshots, so this is good news.

I think dropping objects with reference counts is also a good idea to reduce memory usage, however I have two main concerns:

1. If we store reference counts, that will require a int to be stored. Assuming we allocated efficiently with an array (vs in a class, which may be susceptible to 8-byte alignment), that's an extra 4 bytes off the bat.
2. Implementing a policy and mechanism for removing these non-duplicate values would likely have its own performance (primarily cpu) implication.

However, if interning with hashmaps turns out to add a lot of memory overhead, we may have to go this route or consider not interning entirely.

Using a write state engine was ruled out because of the indirection/modification necessary to avoid a round trip on each new state.

Sunjeet: +1 the primary benefit of faster snapshot/delta processing time outweighs the desire to optimize memory util. In production we dont care about squeezing as many historic states as possible, rather we need to be able to look back X no. of states per dataset where X is different for different datasets. So optimal memory util translates to slightly less RAM costs, but I dont think the difference between the 2 approaches is that drastic, esp. considering it depends on nature of changes in the underlying data.

That said, my hunch is that an ever-growing interning pool could pose a problem for running these history servers in a stable manner- but let me see with data what that creep actually looks like, and if that extra bookkeeping could be valuable.

Sunjeet: An update on testing, more to follow- 

Delta processing performance improvement is less noticeable (9/10th of baseline) with a dataset thats 1GB, and much more noticeable (1/3rd of baseline) for a dataset thats 11GB. Refer images perf1gb and perf11gb respectively, the y-axis is consumer refresh time in seconds.

perf1gb
perf11gb

For heap usage creep, refer image heap1- heap grew by about 34g for 1 month's history for that 11gb sized dataset- i expect growth starts to taper as the object intern pool size grows, but still, unchecked growth does seem less ideal (as compared to additional bookkeeping to be able to reclaim space). I have some more thoughts there, we should sync up before we make any changes. 
heap1

Related to heap usage, what are your thoughts on making the object intern pool a singleton, as opposed to the current per-type index? I suspect the motivation was to minimize synchronization between per-type threads, and I wonder if a singleton object intern pool that uses ConcurrentHashMaps could give us comparable performance at less memory util.  Given any difference there will likely be negligible in the overall delta refresh time, since the majority of time (now with your changes) is spent consumer delta state processing outside of the history computation. 

Functionally, I'm seeing a little regression that I haven't been able to explain yet- there seems to be an edge case where sometimes a primary key defined for a type isn't indexed correctly or at least displayed correctly in the view (refer image edgecase1)

edgecase1

Also noticed that the aggregated diff count per type (and specifically no. of modifications and removals) reported by the original vs new code paths is sometimes different. I will look more into this as well, at first glance it seems to be race condition-ey, but I may be wrong. It could also be related to the other issue I mentioned above. Will look more today.

Sunjeet: [This tweak](https://github.com/Netflix/hollow/commit/488f9c5470c12753e153e4e474a7ce2c637902c5) for handling the edge case where some fields appeared as "null"- to preserve behavior where `addTypeIndex` could be called called additively

stevenewald: Thanks for the data, I have a few thoughts on this. I believe Drew mentioned that the current implementation of `HollowHistory` uses a series of deltas, rather than snapshots. If this is the case, it'll be hard to prevent unrestricted heap growth because all of the field objects for ordinals in the state need to be accessible.

I'm not particularly concerned with the fact that the heap is growing over time, this should be expected as we need to be able to retrieve the values of all historical states. I may be misunderstanding this, though, especially if snapshots are used (note: I'm referring to snapshots within HollowHistory, not when processing the actual dataset). Also, Drew recommended the use of the `intern()` function, which keeps strings in the heap for the lifetime of the program.

Do you have any thoughts on this? I'm less concerned about the amount of growth (as this can be optimized), and more concerned about if the heap creep is expected.

hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowOrdinalMapper.java
```diff
@@ -0,0 +1,196 @@
...
+
+    public Object getFieldObject(int keyOrdinal, int fieldIndex) {
+        int index = assignedOrdinalToIndex.get(keyOrdinal);
+        return indexFieldObjectMapping.get(index)[fieldIndex];
+    }
+
+    private int hashKeyRecord(HollowObjectTypeReadState typeState, int ordinal) {
+        int hashCode = 0;
+        for (int i = 0; i < primaryKey.numFields(); i++) {
+            int fieldHashCode = HollowReadFieldUtils.fieldHashCode(typeState, ordinal, i);
```
Sunjeet: I think there is a regression here for when the field at index i is of REFERENCE type. In that case, the hash is computed based on the value of the ordinal into the referenced type, and over time that ordinal could be reused for a different value, resulting in hashes not being consistently computed. E.g. data model-

```
Movie @PrimaryKey(id) {
    Video id;
    String name;
}

Video {
    int value;
}
```

I think this explains the corruption type behavior I was seeing where the diff counts didn't match with baseline. I've attempted a patch in [this commit](https://github.com/Netflix/hollow/commit/43db4ba9f20b3135fe6f151e70422b715629a669) but haven't tested it yet

hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowHistoryTypeKeyIndex.java
```diff
@@ -18,485 +18,243 @@
...
-            ordinal = iter.next();
-        }
-    }
-
-    public String[] getKeyFields() {
-        return primaryKey.getFieldPaths();
+    public void addMatches(int hashCode, IntList results) {
+        if (!ordinalFieldHashMapping.containsKey(hashCode))
+            return;
+        IntList res2 = ordinalFieldHashMapping.get(hashCode);
```
Sunjeet: A little bug here in that the query returns results based on hash match rather than value match. This causes the search behavior to sometimes return garbage results.

hollow/src/main/java/com/netflix/hollow/tools/util/ObjectInternPool.java
Sunjeet: It's true our implementation relies on listening to delta updates. But also see `HollowHistory::removeHistoricalStates`. Basically, once sufficient no. of historic states have been indexed you can enter a "sliding window" mode where oldest states are evicted in order to accommodate new versions thereby keeping the heap util somewhat stable. It is desirable to have x versions / days of history available at any given time, as opposed of having to wait for a newly initialized history server to catch up (which could be hours for large datasets with high entropy).

hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowOrdinalMapper.java
```diff
@@ -0,0 +1,196 @@
...
+
+import static com.netflix.hollow.core.HollowConstants.ORDINAL_NONE;
+import java.util.Arrays;
+import java.util.HashMap;
+
+public class HollowOrdinalMapper {
+    private int size = 0;
+    private static final double LOAD_FACTOR = 0.7;
+    
+    private Integer[] ordinalMappings;
```
Sunjeet: Indexing into an array will be limiting. For e.g. some datasets churn ~5 million modification+removals a day. But maybe ok for the scope of this PR, and something we can scale in future. Or we could consider reusing Hollow's `FixedLengthElementArray` here- it's indexed into using long and it would also be more compact.

stevenewald:
Can you expand on why indexing into an array will be limiting? I’m not sure
I’m on the same page because the previous implementation also used integer
arrays for ordinal mappings.

hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowOrdinalMapper.java
stevenewald: Can you expand on why indexing into an array will be limiting? I’m not sure
I’m on the same page because the previous implementation also used integer
arrays for ordinal mappings.

Drew specifically suggested I use OA/LP for the ordinal mappings

hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowOrdinalMapper.java
stevenewald: I see, I fixed this in other places but missed it here. Good catch! I think I already have a function for this, will LYK

hollow/src/main/java/com/netflix/hollow/tools/util/ObjectInternPool.java
stevenewald: I see, would it be worth having a 15-20min call to discuss these changes? I've received different suggestions from you and Drew (who said that ordinals in the mapping should remain forever), so I think it may be helpful for us to sync and get on the same page.

Sunjeet: Yes, it would be good to sync up. Would Wed or Thu after 2 PM PST work? Is the email listed on your github profile a good one to send the meeting details out to?

stevenewald: Yes, that works well.

hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowOrdinalMapper.java
```diff
@@ -0,0 +1,196 @@
...
+        if ((double) size / ordinalMappings.length > LOAD_FACTOR) {
+            expandAndRehashTable();
+        }
+
+        int index = indexFromHash(hashedRecord, ordinalMappings.length);
+
+        // Linear probing
+        while (ordinalMappings[index] != ORDINAL_NONE) {
+            if(recordsAreEqual(typeState, ordinal, index)) {
+                this.assignedOrdinalToIndex.put(assignedOrdinal, index);
```
Sunjeet: Multiple `assignedOrdinal`s can map to the same index. But when `expandAndRehashTable` calls `rehashExistingRecord` it only updates one of these assignedOrdinals to point to map to the new index but some entries in `assignedOrdinalToIndex` will still be pointing to the non-remapped index value. Am I reading this right?

hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowOrdinalMapper.java
Sunjeet: Ah yes I missed that. I was worried about exhausting the Integer.MAX_VALUE (~2B) limit for size of an array in java, which I think translates to how many changes can be indexed for a type.
But actually, as you said that limit was already in place, and in fact since the old impl used a state engine per type key index there was a lower limit of ~500M (2^29 is the max ordinals limit).
So I think add another win to the list, this PR has 4X'ed that limit.

hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowOrdinalMapper.java
stevenewald: With deltas, there should never be repeated `assignedOrdinals`, so in practice, this will not happen.

However, I realized that we don't empty the `ordinalMappings` table when there's a snapshot. This won't cause any errors as we still overwrite the key in `assignedOrdinalToIndex`, but it will clog up the hash table because we aren't removing values that will never be retrieved again. Should be a one-line fix that I'll implement later, good catch!

hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowOrdinalMapper.java
stevenewald: Interesting, I discussed this exact issue with Drew and he suggested using ordinals (as opposed to values) because the records are immutable. However, I'm not sure if he considered ordinals being reused, and your theory makes sense to me.

I already have a function elsewhere that retrieves the referenced value, I'll implement that to make it a bit cleaner.

stevenewald:
Update for anyone tracking this - met with @Sunjeet, will have requested changes ready shortly

hollow/src/main/java/com/netflix/hollow/tools/util/ObjectInternPool.java
stevenewald: Resolving due to discussing on call and deciding that further tests will be done. We can followup later on if more discussion is needed.

hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowOrdinalMapper.java
stevenewald: Fixed in 51cc570d

hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowOrdinalMapper.java
stevenewald: Fixed in 1e522166

hollow/src/main/java/com/netflix/hollow/tools/history/keyindex/HollowHistoryTypeKeyIndex.java
stevenewald: Fixed in faad5249

Sunjeet:
The latest changes look good. Test has been running for 5 days on our most complex dataset, the resulting history views look good and performance improvements are maintained. I'm now confident about merging this in.

The only remaining concern is with corruption happening when double snapshots are incurred. I dont see that as blocking this PR merge, since double snapshots are believed to be buggy in the original impl. But this will be followed by a change to fail hard when History sees an incoming double snapshot- to avoid silent corruption. 
Double snapshots would be useful for handling schema changes dynamically. The path fwd for handling schema changes will be to be able to apply them as a part of delta updates (TODO for core hollow), and the interim workaround is to initialize a new instance of `HollowHistory` and discard the old one. 

To summarize the improvements attained in this PR, 

* heap size is the same as before
* cpu util is about 1/3rd as compared to before
* consumer refreshes are about 2.5x faster (down from ~50s to ~20s)

Thank you for the contribution @stevenewald

PR #631, opened 2023-07-19
Make creator serializable

snallapa:
Creator is instantiated with lambdas. We have been using hollow in pipeline environments and so being serializable by either the Java Serializer or others (like Kryo) need lambdas to be serializable as well. I am not sure if this change has any effect in hollow itself so feel free to reject it, but this would help my experience with Hollow right now


snallapa:
@Sunjeet @stevenewald i was wondering if either of you could take a look at this?

PR #634, opened 2023-08-11
generate performance api directory matching package

### LINKED ISSUE #633: PerformanceAPI directory not matching package
The generated files by the HollowPerformanceAPIGenerator is placed in a directory not matching the package. 
This behaviour differs from the [HollowAPIGenerator](https://github.com/Netflix/hollow/blob/1c4f393f3bb4473355d0f94239e960effcd42d1c/hollow/src/main/java/com/netflix/hollow/api/codegen/HollowAPIGenerator.java#L418-L421).

viqtor:
resolves #633


viqtor:
i took the liberty of adding an exclusion in spotbugs for the `fieldNames`  issue it raises

Sunjeet (approved):

Sunjeet:
There was an assumption here that caller would pass api dir and package params that agree with each other, but your change looks like a nice usability addition- basically if they dont agree then assume that the api dir is the root.

PR #637, opened 2023-08-20
propagate errors during async refresh

### LINKED ISSUE #636: errors during consumer async refresh are hidden
Potential exceptions thrown during a consumers async refresh are hidden for an outside caller.

This can cause issues seen in the [hollow-reference-implementation](https://github.com/Netflix/hollow-reference-implementation/blob/a2974a8fef35d806d1b735b185ebc5bfba2cc1d1/src/main/java/how/hollow/consumer/infrastructure/S3AnnouncementWatcher.java#L59-L61) where a AnnouncementWatcher advances its local version without knowing that the consumer is not keeping up. 
In our implementation this has been resolved when the next snapshot has been produced and announced.

viqtor:
resolves #636

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
@@ -244,8 +244,8 @@ public class HollowConsumer {
      * <p>
      * This is an asynchronous call.
      */
-    public void triggerAsyncRefresh() {
-        triggerAsyncRefreshWithDelay(0);
+    public CompletableFuture<Void> triggerAsyncRefresh() {
+        return triggerAsyncRefreshWithDelay(0);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Sunjeet:
Thanks for this contribution. Changing the return type of the `triggerAsyncRefresh` method from `void` to `CompletableFuture<Void>` would cause binary backwards incompatibility. Is there a reason to not simply use the synchronous flavor `triggerRefresh`/`triggerRefreshTo` in your application?

viqtor:
>  Is there a reason to not simply use the synchronous flavor `triggerRefresh`/`triggerRefreshTo` in your application?

We have been following the [reference implementation](https://github.com/Netflix/hollow-reference-implementation/blob/a2974a8fef35d806d1b735b185ebc5bfba2cc1d1/src/main/java/how/hollow/consumer/infrastructure/S3AnnouncementWatcher.java#L59-L61). We have since switched to the sync alternative since dicsovering this issue but it may lead to false assumptions when you pass in a `refreshExectutor` to your consumer that isn't used.

I just noticed that the same method is also used during the [getInitialLoad()](https://github.com/Netflix/hollow/blob/master/hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java#L349-L353) which wont trip the catch block.

Would an alternative solution be to deprecate these methods and add new ones with the new return type?

PR #638, opened 2023-09-07
add VarInt helpers to read and write against primitive byte arrays

eduardoramirez:


hollow/src/main/java/com/netflix/hollow/core/memory/encoding/VarInt.java
```diff
@@ -83,6 +95,30 @@ public static void writeVLong(OutputStream out, long value) throws IOException {
         out.write((byte)(value & 0x7FL));
     }
 
+    /**
+     * Encode the specified long as a variable length integer into the supplied byte array.
```
Sunjeet: typo: variable length long

Sunjeet (approved):

hollow/src/main/java/com/netflix/hollow/core/memory/encoding/VarInt.java
eduardoramirez:
```suggestion
     * Encode the specified long as a variable length long into the supplied byte array.
```

PR #639, opened 2023-09-18
recordFieldAccess introduces a delay in tight loops

jleibund:
avoids the field array access when the director is disabled (this is the default).. skips the shortcut when setting up field-level sampling.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/sampling/HollowObjectSampler.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/sampling/HollowObjectSampler.java
@@ -33,10 +33,12 @@ public class HollowObjectSampler implements HollowSampler {
     private final String fieldNames[];
     private final long sampleCounts[];
     private final HollowSamplingDirector samplingDirectors[];
+    private boolean isSamplingDisabled;
 
     public HollowObjectSampler(HollowObjectSchema schema, HollowSamplingDirector director) {
         this.typeName = schema.getName();
         this.sampleCounts = new long[schema.numFields()];
+        this.isSamplingDisabled = director == DisabledSamplingDirector.INSTANCE;
         HollowSamplingDirector[] samplingDirectors = new HollowSamplingDirector[schema.numFields()];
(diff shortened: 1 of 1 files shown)


hollow/src/main/java/com/netflix/hollow/api/sampling/HollowObjectSampler.java
```diff
@@ -60,6 +63,7 @@ public void setFieldSpecificSamplingDirector(HollowFilterConfig fieldSpec, Hollo
 
         for(int i=0;i<fieldNames.length;i++) {
             if(typeConfig.includesField(fieldNames[i])) {
+                this.isSamplingDisabled = false;
```
nayanika-u, line 66: Shouldn't this field value depend on the director instance type like director == DisabledSamplingDirector.INSTANCE. Why is sampling being disabled by default for this field specific case?

jleibund: Sampling is not being disabled on line 66..  the shortcut to skip sampling without an array index is being disabled.  This boolean does not control sampling.  Its a perf gain when all fields are disabled which is the default.

isSamplingDisabled is a shortcut that only helps when every field is disabled because we skip the array lookup and method call which, though seemingly inexpensive, shows up as a quantifiable delay in YourKit with large data sets processed in a tight loop.  

Typically its rare to call setFieldSpecificSamplingDirector.  There no invocations of it in this repo and there are none I can find in sourcegraph.  This appears to be feature introduced seven years ago that hasn't seen much use but is a net tax on performance.  It is almost universally disabled.

I think there is diminishing value in setting the shortcut here to anything but false because in the rare case that a consumer calls setFieldSpecificSamplingDirector() they intend to set it to something other than disabled - otherwise why call it?  Worst case is they don't get the shortcut but enabled sampling is a net drag anyway, right (incrementing counters)?  My opinion is simplicity is better.

nayanika-u (approved):

PR #642, opened 2023-10-04
HollowObjectTypeDataElements utilities for splitting and joining

Sunjeet:
This is 1 of 4 planned PRs for supporting dynamic type re-sharding:

- [X] utilities for splitting/joining Object types; limited to Object types in this step, extend to Map/Set/List in last step.
- [ ] consumer delta application reshards with O(shard size) extra space.
- [ ] producer-side numShards toggle and successful integrity check verification of delta/revdelta application.
- [ ] Extend resharding to Map,List, and Set types.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectDeltaHistoricalStateCreator.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectDeltaHistoricalStateCreator.java
@@ -17,6 +17,8 @@
 package com.netflix.hollow.core.read.engine.object;
 
 import static com.netflix.hollow.core.HollowConstants.ORDINAL_NONE;
+import static com.netflix.hollow.core.read.engine.object.HollowObjectTypeDataElements.copyRecord;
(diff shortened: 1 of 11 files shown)


hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectDeltaHistoricalStateCreator.java
```diff
@@ -125,66 +133,11 @@ private void populateStats() {
                 historicalDataElements.bitsPerField[i] = (64 - Long.numberOfLeadingZeros(totalVarLengthSizes[i] + 1)) + 1;
             }
 
-            historicalDataElements.nullValueForField[i] = (1L << historicalDataElements.bitsPerField[i]) - 1;
+            historicalDataElements.nullValueForField[i] = historicalDataElements.bitsPerField[i] == 64 ? -1L : (1L << historicalDataElements.bitsPerField[i]) - 1;
```
Sunjeet, line 136: Note unrelated bugfix- double value `0.0` would show up as `NaN`

PR #643, opened 2023-10-11
HistoricalStateCreator- stricter check for var len type against type state schema

Sunjeet:
For the edge case: if shard 0 contains no ordinals it will have null varLengthData for a variable length data field. This will can cause refresh failures or corruption since other shards may have non-zero no. of ordinals for variable len field


Sunjeet:
Fixed as part of resharding work here https://github.com/Netflix/hollow/commit/ec5dcb4fe45385ddac8c906a4ee0a2debe9ed6f9

PR #644, opened 2023-10-11
Consumer delta application supports re-sharding for Object type

Sunjeet:
This is 2 of 4 planned PRs for supporting dynamic type re-sharding:

- [X] [PR#642](https://github.com/Netflix/hollow/pull/642) <s>utilities for splitting/joining Object types; limited to Object types in this step, extend to Map/Set/List in last step</s> 
- [X] consumer delta application reshards with O(shard size) extra space (this PR) 
- [ ] producer-side numShards toggle and signaling
- [ ] Extend resharding to Map,List, and Set types

### Benchmarking with jmh
Benchmark 1: No significant regression for consumers when resharding isn't invoked, avg is upto 6% worse [`Baseline.json` vs. `PR.json`] 

Benchmark 2: Read performance during delta transition with resharding vs. without resharding:  avg is upto 6% worse again [`PR_withDeltaTransitionsNoResharding.json` vs. `PR_withDeltaTransitionsWithResharding.json`]

[Baseline.json](https://github.com/Netflix/hollow/files/13090264/Baseline.json)
[PR.json](https://github.com/Netflix/hollow/files/13090267/PR.json)
[PR_withDeltaTransitionsNoResharding.json](https://github.com/Netflix/hollow/files/13090292/PR_withDeltaTransitionsNoResharding.json)
[PR_withDeltaTransitionsWithResharding.json](https://github.com/Netflix/hollow/files/13090294/PR_withDeltaTransitionsWithResharding.json)

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectDeltaHistoricalStateCreator.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectDeltaHistoricalStateCreator.java
@@ -24,8 +24,10 @@ import com.netflix.hollow.core.memory.SegmentedByteArray;
 import com.netflix.hollow.core.memory.encoding.FixedLengthElementArray;
 import com.netflix.hollow.core.memory.pool.WastefulRecycler;
 import com.netflix.hollow.core.read.engine.PopulatedOrdinalListener;
+import com.netflix.hollow.core.schema.HollowObjectSchema;
 import com.netflix.hollow.core.util.IntMap;
 import com.netflix.hollow.core.util.RemovedOrdinalIterator;
+import java.util.Arrays;
 
 /**
  * This class contains the logic for extracting the removed records from an OBJECT type state
@@ -37,11 +39,8 @@ public class HollowObjectDeltaHistoricalStateCreator {
 
     private final HollowObjectTypeDataElements historicalDataElements;
 
-    private final int shardNumberMask;
-    private final int shardOrdinalShift;
-
     private HollowObjectTypeReadState typeState;
-    private HollowObjectTypeDataElements stateEngineDataElements[];
+    private HollowObjectTypeReadState.ShardsHolder shardsHolder;
     private RemovedOrdinalIterator iter;
     private IntMap ordinalMapping;
     private int nextOrdinal;
@@ -49,12 +48,10 @@ public class HollowObjectDeltaHistoricalStateCreator {
 
     public HollowObjectDeltaHistoricalStateCreator(HollowObjectTypeReadState typeState, boolean reverse) {
         this.typeState = typeState;
-        this.stateEngineDataElements = typeState.currentDataElements();
         this.historicalDataElements = new HollowObjectTypeDataElements(typeState.getSchema(), WastefulRecycler.DEFAULT_INSTANCE);
         this.iter = new RemovedOrdinalIterator(typeState.getListener(PopulatedOrdinalListener.class), reverse);
         this.currentWriteVarLengthDataPointers = new long[typeState.getSchema().numFields()];
-        this.shardNumberMask = stateEngineDataElements.length - 1;
-        this.shardOrdinalShift = 31 - Integer.numberOfLeadingZeros(stateEngineDataElements.length);
+        this.shardsHolder = typeState.shardsVolatile;
     }
 
     public void populateHistory() {
@@ -63,7 +60,7 @@ public class HollowObjectDeltaHistoricalStateCreator {
         historicalDataElements.fixedLengthData = new FixedLengthElementArray(historicalDataElements.memoryRecycler, (long)historicalDataElements.bitsPerRecord * (historicalDataElements.maxOrdinal + 1));
 
         for(int i=0;i<historicalDataElements.schema.numFields();i++) {
-            if(stateEngineDataElements[0].varLengthData[i] != null) {
+            if(isVarLengthField(typeState.getSchema().getFieldType(i))) {
                 historicalDataElements.varLengthData[i] = new SegmentedByteArray(historicalDataElements.memoryRecycler);
             }
         }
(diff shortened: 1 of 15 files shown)


hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectDeltaHistoricalStateCreator.java
```diff
@@ -74,9 +71,9 @@ public void populateHistory() {
         while(ordinal != ORDINAL_NONE) {
             ordinalMapping.put(ordinal, nextOrdinal);
 
-            int shard = ordinal & shardNumberMask;
-            int shardOrdinal = ordinal >> shardOrdinalShift;
-            copyRecord(historicalDataElements, nextOrdinal, stateEngineDataElements[shard], shardOrdinal, currentWriteVarLengthDataPointers);
+            int whichShard = ordinal & shardsHolder.shardNumberMask;
+            int shardOrdinal = ordinal >> shardsHolder.shards[whichShard].shardOrdinalShift;
```
dkoszewnik, line 75: I see this `holder.shards[x].shardOrdinalShift` used in a couple of places, I don't think the shard ordinal shift can actually be different between shards -- am I correct?

If so, why isn't the `shardOrdinalShift` a property of the holder, rather than the shard itself?

hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectDeltaHistoricalStateCreator.java
```diff
@@ -126,9 +123,12 @@ private void populateStats() {
...
 
-        for(int i=0;i<stateEngineDataElements[0].bitsPerField.length;i++) {
-            if(stateEngineDataElements[0].varLengthData[i] == null) {
-                historicalDataElements.bitsPerField[i] = stateEngineDataElements[0].bitsPerField[i];
+        for(int i=0;i<typeState.getSchema().numFields();i++) {
+            if(!isVarLengthField(typeState.getSchema().getFieldType(i))) {
+                final int fieldIdx = i;
+                historicalDataElements.bitsPerField[i] = Arrays.stream(shardsHolder.shards)
+                        .map(shard -> shard.dataElements.bitsPerField[fieldIdx])
+                        .max(Integer::compare).get();
```
dkoszewnik, line 131: Nice catch

hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectTypeReadState.java
```diff
@@ -265,39 +696,43 @@ public void setFieldSpecificSamplingDirector(HollowFilterConfig fieldSpec, Hollo
...
-        HollowObjectTypeDataElements currentDataElements[] = new HollowObjectTypeDataElements[shards.length];
-        
-        for(int i=0;i<shards.length;i++)
-            currentDataElements[i] = shards[i].currentDataElements();
-        
-        return currentDataElements;
+        final HollowObjectTypeReadStateShard[] shards = this.shardsVolatile.shards;
+        return Arrays.stream(shards)
+                .map(shard -> shard.dataElements)
+                .toArray(HollowObjectTypeDataElements[]::new);
```
dkoszewnik: The stream approach feels slightly harder to read, that might be my own bias.

hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectTypeReadStateShard.java
```diff
@@ -34,286 +31,127 @@
...
-            int numBits = currentData.bitsPerField[fieldIndex];
-            return (fixedLengthValue & (1L << (numBits - 1))) != 0;
-        case FLOAT:
-            return (int)fixedLengthValue == HollowObjectWriteRecord.NULL_FLOAT_BITS;
-        case DOUBLE:
-            return fixedLengthValue == HollowObjectWriteRecord.NULL_DOUBLE_BITS;
-        default:
-            return fixedLengthValue == currentData.nullValueForField[fieldIndex];
-        }
+    public long isNull(int ordinal, int fieldIndex) {
```
dkoszewnik: Feels like this method has the wrong name now -- it is now just reading the actual field data whether or not it is beyond the max bits for a single (possibly unaligned) read.

dkoszewnik (approved):

hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectDeltaHistoricalStateCreator.java
Sunjeet: `shardOrdinalShift ` can be different for different shards when we're in the middle of resharding.

Consider the case where we're resharding from 2 shards ```[S0, S1]```  to 4 shards .

In the first step of resharding, we will expand the shard sharray but retain the original shards as is to get-
```[S0, S1, S0, S1]```

In the next step of resharding, we process one shard at a time, so processing S0 yields-
```[S00, S1, S01, S1]```
Here the `dataElements` and `shardOrdinalShift` for the processed shard are different than the original shard but they co exist in the shard holder with shards that havent been processed yet (S1). 

Ultimately when we've finished resharding we arrive at 
```[S00, S10, S01, S11]``` 
and the `shardOrdinalShift`s are once again equal for all shards.

hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectTypeReadState.java
Sunjeet: Sure, reworked it

hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectTypeReadStateShard.java
Sunjeet: Good point, renamed to `readValue`

PR #648, opened 2023-11-08
add to version tag to header on producer side

workeatsleep:
add a tag in header to imply the version

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
@@ -388,6 +388,7 @@ abstract class AbstractHollowProducer {
                 } else {
                     writeEngine.getHeaderTags().remove(HollowStateEngine.HEADER_TAG_SCHEMA_CHANGE);
                 }
+                writeEngine.addHeaderTag(HollowStateEngine.HEADER_TAG_PRODUCER_TO_VERSION, String.valueOf(toVersion));
 
                 // 3a. Publish, run checks & validation, then announce new state consumers
                 publish(listeners, toVersion, artifacts);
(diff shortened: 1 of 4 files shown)


hollow/src/main/java/com/netflix/hollow/core/HollowStateEngine.java
```diff
@@ -63,6 +63,11 @@ public interface HollowStateEngine extends HollowDataset {
      * */
     String HEADER_TAG_METRIC_ANNOUNCEMENT = "hollow.metric.announcement";
 
+    /**
+     * A header tag indicating which version is this snapshot produce to.
+     */
+    String HEADER_TAG_PRODUCER_TO_VERSION = "hollow.producer.to.version";
```
Sunjeet: nit: name `hollow.blob.to.version` so it is more clear that this is the one in the blob (differentiating the one in metadata)

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -388,6 +388,7 @@ long runCycle(
                 } else {
                     writeEngine.getHeaderTags().remove(HollowStateEngine.HEADER_TAG_SCHEMA_CHANGE);
                 }
+                writeEngine.addHeaderTag(HollowStateEngine.HEADER_TAG_PRODUCER_TO_VERSION, String.valueOf(toVersion));
```
Sunjeet, line 391: The reverse delta header is a special case. When running cycle v2 we want the reverse delta header to have
```hollow.producer.to.version=v1```
and I think that should be getting handled already [here](https://github.com/Netflix/hollow/blob/ec5dcb4fe45385ddac8c906a4ee0a2debe9ed6f9/hollow/src/main/java/com/netflix/hollow/core/write/HollowBlobWriter.java#L296-L297) but its worth testing this out and if possible write a unit test, maybe [here](https://github.com/Netflix/hollow/blob/ec5dcb4fe45385ddac8c906a4ee0a2debe9ed6f9/hollow/src/test/java/com/netflix/hollow/core/write/HollowWriteStateEngineTest.java)

Sunjeet (approved):

PR #649, opened 2023-11-09
fix: the diff UI shows Dates that include millisecond precision

eduardoramirez:


hollow-diff-ui/src/main/java/com/netflix/hollow/diffview/effigy/HollowEffigyFactory.java
```diff
@@ -111,7 +114,7 @@ List<Field> createObjectFields(HollowEffigy effigy) {
             case LONG:
                 long longVal = typeDataAccess.readLong(effigy.ordinal, i);
                 if(longVal != Long.MIN_VALUE && "Date".equals(typeDataAccess.getSchema().getName()))
-                    fieldValue = new Date(longVal).toString();
+                    fieldValue = simpleDateFormat.format(new Date(longVal));
```
dvehar, line 117: Current change lgtm. I think this might be more helpful since it better reflects the stored value while still making the value easy to interpret. 

```suggestion
                    fieldValue = String.format("%d (%s)", longVal, simpleDateFormat.format(new Date(longVal)));
```

Same idea but will keep the same format we've been using
```suggestion
                    fieldValue = String.format("%d (%s)", longVal, new Date(longVal).toString());
```

Sunjeet (approved):

hollow-diff-ui/src/main/java/com/netflix/hollow/diffview/effigy/HollowEffigyFactory.java
eduardoramirez: yea, that's a good idea. for now, i want to make the minimal amount of changes to solve the diffing issue to avoid surprises

PR #650, opened 2023-11-16
change consumer to read metadata in blob header for version number

workeatsleep:
When the consumer apply a delta transition, it will check the blob version in blob metadata. If this is a transition happen during the app start up, keep the existing behavior, otherwise, throw an exception when there is a mismatch of blob metadata to version and delta to version.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/client/HollowDataHolder.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/client/HollowDataHolder.java
@@ -19,7 +19,6 @@ package com.netflix.hollow.api.client;
 import com.netflix.hollow.api.consumer.HollowConsumer;
 import com.netflix.hollow.api.consumer.HollowConsumer.TransitionAwareRefreshListener;
 import com.netflix.hollow.api.custom.HollowAPI;
-import com.netflix.hollow.core.HollowConstants;
 import com.netflix.hollow.core.memory.MemoryMode;
 import com.netflix.hollow.core.read.HollowBlobInput;
 import com.netflix.hollow.core.read.OptionalBlobPartInput;
@@ -34,6 +33,8 @@ import com.netflix.hollow.tools.history.HollowHistoricalStateDataAccess;
 import java.io.IOException;
 import java.lang.ref.WeakReference;
 import java.util.logging.Logger;
+
+import static com.netflix.hollow.core.HollowConstants.VERSION_NONE;
 
 /**
  * A class comprising much of the internal state of a {@link HollowConsumer}.  Not intended for external consumption.
@@ -56,7 +57,7 @@ class HollowDataHolder {
 
     private WeakReference<HollowHistoricalStateDataAccess> priorHistoricalDataAccess;
 
-    private long currentVersion = HollowConstants.VERSION_NONE;
+    private long currentVersion = VERSION_NONE;
 
     HollowDataHolder(HollowReadStateEngine stateEngine,
                             HollowAPIFactory apiFactory,
@@ -106,7 +107,7 @@ class HollowDataHolder {
     }
 
     void update(HollowUpdatePlan updatePlan, HollowConsumer.RefreshListener[] refreshListeners,
-            Runnable apiInitCallback) throws Throwable {
+            Runnable apiInitCallback, boolean isInitialUpdate) throws Throwable {
         // Only fail if double snapshot is configured.
         // This is a short term solution until it is decided to either remove this feature
         // or refine it.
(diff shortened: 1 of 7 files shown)


hollow/src/main/java/com/netflix/hollow/api/client/HollowDataHolder.java
```diff
@@ -177,7 +181,21 @@ private void applyStateEngineTransition(HollowBlobInput in, OptionalBlobPartInpu
             reader.applyDelta(in, optionalPartIn);
         }
 
-        setVersion(transition.getToVersion());
+        long expectedToVersion = transition.getToVersion();
+        Long actualToVersion = null;
+        String actualToVersionStr = stateEngine.getHeaderTag(HEADER_TAG_PRODUCER_TO_VERSION);
+        if (actualToVersionStr != null) {
```
Sunjeet: nit: can collapse the two if blocks that are checking the same condition

hollow/src/main/java/com/netflix/hollow/api/client/HollowDataHolder.java
```diff
@@ -177,7 +181,21 @@ private void applyStateEngineTransition(HollowBlobInput in, OptionalBlobPartInpu
...
+            actualToVersion = Long.valueOf(actualToVersionStr);
+        }
+
+        if (actualToVersion != null) {
+            if (actualToVersion.longValue() != expectedToVersion) {
+                LOG.warning("toVersion in blob didn't match toVersion seen in metadata; actualToVersion=" + actualToVersion + ", expectedToVersion=" + expectedToVersion);
```
Sunjeet: nit: the terms blob and metadata dont apply in the context of this class as we want them to, we can rephrase the msg to something like
```
toVersion encoded in incoming blob header did not match toVersion requested in transition; the one from blob header will be assumed
```

Sunjeet (approved):

Sunjeet (changes requested):
We discussed offline, there may be some side effects from updating the toVersion in one place (with the consumer) but continuing to use the bad toVersion later (for e.g. signaling to listeners).

hollow/src/main/java/com/netflix/hollow/api/error/VersionMismatchException.java
```diff
@@ -0,0 +1,21 @@
+package com.netflix.hollow.api.error;
+
+public class VersionMismatchException extends HollowException {
+    private final long expectedVersion;
+
+    private final long actualVersion;
+
+    public VersionMismatchException(long expectedVersion, long actualVersion) {
+        super("toVersion in blob didn't match toVersion seen in metadata; actualToVersion=" + actualVersion + ", expectedToVersion=" + expectedVersion);
```
Sunjeet: nit: metadata doesn't really mean apply in this context, we can drop it from the error msg

hollow/src/test/java/com/netflix/hollow/api/client/HollowClientUpdaterTest.java
```diff
@@ -261,6 +267,59 @@ private static void addActor(HollowWriteStateEngine stateEngine, int id) {
         stateEngine.add("Actor", rec);
     }
 
+    @Test
+    public void testReadBlobVersionInitialInconsistent() throws IOException {
+        testReadBlobVersion(2, 3, 1,  true);
```
Sunjeet, line 272: nit: (as discussed) can we add some asserts corresponding to versions that the consumer transitioned to

hollow/src/main/java/com/netflix/hollow/core/read/engine/HollowBlobReader.java
```diff
@@ -258,16 +264,34 @@ public void applyDelta(HollowBlobInput in, OptionalBlobPartInput optionalParts)
         notifyEndUpdate();
     }
 
-    private HollowBlobHeader readHeader(HollowBlobInput in, boolean isDelta) throws IOException {
+
+    private HollowBlobHeader readHeaderAndCheckVersion(HollowBlobInput in, boolean isDelta, long expectedVersion, boolean isInitialUpdate) throws IOException {
         HollowBlobHeader header = headerReader.readHeader(in);
+        if (expectedVersion != 0 && !isInitialUpdate) {
```
Sunjeet: (as discussed), maybe better approach to parse headers in HollowDataHolder before applying delta, so that we dont have to pass isInitialUpdate and expectedVersion context all the way down here

hollow/src/main/java/com/netflix/hollow/api/client/HollowDataHolder.java
workeatsleep: not apply anymore

hollow/src/main/java/com/netflix/hollow/api/client/HollowDataHolder.java
workeatsleep: not apply anymore

hollow/src/main/java/com/netflix/hollow/api/error/VersionMismatchException.java
workeatsleep: will change to "did not match toVersion requested in transition"

PR #654, opened 2023-12-21
fix: PerfAPI codegen supports reading a feed that has a subset of the schema

eduardoramirez:
Prevents the generated Perf API from throwing exceptions when used against an older version of the feed. 

repro 
- add a new field to schema
- re-generate the client
- populate the feed in test, test works
- release code to prod but don't update feed in prod yet. PerfAPI throws ArrayOutOfBounds on getter for field that isn't populated in feed but exists in schema

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/codegen/perfapi/HollowObjectTypePerfAPIClassGenerator.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/codegen/perfapi/HollowObjectTypePerfAPIClassGenerator.java
@@ -84,10 +84,14 @@ class HollowObjectTypePerfAPIClassGenerator {
         switch(fieldType) {
             case INT:
                 builder.append("    public int get" + HollowCodeGenerationUtils.upperFirstChar(fieldName) + "(long ref) {\n");
+                builder.append("        if(fieldIdx[" + fieldIdx + "] == -1)\n");
+                builder.append("            return Integer.MIN_VALUE;\n");
                 builder.append("        return typeAccess.readInt(ordinal(ref), fieldIdx[" + fieldIdx + "]);\n");
                 builder.append("    }\n\n");
 
                 builder.append("    public Integer get" + HollowCodeGenerationUtils.upperFirstChar(fieldName) + "Boxed(long ref) {\n");
+                builder.append("        if(fieldIdx[" + fieldIdx + "] == -1)\n");
+                builder.append("            return null;\n");
                 builder.append("        int val = typeAccess.readInt(ordinal(ref), fieldIdx[" + fieldIdx + "]);\n");
                 builder.append("        if(val == Integer.MIN_VALUE)\n");
                 builder.append("            return null;\n");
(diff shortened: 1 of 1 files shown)


hollow/src/main/java/com/netflix/hollow/api/codegen/perfapi/HollowObjectTypePerfAPIClassGenerator.java
```diff
@@ -84,10 +84,14 @@ public void appendFieldMethod(StringBuilder builder, FieldType fieldType, String
...
             case INT:
                 builder.append("    public int get" + HollowCodeGenerationUtils.upperFirstChar(fieldName) + "(long ref) {\n");
+                builder.append("        if(fieldIdx[" + fieldIdx + "] == -1)\n");
+                builder.append("            return Integer.MIN_VALUE;\n");
                 builder.append("        return typeAccess.readInt(ordinal(ref), fieldIdx[" + fieldIdx + "]);\n");
                 builder.append("    }\n\n");
 
                 builder.append("    public Integer get" + HollowCodeGenerationUtils.upperFirstChar(fieldName) + "Boxed(long ref) {\n");
+                builder.append("        if(fieldIdx[" + fieldIdx + "] == -1)\n");
+                builder.append("            return null;\n");
```
Sunjeet, line 94: Shouldn't these (Integer, Long, Float, and Double) return `Integer.MIN_VALUE`/`Float.NaN` etc.? Just going by the classic generated API-
In `LongTypeAPI.java`,
```
public Long getValueBoxed(int ordinal) {
        long l;
        if(fieldIndex[0] == -1) {
            l = missingDataHandler().handleLong("Long", ordinal, "value");
        ...
```

hollow/src/main/java/com/netflix/hollow/api/codegen/perfapi/HollowObjectTypePerfAPIClassGenerator.java
```diff
@@ -132,29 +148,41 @@ public void appendFieldMethod(StringBuilder builder, FieldType fieldType, String
...
             case BOOLEAN:
                 builder.append("    public boolean get" + HollowCodeGenerationUtils.upperFirstChar(fieldName) + "(long ref) {\n");
+                builder.append("        if(fieldIdx[" + fieldIdx + "] == -1)\n");
+                builder.append("            return false;\n");
                 builder.append("        return Boolean.TRUE.equals(typeAccess.readBoolean(ordinal(ref), fieldIdx[" + fieldIdx + "]));\n");
                 builder.append("    }\n\n");
 
                 builder.append("    public Boolean get" + HollowCodeGenerationUtils.upperFirstChar(fieldName) + "Boxed(long ref) {\n");
+                builder.append("        if(fieldIdx[" + fieldIdx + "] == -1)\n");
+                builder.append("            return null;\n");
```
Sunjeet, line 158: For Boolean this seems to match what the classic API would return

hollow/src/main/java/com/netflix/hollow/api/codegen/perfapi/HollowObjectTypePerfAPIClassGenerator.java
eduardoramirez: hmm interesting... in the actual generated code for something like `MidInsightsDelegateLookupImpl` the getter for primitive and boxed long looks like this: 

```java
    public long getMovieId(int ordinal) {
        ordinal = typeAPI.getMovieIdOrdinal(ordinal);
        return ordinal == -1 ? Long.MIN_VALUE : typeAPI.getAPI().getLongTypeAPI().getValue(ordinal);
    }

    public Long getMovieIdBoxed(int ordinal) {
        ordinal = typeAPI.getMovieIdOrdinal(ordinal);
        return ordinal == -1 ? null : typeAPI.getAPI().getLongTypeAPI().getValueBoxed(ordinal);
    }
```

Sunjeet: Hmm yeah looks like the defaults aren't consistently applied in the generated api

Sunjeet: Lets ship this improvement for now, I've tracked the fix for consistently applying defaults in [this issue](https://github.com/Netflix/hollow/issues/656)

Sunjeet (approved):

PR #660, opened 2024-01-22
Various Hashing Improvements

stevenewald:


hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
```diff
@@ -59,6 +59,20 @@ public class ByteArrayOrdinalMap {
 
     private long[] pointersByOrdinal;
 
+    private static int nextLargestPrime(int num) {
+        int[] precomputedPrimes = new int[]
```
Sunjeet, line 63: If we pick prime numbers close to these numbers that are also (2^n-1), then we can replace the modulo with simpler bitwise operations

stevenewald: Good idea! Note the sizable gaps between sequential Mersenne primes (like 127 and 521), which might impact resizing memory efficiency. Considering we're not using modulo during LP, I think the current approach is acceptable.

dkoszewnik: Restricting to mersenne primes will result in significant memory overhead -- each mersenne number is ~double the previous, and many of these are not prime, so an increase will often result in a significant step larger than doubling the space.

Sunjeet: Yeah they seem too few and far between 👍

PR #666, opened 2024-02-22
fix: HollowObjectMapper throws if it encounters a duplicate field name while walking the POJO class hierarchy

eduardoramirez:


stevenewald (approved):

hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/HollowObjectTypeMapper.java
```diff
@@ -120,7 +120,11 @@ public HollowObjectTypeMapper(HollowObjectMapper parentMapper, Class<?> clazz, S
 
         this.schema = new HollowObjectSchema(typeName, mappedFields.size(), getKeyFieldPaths(clazz));
 
+        Set<String> fieldNamesSeen = new HashSet<>();
         for(MappedField field : mappedFields) {
+            if(!fieldNamesSeen.add(field.getFieldName()))
+                throw new IllegalArgumentException("Duplicate field name '" + field.getFieldName() + "' found in class hierarchy for class " + clazz.getName());
```
Sunjeet, line 126: I wonder if instead of failing hard we should ignore the duplicate field (i.e. replace `throw` with `continue`).  Not a big deal either way since a field with the same name in superclass and subclass is generally frowned upon (hard to find bugs due to field hiding).

eduardoramirez: that was my first instinct as well, i'm open to either approach. i leaned towards the hard error to be explicit and avoid surprises later. i suspect dupes are typically a mistake when it happens (this was true for the case that exposed this bug) and the exception would sunshine this to the user much sooner.

Issue #492, opened 2020-11-12
Json-ifying string with newlines in it creates invalid JSON


akhaku:

yalinhuang:
We ran into another similar issue due to special characters not escaped in `HollowRecordJsonStringifier.stringify()` -- it's a tab character ('\t'). Despite that such characters might not make any valuable meaning to the data and should be sanitized before passing down to Cinder, it'd be nice that the `HollowRecordJsonStringifier.stringify()` can handle general data values.

nayanika-u:
Hi @yalinhuang, looking into this one.

PR #672, opened 2024-03-21
Fix: Escape control chars when stringifing JSON (#492)

clay-mayers:
As it was originally coded, this change assumes escaping is uncommon.  The string is scanned first for characters that require escaping (i.e \ " and control characters).  A single scan is performed checking for the presence of any of these three cases.  The type of characters found are accumulated and used to only replace those that were found in the string.

The presence of a given category/type of escape character is done using bit-wise operations to avoid comparisons for performance (testing a character this way is 12% faster with jdk11 on an m1).


clay-mayers:
OBE

PR #685, opened 2024-07-17
add cycle start time and end time to comsumer metrics

workeatsleep:
Add a duration gauge value for consumer metrics

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/metrics/HollowConsumerMetrics.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/metrics/HollowConsumerMetrics.java
@@ -18,9 +18,31 @@ package com.netflix.hollow.api.metrics;
 
 import com.netflix.hollow.core.read.engine.HollowReadStateEngine;
 
+import java.util.concurrent.atomic.AtomicLong;
+
 public class HollowConsumerMetrics extends HollowMetrics {
     private int refreshFailed;      // TODO: Move these metrics over to com.netflix.hollow.api.consumer.metrics.AbstractRefreshMetricsListener
     private int refreshSucceeded;
+    private final AtomicLong lastRefreshStartNs = new AtomicLong(); //last time a refresh started
+    private final AtomicLong lastRefreshEndNs = new AtomicLong();
+
+    public void setLastRefreshStartNs(long startNs) {
+        lastRefreshStartNs.set(startNs);
+    }
+
+    public AtomicLong getLastRefreshStartNs() {
+        return lastRefreshStartNs;
+    }
+
+    public void setLastRefreshEndNs(long endNs) {
+        lastRefreshEndNs.set(endNs);
+    }
+
+    public AtomicLong getLastRefreshEndNs() {
+       return lastRefreshEndNs;
+    }
+
+
 
     /**
      * Updates the consumer metrics:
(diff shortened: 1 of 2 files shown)


hollow/src/main/java/com/netflix/hollow/api/metrics/HollowConsumerMetrics.java
```diff
@@ -18,9 +18,31 @@
 
 import com.netflix.hollow.core.read.engine.HollowReadStateEngine;
 
+import java.util.concurrent.atomic.AtomicLong;
+
 public class HollowConsumerMetrics extends HollowMetrics {
     private int refreshFailed;      // TODO: Move these metrics over to com.netflix.hollow.api.consumer.metrics.AbstractRefreshMetricsListener
     private int refreshSucceeded;
+    private AtomicLong lastRefreshStartTimeNs; //last time a refresh started
```
azeng-netflix: wonder if they should all be moved to `AbstractRefreshMetricsListener` in the future as suggested by the comment above...
`AtomicLong` seems to be unnecessary as it's supposed to be used in a thread-safe manner? Concurrent writes to  `lastRefreshStartTimeNs` and `lastRefreshEndTimeNs` seems meaningless.

hollow/src/main/java/com/netflix/hollow/api/metrics/HollowConsumerMetrics.java
```diff
@@ -18,9 +18,31 @@
...
+    private AtomicLong lastRefreshEndTimeNs;
+
+    public void setLastRefreshStartTimeNs() {
+        lastRefreshStartTimeNs = new AtomicLong(System.nanoTime());
```
Sunjeet: I think we're going for thread safety here.. Assigning a non volatile variable to a new AtomicLong wont achieve that

workeatsleep: fixed it and updated the PR

azeng-netflix (approved):

hollow/src/main/java/com/netflix/hollow/api/metrics/HollowConsumerMetrics.java
workeatsleep: We cannot move this to the listener because listeners are not invoked until [later in the update](https://github.com/Netflix/hollow/blob/master/hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java#L142-L144), and also because of the sequence of the listeners, it cannot record both start and end time at the very beginning and end. The value will be read from another thread to report metrics, hence we need it to be threadsafe.

hollow/src/main/java/com/netflix/hollow/api/metrics/HollowConsumerMetrics.java
```diff
@@ -18,9 +18,31 @@
...
 
+import java.util.concurrent.atomic.AtomicLong;
+
 public class HollowConsumerMetrics extends HollowMetrics {
     private int refreshFailed;      // TODO: Move these metrics over to com.netflix.hollow.api.consumer.metrics.AbstractRefreshMetricsListener
     private int refreshSucceeded;
+    private final AtomicLong lastRefreshStartTimeNs = new AtomicLong(); //last time a refresh started
+    private final AtomicLong lastRefreshEndTimeNs = new AtomicLong();
+
+    public void setLastRefreshStartTimeNs() {
```
Sunjeet: nit: since System.nanoTime is not related to wall clock time, we can drop "Time" from the method name, so like `setLastRefreshStartNanos`.

Also nit: convention with setters and getters, setters will take the value as an argument

Sunjeet (approved):

PR #687, opened 2024-07-24
chore: if the bool field is null set the inlined boolean as null

eduardoramirez:
for hollow inline, we want to keep the concept of null like we do for the other ones. it's not 100% backwards compatible, but since this is only used within RH we can see how safe it is based on usage. ideally, we patch this before broader usage appears


eduardoramirez:
i addressed the existing usage that this change would have broken. i will land this next week.

PR #688, opened 2024-07-25
fix: HollowMapWriteRecord serialization should sort on key and value

Sunjeet:
For cases where duplicate keys exist in the map (e.g. equals/hashcode was not overridden in data model).
Sorting on both key and value ordinals ensures consistent serializarion/checksum (and prevents integrity check failures).

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/write/HollowMapWriteRecord.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/write/HollowMapWriteRecord.java
@@ -26,12 +26,18 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
+import java.util.logging.Logger;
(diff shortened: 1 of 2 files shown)


hollow/src/main/java/com/netflix/hollow/core/write/HollowMapWriteRecord.java
```diff
@@ -26,12 +26,18 @@
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
+import java.util.logging.Logger;
 
 public class HollowMapWriteRecord implements HollowHashableWriteRecord {
+    private static final Logger LOG = Logger.getLogger(HollowMapWriteRecord.class.getName());
```
eduardoramirez: ✂️ ?

eduardoramirez (approved):

PR #697, opened 2024-08-27
add validator to validate percentage of record change

workeatsleep:


hollow/src/main/java/com/netflix/hollow/api/producer/validation/RecordCountChangeValidator.java
```diff
@@ -0,0 +1,125 @@
...
+
+import com.netflix.hollow.api.consumer.data.AbstractHollowDataAccessor;
+import com.netflix.hollow.api.producer.HollowProducer;
+import com.netflix.hollow.core.read.engine.HollowReadStateEngine;
+import com.netflix.hollow.core.read.engine.HollowTypeReadState;
+import java.util.logging.Logger;
+
+import static java.util.Objects.requireNonNull;
+
+public class RecordCountChangeValidator implements ValidatorListener {
```
Sunjeet: Lets add to the javadoc that this only supports types that have a primary key definition

Sunjeet: naming: since we're observing percent change lets call it `RecordCountPercentChangeValidator`

hollow/src/main/java/com/netflix/hollow/api/producer/validation/RecordCountChangeValidator.java
```diff
@@ -0,0 +1,125 @@
...
+            this.removedPercentageThreshold = removedPercentageThreshold;
+            this.addedPercentageThreshold = addedPercentageThreshold;
+            this.updatedPercentageThreshold = updatedPercentageThreshold;
+        }
+
+        public static ThresholdBuilder builder() {
+            return new ThresholdBuilder();
+        }
+
+        public static class ThresholdBuilder {
```
Sunjeet: if a use case using the builder pattern only supplies the deletion threshold, i think we treat it as- they dont care about added/updated percent. Those should be pass through.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/RecordCountChangeValidator.java
```diff
@@ -0,0 +1,125 @@
...
+        int removeRecordNumber = accessor.getRemovedRecords().size();
+        int updatedRecordNumber = accessor.getUpdatedRecords().size();
+        int previousRecordNumber = typeState.getPreviousOrdinals().cardinality();
+
+        float addedPercent = (float) addRecordNumber / previousRecordNumber;
+        float removedPercent = (float) removeRecordNumber / previousRecordNumber;
+        float updatedPercent = (float) updatedRecordNumber / previousRecordNumber;
+
+
+        float addedPercentageThreshold = threshold.addedPercentageThreshold;
```
Sunjeet: lets make these dynamically configurable for when producers want to let a publish through

hollow/src/main/java/com/netflix/hollow/api/producer/validation/RecordCountChangeValidator.java
```diff
@@ -0,0 +1,125 @@
...
+            private float removedPercentageThreshold = -1f;
+            private float addedPercentageThreshold = -1f;
+            private float updatedPercentageThreshold = -1f;
+
+            public ThresholdBuilder withRemovedPercentageThreshold(float removedPercentageThreshold) {
+                this.removedPercentageThreshold = removedPercentageThreshold;
+                return this;
+            }
+
+            public ThresholdBuilder withAddedPercentageThreshold(float addedPercentageThreshold) {
```
Sunjeet: since the threshold type is float, some javadoc explaining how the input parameter is interpreted would be helpful here, for e.g. is passing 0.01 treated as 1% or 0.01%? for deleted and updated, we can validate that the input is in the range (0,100]

hollow/src/main/java/com/netflix/hollow/api/producer/validation/RecordCountChangeValidator.java
Sunjeet: This seems already handed since we're initializing the thresholds to -1 👍

Sunjeet: maybe this usage can also be documented in javadoc

Sunjeet:
Also lets add this to hollow user docs https://hollow.how/validation/?h=validato#pre-defined-validators

hollow/src/main/java/com/netflix/hollow/api/producer/validation/RecordCountChangeValidator.java
azeng-netflix: Maybe also throw exception to explicitly notify an incorrect use of the validator?

Why does this only support types with primary key definition?

Sunjeet: The AbstractHollowDataAccessor will throw an exception if the type doesn't have a primary key associated, that works. 

>Why does this only support types with primary key definition?
Without a primary key we don't know if an add+remove was a modification or not (the way to detect modification is to match added and removed record for primary key)

Sunjeet (approved):

PR #713, opened 2024-12-19
fix: if the max ordinal for a map value is zero then reserve 1 bit for it instead of nothing

eduardoramirez:
Currently, when a Map is written where all the keys reference the same value ordinal and that ordinal is zero then we don't reserve any space for the value ordinal. when a map entry is hashed into the last bucket, it overflows when trying to set the value in the data array.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/write/HollowMapTypeWriteState.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/write/HollowMapTypeWriteState.java
@@ -134,7 +134,7 @@ public class HollowMapTypeWriteState extends HollowTypeWriteState {
         }
 
         bitsPerKeyElement = 64 - Long.numberOfLeadingZeros(maxKeyOrdinal + 1);
-        bitsPerValueElement = 64 - Long.numberOfLeadingZeros(maxValueOrdinal);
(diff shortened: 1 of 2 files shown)


hollow/src/main/java/com/netflix/hollow/core/write/HollowMapTypeWriteState.java
```diff
@@ -134,7 +134,7 @@ private void gatherStatistics() {
         }
 
         bitsPerKeyElement = 64 - Long.numberOfLeadingZeros(maxKeyOrdinal + 1);
-        bitsPerValueElement = 64 - Long.numberOfLeadingZeros(maxValueOrdinal);
+        bitsPerValueElement = maxValueOrdinal == 0 ? 1 : 64 - Long.numberOfLeadingZeros(maxValueOrdinal);
```
Sunjeet, line 137: Repro
```
    @Test
    public void triggerBugWithCurrentHashImpl() {
    ...    
        producer.runCycle((state) -> {
            for (int i=0; i<16384; i++) {  // exactly 16384 entries in the map, or a multiple thereof
                final long id = new Long(i);
                state.add(
                        new MapOfMovies(
                            new HashMap<Long, MyMovie>() {{
                                put(id, new MyMovie(0, "The Matrix")); // all entries pointing to same object at ordinal 0
                }}));
            }
        });
    }
```

Sunjeet (approved):

Issue #566, opened 2022-05-30
Condition 'doubleSnapshotConfig.allowDoubleSnapshot()' should be negated


wangsfx:
https://github.com/Netflix/hollow/blob/d4d3dbed944a8885ba128843a61e569ddbcce08e/hollow/src/main/java/com/netflix/hollow/api/client/HollowDataHolder.java#L121

The code is inconsistent with the comments 'If the consumer is configured to only follow deltas (**no double snapshot**) then any failure to transition will cause the consumer to become "stuck" on stale data', so the condition 'doubleSnapshotConfig.allowDoubleSnapshot()' should be **negated**

PR #720, opened 2025-02-08
Improvements in failure handling of transition in update processes  

Brijeshthummar02:
issue #566 

- Added a check to prevent updates when `doubleSnapshot` is disabled and a known failing transition exists.  
- Introduced the `STRICT_TRANSITION_FAILURE_HANDLING` system property:  
  - If set to true, an exception is thrown to prevent stale data.  
  - If set to false, a warning is logged instead.  
- To provide better control over transition failures and to avoid the consumers being stuck on stale data.


Brijeshthummar02:
@Sunjeet good to go?

Brijeshthummar02:
@yujiaaa2019 is it good to go??

Brijeshthummar02:
@yujiaaa2019 need your review on PR

PR #726, opened 2025-03-20
adds expand all feature to the history diff ui

AustinWheel:


Sunjeet (approved):

hollow-diff-ui/src/main/resources/history-header.vm
```diff
@@ -45,10 +47,32 @@ Lookup Key:
...
+            if (parentRow && typeof parentRow.onclick === "function") {
+                if (parentRow.onclick.toString().includes("uncollapseRow") && isCollapsed) {
+                    parentRow.onclick(i);
+                } else if (parentRow.onclick.toString().includes("collapseRow") && !isCollapsed) {
+                    parentRow.onclick(i);
+                }
+            }
+        }
+        isCollapsed = !isCollapsed;
+        if (isCollapsed) {
```
workeatsleep, line 65: what would it show when it's partially expanded?

workeatsleep: should we show both expand all and collapse all?

AustinNflx: Maybe having two buttons with separate functions would be more clear to the user. But right now, if its partially expanded, then those expanded rows will remain expanded when the user presses "expand all".

PR #727, opened 2025-04-01
fix: consumer init to an announced version should only retrieve snapshots corresponding to announced versions

Sunjeet:
### The bug
```
1. A producer announces V0, 
2. Then starts failing validations, so it publishes a snapshot for V1 which is not announced due to validation failure,
3. Then later on announces V2 but with no corresponding snapshot because numStatesBetweenSnapshot is not met,
4. And later on announces Vn with a corresponding snapshot.

Then a consumer starting between 3 and 4 will attempt to initialize to the latest announcement V2 but will initialize to V1 -thus initializing to a version that wasn’t announced because it failed validation.
```
### Impact
* Risk of consumers landing on a version of data that was not announced (described above)
* Consumer can fail to recover from forked delta chains using double-snapshot until a snapshot on the new delta chain is announced.

### Fix
Introduced a `UpdatePlanVerifier` that can be configured when building a consumer instance to enable this verification. If enabled, a consumer update is triggered to an announced version (as indicated in the trigger refresh to version info), the update plan computation will verify that retrieved snapshot blobs were also announced- until it finds an announced snapshot or exhausts configured max lookback attempts.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
@@ -76,7 +76,7 @@ public class HollowClientUpdater {
                                HollowMetricsCollector<HollowConsumerMetrics> metricsCollector) {
         this(transitionCreator, refreshListeners, apiFactory, doubleSnapshotConfig, hashCodeFinder, memoryMode,
                 objectLongevityConfig, objectLongevityDetector, metrics, metricsCollector,
-                null, HollowConsumer.UpdatePlanBlobVerifier.DEFAULT_INSTANCE);
+                HollowConsumer.UpdatePlanBlobVerifier.DEFAULT_INSTANCE);
     }
 
     public HollowClientUpdater(HollowConsumer.BlobRetriever transitionCreator,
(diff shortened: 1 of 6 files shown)


hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
```diff
@@ -73,7 +74,24 @@ public HollowClientUpdater(HollowConsumer.BlobRetriever transitionCreator,
...
+                               List<HollowConsumer.RefreshListener> refreshListeners,
+                               HollowAPIFactory apiFactory,
+                               HollowConsumer.DoubleSnapshotConfig doubleSnapshotConfig,
+                               HollowObjectHashCodeFinder hashCodeFinder,
+                               MemoryMode memoryMode,
+                               HollowConsumer.ObjectLongevityConfig objectLongevityConfig,
+                               HollowConsumer.ObjectLongevityDetector objectLongevityDetector,
+                               HollowConsumerMetrics metrics,
+                               HollowMetricsCollector<HollowConsumerMetrics> metricsCollector,
+                               HollowConsumer.AnnouncementWatcher announcementWatcher,
```
workeatsleep: announcementWatcher is not used

workeatsleep (approved):

azeng-netflix:
Should all the public method that using `long` instead of `HollowConsumer.VersionInfo` be marked as `deprecated`?
i.e. this [updateTo](https://github.com/Netflix/hollow/blob/master/hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java#L116)

hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java
Sunjeet: thanks for that, it has been cleaned up

Sunjeet:
> Should all the public method that using `long` instead of `HollowConsumer.VersionInfo` be marked as `deprecated`? i.e. this [updateTo](https://github.com/Netflix/hollow/blob/master/hollow/src/main/java/com/netflix/hollow/api/client/HollowClientUpdater.java#L116)


Would be good to continue to support the simple usage since it is more concise, and the optional version info fields are irrelevant in many cases for e.g. consumer.triggerRefreshTo(Long.MAX_VALUE)

PR #736, opened 2025-05-28
fix: Use thread-safe datastructure in SimultaneousExecutor

ramarajesh2112:
Fixed potential thread-safety issue by using ConcurrentLinkedQueue instead of ArrayList.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/util/SimultaneousExecutor.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/util/SimultaneousExecutor.java
@@ -234,11 +234,6 @@ public class SimultaneousExecutor extends ThreadPoolExecutor {
      * Await successful completion of all submitted tasks. Throw exception of the first failed task
      * if 1 or more tasks failed.
      *
-     * If tasks are being submitted concurrently from other threads while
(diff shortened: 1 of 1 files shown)


Sunjeet (approved):
The change looks good.. just note that the iterator on the ConcurrentLinkedQueue will be weakly consistent
 for (final Future<?> f : futures) {
            f.get();
        }
so when adding and awaiting completion from different threads so be mindful of that behavior- may even be good to add that bit about thread safety to the javadoc on this class.
Thanks Ram!


p.s. also if i may suggest adding to the javadoc that this class isn't really meant for external consumption :)

ramarajesh2112:
> The change looks good.. just note that the iterator on the ConcurrentLinkedQueue will be weakly consistent for (final Future<?> f : futures) { f.get(); } so when adding and awaiting completion from different threads so be mindful of that behavior- may even be good to add that bit about thread safety to the javadoc on this class. Thanks Ram!
> 
> p.s. also if i may suggest adding to the javadoc that this class isn't really meant for external consumption :)

Sounds good. Added javadocs as suggested

Sunjeet (approved):

PR #739, opened 2025-06-16
sort stringified set elements with one field

kvoli:
Stringified set elements are now outputted in sorted order when each
element has exactly one field. Previously the output order was
determined by the element ordinals.

This change only affects the Hollow Explorer UI.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/tools/stringifier/HollowRecordJsonStringifier.java
+++ b/hollow/src/main/java/com/netflix/hollow/tools/stringifier/HollowRecordJsonStringifier.java
@@ -17,6 +17,7 @@
 package com.netflix.hollow.tools.stringifier;
 
 import static com.netflix.hollow.core.HollowConstants.ORDINAL_NONE;
+import static com.netflix.hollow.core.read.HollowReadFieldUtils.compareFieldValues;
 
 import com.netflix.hollow.api.objects.HollowRecord;
 import com.netflix.hollow.core.read.dataaccess.HollowDataAccess;
@@ -40,6 +41,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Set;
 
 /**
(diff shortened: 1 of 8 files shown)


hollow/src/main/java/com/netflix/hollow/tools/stringifier/HollowRecordJsonStringifier.java
```diff
@@ -51,21 +53,24 @@ public class HollowRecordJsonStringifier implements HollowStringifier<HollowReco
...
     private final boolean prettyPrint;
+    private final boolean sortSingleFieldSetElements;
 
     public HollowRecordJsonStringifier() {
-        this(true, true);
+        this(true, true, false);
     }
 
-    public HollowRecordJsonStringifier(boolean prettyPrint, boolean collapseAllSingleFieldObjects) {
+    public HollowRecordJsonStringifier(boolean prettyPrint, boolean collapseAllSingleFieldObjects, boolean sortSingleFieldSetElements) {
```
Sunjeet, line 66: We'll need to keep the old constructor around too for backwards compatibility

hollow/src/main/java/com/netflix/hollow/tools/stringifier/HollowRecordStringifier.java
```diff
@@ -48,15 +50,17 @@ public class HollowRecordStringifier implements HollowStringifier<HollowRecordSt
...
     private final boolean collapseSingleFieldObjects;
+    private final boolean sortSingleFieldSetElements;
 
     public HollowRecordStringifier() {
-        this(false, false, true);
+        this(false, false, true, false);
     }
 
-    public HollowRecordStringifier(boolean showOrdinals, boolean showTypes, boolean collapseSingleFieldObjects) {
+    public HollowRecordStringifier(boolean showOrdinals, boolean showTypes, boolean collapseSingleFieldObjects, boolean sortSingleFieldSetElements) {
```
Sunjeet, line 63: same backwards compatibility thing here

hollow/src/test/java/com/netflix/hollow/core/write/objectmapper/HollowObjectMapperTest.java
```diff
@@ -52,7 +52,7 @@ public void testBasic() throws IOException {
 
 
         Assert.assertEquals("{\"a1\": \"two\",\"a2\": 2,\"b\": {\"b1\": 20,\"b2\": 20000000,\"b3\": 2.2,\"b4\": \"two\",\"b5\": [2, 2, 2]},\"cList\": []}",
-                new HollowRecordJsonStringifier(false, true).stringify(readStateEngine, "TypeA", 0));
+                new HollowRecordJsonStringifier(false, true, false).stringify(readStateEngine, "TypeA", 0));
```
Sunjeet: the backwards compatibility concern: so we don't break any external callers of that constructor, like we had to fix fwd here

hollow/src/main/java/com/netflix/hollow/tools/stringifier/HollowRecordJsonStringifier.java
kvoli: Updated.

hollow/src/main/java/com/netflix/hollow/tools/stringifier/HollowRecordStringifier.java
kvoli: Updated.

hollow/src/test/java/com/netflix/hollow/core/write/objectmapper/HollowObjectMapperTest.java
kvoli: Makes sense, updated the two constructors.

kvoli:
TYFTR! Updated with backwards compatibility constructor changes @Sunjeet.

Sunjeet (approved):

PR #741, opened 2025-06-17
handle null pkeys in explorer ui

kvoli:
Null primary keys are not supported, however when a null primary key is present, browsing a selected type in the Hollow Explorer UI will fail with an OOB exception.

Handle null primary keys so that they are still displayed in the selected type page, with "NULL" substituted in-place of missing fields.


Sunjeet (approved):

kvoli:
Updated to display "null" instead of "NULL".

kvoli:
TYFTR!

PR #742, opened 2025-06-18
throw null ptr exception when constructing null pkey idx

kvoli:
When constructing a primary key index where a referenced key field
contains a null value, an array out of bounds exception would be thrown:

```
java.lang.ArrayIndexOutOfBoundsException: 0
```

Now, throw a null pointer exception instead, with information detailing
the type, field and ordinal:

```
java.lang.NullPointerException: Cannot hash null field id in type Movie at ordinal 0
```


Sunjeet (approved):

kvoli:
TYFTR!

PR #743, opened 2025-06-20
add top level schema page in explorer ui

kvoli:
Clicking on `Browse Top Level Schemas` will return a schema page with all of the top level schemas. A top level schema is not referenced by any other schema.

As part of this change, the `BrowseSchemaPage` was updated to support displaying multiple types.

### DIFF
--- a/hollow-explorer-ui/src/main/resources/explorer-header.vm
+++ b/hollow-explorer-ui/src/main/resources/explorer-header.vm
@@ -34,10 +34,15 @@
 			#end
 			#if($stateVersion)
 				<td class="nav">[ State Version: $stateVersion ]</td>
-			#end			
-			#if($type)
+			#end
+			## When $type is defined and is not a list, display links to browse
+			## data and schema for that type.
+			#if($type && !($type.size() > 0))
 				<td class="nav">[ Type: <span style="font-family: monospace;">$esc.html($type)</span> ]</td>
 				<td class="nav">[ <a href="$basePath/type?type=$esc.url($type)">Browse Data</a> | <a href="$basePath/schema?type=$esc.url($type)">Browse Schema</a> ]</td>
+			#end
+			#if(!$type && $topLevelTypes && $topLevelTypes.size() > 0)
+				<td class="nav">[ <a href="$basePath/schema?#foreach($t in $topLevelTypes)type=$esc.url($t)#if($foreach.hasNext)&amp;#end#end">Browse Top Level Schemas</a> ]</td>
 			#end
 			<td class="nav">[ <a href="$basePath/query">SEARCH</a> ]</td>
 		</tr>
--- a/hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/pages/BrowseSchemaPage.java
+++ b/hollow-explorer-ui/src/main/java/com/netflix/hollow/explorer/ui/pages/BrowseSchemaPage.java
@@ -21,6 +21,8 @@ import com.netflix.hollow.explorer.ui.model.SchemaDisplay;
 import com.netflix.hollow.explorer.ui.model.SchemaDisplayField;
 import com.netflix.hollow.ui.HollowUISession;
 import java.io.Writer;
+import java.util.ArrayList;
+import java.util.List;
 import javax.servlet.http.HttpServletRequest;
 import org.apache.velocity.VelocityContext;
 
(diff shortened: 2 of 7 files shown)


hollow/src/main/java/com/netflix/hollow/core/read/engine/HollowReadStateEngine.java
```diff
@@ -200,6 +201,43 @@ public Map<String, Long> calcApproxShardSizePerType() {
...
     }
 
+    /* *
+     * Finds the top-level types in the Hollow dataset. A top-level type is one that is not a
+     * reference from another type, i.e., it is not an element of a list, set, or map, nor is it
+     * a field in an object schema.
+     *
+     * @return a set of top-level type names
+     */
+    public Set<String> topLevelTypes() {
```
Sunjeet: consider moving this to a HollowSchemaUtil in `com.netflix.hollow.core.schema`

hollow/src/main/java/com/netflix/hollow/core/read/engine/HollowReadStateEngine.java
```diff
@@ -41,6 +41,7 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
```
Sunjeet: nit: unused import?

hollow-explorer-ui/src/main/resources/explorer-header.vm
```diff
@@ -34,11 +34,14 @@
 			#end
 			#if($stateVersion)
 				<td class="nav">[ State Version: $stateVersion ]</td>
-			#end			
-			#if($type)
+			#end
+			#if($type && !($type.size() > 0))
```
Sunjeet, line 40: trying to understand this if condition, a code comment might be useful here

hollow-explorer-ui/src/main/resources/explorer-header.vm
```diff
@@ -34,11 +34,14 @@
...
 				<td class="nav">[ Type: <span style="font-family: monospace;">$esc.html($type)</span> ]</td>
 				<td class="nav">[ <a href="$basePath/type?type=$esc.url($type)">Browse Data</a> | <a href="$basePath/schema?type=$esc.url($type)">Browse Schema</a> ]</td>
 			#end
+			#if(!$type)
```
Sunjeet: should we be checking that $topLevelTypes is non null?

hollow/src/main/java/com/netflix/hollow/core/read/engine/HollowReadStateEngine.java
kvoli: Good catch, removed.

hollow-explorer-ui/src/main/resources/explorer-header.vm
kvoli: Added a comment.

hollow-explorer-ui/src/main/resources/explorer-header.vm
kvoli: Good idea, updated.

hollow/src/main/java/com/netflix/hollow/core/read/engine/HollowReadStateEngine.java
kvoli: Done.

hollow/src/main/java/com/netflix/hollow/core/read/engine/HollowReadStateEngine.java
kvoli: Good catch, removed.

hollow-explorer-ui/src/main/resources/explorer-header.vm
kvoli: Added a code comment.

hollow-explorer-ui/src/main/resources/explorer-header.vm
kvoli: It isn't strictly necessary given we populate the type on every other page, other than the home page. It is a good idea for the future, in-case other pages are added which also don't populate $type. Updated to check `$topLevelTypes`.

Sunjeet (approved):

kvoli:
TYFTR!

PR #744, opened 2025-06-20
add null primary key validator

kvoli:
Null primary keys are not supported. Add a validator which can be used to detect when a record with a null primary key field is produced.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/validation/NullPrimaryKeyFieldValidator.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/validation/NullPrimaryKeyFieldValidator.java
@@ -1,0 +1,180 @@
+package com.netflix.hollow.api.producer.validation;
+
+import com.netflix.hollow.api.producer.HollowProducer;
+import com.netflix.hollow.core.index.key.PrimaryKey;
+import com.netflix.hollow.core.read.HollowReadFieldUtils;
+import com.netflix.hollow.core.read.engine.HollowReadStateEngine;
+import com.netflix.hollow.core.read.engine.object.HollowObjectTypeReadState;
+import com.netflix.hollow.core.schema.HollowObjectSchema;
+import com.netflix.hollow.core.schema.HollowSchema;
+import com.netflix.hollow.core.write.objectmapper.HollowPrimaryKey;
+import com.netflix.hollow.core.write.objectmapper.HollowTypeMapper;
+
+import java.util.Arrays;
+import java.util.BitSet;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+import static com.netflix.hollow.core.HollowConstants.ORDINAL_NONE;
+
+/**
+ * A validator that fails if any records of a given type have null primary key fields.
+ * <p>
+ * The primary key may be declared on a data model class using the {@link HollowPrimaryKey} annotation, may be
+ * declared more directly using {@link PrimaryKey} with {@link HollowObjectSchema}, or declared explicitly when
+ * instantiating this validator.
+ */
+public class NullPrimaryKeyFieldValidator implements ValidatorListener {
+    private static final String NAME = NullPrimaryKeyFieldValidator.class.getName();
+    private static final String NULL_PRIMARY_KEYS_FOUND_ERROR_MSG_FORMAT =
+            "Null primary key fields found for type %s. Primary Key in schema is %s. "
+                    + "Null records: [%s]";
+    private static final String NO_PRIMARY_KEY_ERROR_MSG_FORMAT =
+            "NullPrimaryKeyFieldValidator defined but unable to find primary key for data type %s. "
+                    + "Please check schema definition.";
+
+    private static final String NO_SCHEMA_FOUND_MSG_FORMAT =
+            "NullPrimaryKeyFieldValidator defined for data type %s but schema not found. "
+                    + "Please check that the HollowProducer is initialized with "
+                    + "the data type's schema (see initializeDataModel)";
+    private static final String NOT_AN_OBJECT_ERROR_MSG_FORMAT =
+            "NullPrimaryKeyFieldValidator is defined but schema type of %s is not Object. "
+                    + "This validation cannot be done.";
+
+    private static final String FIELD_PATH_NAME = "FieldPaths";
+    private static final String DATA_TYPE_NAME = "Typename";
+
+    private final String dataTypeName;
+
+    /**
+     * Creates a validator to detect records with null primary key fields for the type
+     * that corresponds to the given data type class annotated with {@link HollowPrimaryKey}.
+     *
+     * @param dataType the data type class
+     * @throws IllegalArgumentException if the data type class is not annotated with {@link HollowPrimaryKey}
+     */
+    public NullPrimaryKeyFieldValidator(Class<?> dataType) {
+        Objects.requireNonNull(dataType);
+
+        if (!dataType.isAnnotationPresent(HollowPrimaryKey.class)) {
+            throw new IllegalArgumentException("The data class " +
+                    dataType.getName() +
+                    " must be annotated with @HollowPrimaryKey");
+        }
+
+        this.dataTypeName = HollowTypeMapper.getDefaultTypeName(dataType);
+    }
+
+    /**
+     * Creates a validator to detect records with null primary keys of the type
+     * that corresponds to the given data type class annotated with {@link HollowPrimaryKey}.
+     * <p>
+     * The validator will fail, when {@link #onValidate validating}, if a schema with a
+     * primary key definition does not exist for the given data type name.
+     *
+     * @param dataTypeName the data type name
+     */
+    public NullPrimaryKeyFieldValidator(String dataTypeName) {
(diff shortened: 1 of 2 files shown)


hollow/src/main/java/com/netflix/hollow/core/index/key/HollowPrimaryKeyValueDeriver.java
```diff
@@ -144,11 +146,20 @@ private Object readValue(int ordinal, int fieldIdx) {
         int lastFieldPath = fieldPathIndexes[fieldIdx].length - 1;
         for (int i = 0; i < lastFieldPath; i++) {
             int fieldPosition = fieldPathIndexes[fieldIdx][i];
+            if (ordinal == ORDINAL_NONE) {
```
kvoli: This seemed like a logical place to add null checks, since `HollowPrimaryKeyValueDeriver` supports getting the resolved pkey value. But this is also used for indexing. Should the logic to resolve a pkey, while handling null fields exist here or somewhere else?

hollow/src/main/java/com/netflix/hollow/api/producer/validation/NullPrimaryKeyValidator.java
```diff
@@ -0,0 +1,197 @@
...
+
+
+    /**
+     * Creates a validator to detect records with null primary keys of the type
+     * that corresponds to the given data type class annotated with {@link HollowPrimaryKey}.
+     *
+     * @param dataType the data type class
+     * @throws IllegalArgumentException if the data type class is not annotated with {@link HollowPrimaryKey}
+     */
+    public NullPrimaryKeyValidator(Class<?> dataType) {
```
Sunjeet: nit: distinguish from primary key being null in name, for if we ever drop the constraints that fields comprising the pkey cant be null (but still constrain that pkey cant be null)

hollow/src/main/java/com/netflix/hollow/api/producer/validation/NullPrimaryKeyValidator.java
```diff
@@ -0,0 +1,197 @@
...
+     * Creates a validator to detect records with null primary keys of the type
+     * that corresponds to the given data type class annotated with {@link HollowPrimaryKey}.
+     * <p>
+     * The validator will fail, when {@link #onValidate validating}, if a schema with a primary key definition does not
+     * exist for the given data type name.
+     *
+     * @param dataTypeName the data type name
+     * @param fieldPathNames the field paths defining the primary key
+     */
+    public NullPrimaryKeyValidator(String dataTypeName, String[] fieldPathNames) {
```
Sunjeet: Can drop this constructor. A different validator (i'm thinking duplicate data detection) might have something like this to support combinations of fields other than the set of primary keys defined in the data model. Seems ok to restrict this one to primary keys defined in the data model.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/NullPrimaryKeyValidator.java
```diff
@@ -0,0 +1,197 @@
...
+     * {@code NullPrimaryKeyValidator} validator is not already registered for the same primary key type name).
+     *
+     * @param producer the producer
+     * @apiNote This method registers a {@code NullPrimaryKeyValidator} validator with only the primary key type
+     * name and not, in addition, the primary key fields.  This is to ensure, for the common case, duplicate listeners
+     * are not registered by this method if listeners with the same type names were explicitly registered when
+     * building the producer.
+     * @see HollowProducer#initializeDataModel(Class[])
+     */
+    public static void addValidatorsForSchemaWithPrimaryKey(HollowProducer producer) {
```
Sunjeet: Usage of this api relying on the user to call this after initializeDataModel and before runCycle doesn't seem great, maybe better to stick with the generic validator contract. We can drop this method, and explicitly invoke this validator for each type from elsewhere (cinder for netflix internal)

Fwiw the most common producer usage I've encountered is-
```
producer = producerBuilder
    .withListener(...)
    .withListener(...)
    .build();
producer.initializeDataModel(...);
producer.runCycle();
```

hollow/src/main/java/com/netflix/hollow/core/index/key/HollowPrimaryKeyValueDeriver.java
Sunjeet: Not feeling great about introducing branching in the happy path, for the validator use case. Maybe for the validator it would be better to iterate over the fields in a primary key, then for each field iterate over the ordinals and call read ordinal at every part of the field path to check if we get an ORDINAL_NONE back?

hollow/src/main/java/com/netflix/hollow/api/producer/validation/NullPrimaryKeyValidator.java
kvoli: Done.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/NullPrimaryKeyValidator.java
kvoli: Done.

hollow/src/main/java/com/netflix/hollow/core/index/key/HollowPrimaryKeyValueDeriver.java
kvoli: Updated to iterate over the primary key fields in the validator class instead of in the `HollowPrimaryKeyValueDeriver`.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/NullPrimaryKeyValidator.java
kvoli: Unsure I follow, could you explain further -- If the fields comprising the primary key are null, isn't that the same as the primary key being null?

azeng-netflix:
Looks good. Also wonder if it'd be helpful to directly add the null-check into the write path? I have briefly looked into it and think it might look like [this](https://github.com/Netflix/hollow/pull/745), probably also want to add a feature flag for it.

kvoli:
> Looks good. Also wonder if it'd be helpful to directly add the null-check into the write path? I have briefly looked into it and think it might look like [this](https://github.com/Netflix/hollow/pull/745), probably also want to add a feature flag for it.

Nice. Do you think it would be helpful to have both the validator and the null-check on the write path? With either enabled, the producer cycle would fail if I understand correctly.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/NullPrimaryKeyValidator.java
Sunjeet: this was just a little naming thing, to make the distinction between- any field comprising primary key being null vs. all fields being null

Sunjeet:
> > Looks good. Also wonder if it'd be helpful to directly add the null-check into the write path? I have briefly looked into it and think it might look like [this](https://github.com/Netflix/hollow/pull/745), probably also want to add a feature flag for it.
> 
> Nice. Do you think it would be helpful to have both the validator and the null-check on the write path? With either enabled, the producer cycle would fail if I understand correctly.

Re: enforcing in the write path- there maybe some rollout challenges there for e.g. incremental producer is known to duplicate primary keys in non top level types

hollow/src/main/java/com/netflix/hollow/api/producer/validation/NullPrimaryKeyValidator.java
kvoli: Updated the class name to `NullPrimaryKeyFieldValidator`. Let me know if that is what you had in mind.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/NullPrimaryKeyFieldValidator.java
```diff
@@ -0,0 +1,180 @@
...
+        vrb.detail(FIELD_PATH_NAME, fieldPaths);
+
+        Map<Integer, Object[]> ordinalToNullPkey = getNullPrimaryKeyValues(readState, primaryKey);
+        if (!ordinalToNullPkey.isEmpty()) {
+            return vrb.failed(String.format(NULL_PRIMARY_KEYS_FOUND_ERROR_MSG_FORMAT,
+                    dataTypeName, fieldPaths,
+                    nullKeysToString(ordinalToNullPkey)));
+        }
+
+        return vrb.passed();
```
spjegan: It would be good to add a message here, similar to [other](https://github.com/Netflix/hollow/blob/master/hollow/src/main/java/com/netflix/hollow/api/producer/validation/RecordCountPercentChangeValidator.java#L82-L85) validators

spjegan (approved):

hollow/src/main/java/com/netflix/hollow/api/producer/validation/NullPrimaryKeyFieldValidator.java
kvoli: Good idea, done.

kvoli:
@Sunjeet could take another pass when you get a chance?

Sunjeet (approved):

kvoli:
TYFTRs!

PR #746, opened 2025-06-24
Fix RecordCountPercentChangeValidator name

kvoli:
The name was erroneously set to the `RecordCountVarianceValidator` class name instead of `RecordCountPercentChangeValidator`.


spjegan (approved):

kvoli:
TYFTR!

PR #747, opened 2025-06-24
Include validator name in pass message for RecordCountPercentChangeValidator and RecordCountVarianceValidator

kvoli:
This is useful when debugging producer cycles, to view which validators have run successfully.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/validation/RecordCountPercentChangeValidator.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/validation/RecordCountPercentChangeValidator.java
@@ -30,7 +30,7 @@ public class RecordCountPercentChangeValidator implements ValidatorListener {
 
     @Override
     public String getName() {
-        return NAME;
+        return NAME + "_" + typeName;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


azeng-netflix (approved):

hollow/src/main/java/com/netflix/hollow/api/producer/validation/RecordCountPercentChangeValidator.java
```diff
@@ -79,7 +79,10 @@ private ValidationResult validateChanges(HollowTypeReadState typeState) {
                         (removedPercentageThreshold < 0 || removedPercent < removedPercentageThreshold) &&
                         (updatedPercentageThreshold < 0 || updatedPercent < updatedPercentageThreshold);
         if (pass) {
-            return builder.passed();
+            return builder.passed(String.format(
+                    "%s added=%.2f%% (<%.2f%%), removed=%.2f%% (<%.2f%%), updated=%.2f%% (<%.2f%%)",
+                    getName(), addedPercent * 100, addedPercentageThreshold * 100,  removedPercent * 100,
+                    removedPercentageThreshold * 100, updatedPercent * 100, updatedPercentageThreshold * 100));
```
spjegan, line 85: It would be nice to have the `type` here. Can you update the `getName` method to include the `type` also, similar to [this](https://github.com/Netflix/hollow/blob/master/hollow/src/main/java/com/netflix/hollow/api/producer/validation/RecordCountVarianceValidator.java#L90)

kvoli: Good idea, done.

spjegan (approved):

kvoli:
TYFTR!

PR #748, opened 2025-07-07
add max blob age to local filesystem blob retriever

kvoli:
Add `existingBlobMaxAge` to the local filesystem blob retriever, used in local development. When `existingBlobMaxAge` is configured, local blobs older than the conigured duration will be deleted and not used when retrieivng snapshot/header/delta blobs.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
@@ -42,6 +42,7 @@ import com.netflix.hollow.tools.history.HollowHistory;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.BitSet;
 import java.util.Collections;
@@ -1152,6 +1153,7 @@ public class HollowConsumer {
         protected HollowConsumer.ObjectLongevityDetector objectLongevityDetector = ObjectLongevityDetector.DEFAULT_DETECTOR;
         protected File localBlobStoreDir = null;
         protected boolean useExistingStaleSnapshot;
+        protected Duration existingBlobMaxAge = Duration.ofSeconds(-1); // Disabled by default, no existing blob will be deleted.
         protected Executor refreshExecutor = null;
         protected MemoryMode memoryMode = MemoryMode.ON_HEAP;
         protected HollowMetricsCollector<HollowConsumerMetrics> metricsCollector;
@@ -1210,6 +1212,25 @@ public class HollowConsumer {
          */
         public B withLocalBlobStore(String localBlobStoreDir, boolean useExistingStaleSnapshot) {
             return withLocalBlobStore(new File(localBlobStoreDir), useExistingStaleSnapshot);
+        }
+
+        /**
+         * Provide a directory that will be used to cache blobs.
+         *
+         * This is the same as {@link #withLocalBlobStore(File, boolean)} but also allows
+         * specifying the maximum age of an existing local blob. When a local blob is found
+         * that is older than the specified maximum age, it will not be used and Hollow will
+         * delete it, falling back to retrieving the blob using the configured blobRetriever.
+         *
+         * @param existingBlobMaxAge the maximum age an existing local blob can be before
+         *   it is deleted and not used. When set to null, or a negative duration, no maximum age
+         *   is applied and existing blobs are never deleted.
+         */
+        public B withLocalBlobStore(File localBlobStoreDir, boolean useExistingStaleSnapshot, Duration existingBlobMaxAge) {
+            this.localBlobStoreDir = localBlobStoreDir;
+            this.useExistingStaleSnapshot = useExistingStaleSnapshot;
+            this.existingBlobMaxAge = existingBlobMaxAge;
+            return (B)this;
         }
 
         public B withAnnouncementWatcher(HollowConsumer.AnnouncementWatcher announcementWatcher) {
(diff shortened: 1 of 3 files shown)


hollow/src/main/java/com/netflix/hollow/api/consumer/fs/HollowFilesystemBlobRetriever.java
```diff
@@ -312,7 +339,25 @@ private boolean allRequestedPartsExist(HollowConsumer.Blob.BlobType type, long c
...
+            for (Path blobPath : directoryStream) {
+                long lastModifiedTime = Files.getLastModifiedTime(blobPath).toMillis();
+                long currentTime = System.currentTimeMillis();
+                if ((currentTime - lastModifiedTime) > existingBlobMaxAge.toMillis()) {
+                    Files.delete(blobPath);
+                    LOG.info("Deleted old blob: " + blobPath + " last modified at " + lastModifiedTime);
+                }
+            }
+        } catch (IOException ex) {
+            throw new RuntimeException("Error listing blob files for deletion; path=" + blobStorePath, ex);
```
Sunjeet: nit: exception msg seems too specific e.g. the exception could also originate in `Files.delete`.. and it could also be useful to log the modified time seen on the file and the configured max age for deletion

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
```diff
@@ -1212,6 +1214,31 @@ public B withLocalBlobStore(String localBlobStoreDir, boolean useExistingStaleSn
...
+         *
+         * This is the same as {@link #withLocalBlobStore(File, boolean)} but also allows
+         * specifying the maximum age of an existing local blob. When a local blob is found
+         * that is older than the specified maximum age, it will not be used and Hollow will
+         * delete it, falling back to retrieving the blob using the configured blobRetriever.
+         *
+         * @param existingBlobMaxAge the maximum age an existing local blob can be before
+         *   it is deleted and not used.
+         */
+        public B withLocalBlobStore(File localBlobStoreDir, boolean useExistingStaleSnapshot, Duration existingBlobMaxAge) {
```
Sunjeet, line 1229: nit: maybe just the one that takes `File` seems sufficient. Not sure why we did both `File` and `String` variants earlier, guessing `File` came later

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
```diff
@@ -1152,6 +1153,7 @@ public static class Builder<B extends HollowConsumer.Builder<B>> {
         protected HollowConsumer.ObjectLongevityDetector objectLongevityDetector = ObjectLongevityDetector.DEFAULT_DETECTOR;
         protected File localBlobStoreDir = null;
         protected boolean useExistingStaleSnapshot;
+        protected Duration existingBlobMaxAge;
```
Sunjeet: nit: for readability consider setting to -1 explicitly and a comment that this makes max age infinite

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
```diff
@@ -1212,6 +1214,31 @@ public B withLocalBlobStore(String localBlobStoreDir, boolean useExistingStaleSn
...
 
+        /**
+         * Provide a directory that will be used to cache blobs.
+         *
+         * This is the same as {@link #withLocalBlobStore(File, boolean)} but also allows
+         * specifying the maximum age of an existing local blob. When a local blob is found
+         * that is older than the specified maximum age, it will not be used and Hollow will
+         * delete it, falling back to retrieving the blob using the configured blobRetriever.
+         *
+         * @param existingBlobMaxAge the maximum age an existing local blob can be before
```
Sunjeet, line 1225: nit: can document the case where a developer might want to set to negative or null

hollow/src/main/java/com/netflix/hollow/api/consumer/fs/HollowFilesystemBlobRetriever.java
```diff
@@ -120,6 +144,7 @@ private void ensurePathExists(Path blobStorePath) {
 
     @Override
     public HollowConsumer.HeaderBlob retrieveHeaderBlob(long desiredVersion) {
+        cleanupOldBlobs();
```
Sunjeet: We could scope to deletion to a prefix for e.g. "header-" here, so we're a bit closer to only deleting files we created

kvoli: Updated to scope the deletion to a prefix for each of the `retrieveXBlob` calls.

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
kvoli: Extended the param description further to cover the case where a null or negative duration is given.

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
kvoli: Done.

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
kvoli: Updated to just have the `File` constructor variant.

hollow/src/main/java/com/netflix/hollow/api/consumer/fs/HollowFilesystemBlobRetriever.java
kvoli: Updated the exception message to be more general. The file last modify time was already being logged above on deletion, updated to also include the configured max age for deletion.

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
Sunjeet: nit: could init to -1 here to for readability

Sunjeet (approved):

hollow/src/main/java/com/netflix/hollow/api/consumer/HollowConsumer.java
kvoli: Updated to -1s and added a comment.

kvoli:
TYFTR!

PR #749, opened 2025-07-08
fix resource leaks

skehrli:
This PR fixes a resource leak in the HollowBlobInput#randomAccess method.

This is done with a small class-internal refactor to make the resource leak freedom provable using the Resource Leak Checker from the Java Checker Framework. Objectively, this refactor makes the code cleaner. The annotations are not included, they would require a dependency on the checker framework.


kilink:
Thanks, I believe this was fixed in #753.

PR #753, opened 2025-07-28
Clean up HollowBlobInput

kilink:
Clean up HollowBlobInput: by making the input field a DataInput, many of the methods no longer need to have instanceof checks. Additionally, make all fields final.

Note, I think this also fixes the resource leak in #749, but the main motivator here was simplifying the code.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/read/HollowBlobInput.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/read/HollowBlobInput.java
@@ -9,6 +9,7 @@ import com.netflix.hollow.core.memory.MemoryMode;
 import com.netflix.hollow.core.memory.encoding.BlobByteBuffer;
 import java.io.ByteArrayInputStream;
 import java.io.Closeable;
+import java.io.DataInput;
 import java.io.DataInputStream;
 import java.io.File;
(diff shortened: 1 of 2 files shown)


hollow/src/main/java/com/netflix/hollow/core/read/HollowBlobInput.java
```diff
@@ -127,13 +133,7 @@ public static HollowBlobInput serial(InputStream is) {
...
      */
     public int read() throws IOException {
-        if (input instanceof RandomAccessFile) {
-            return ((RandomAccessFile) input).read();
-        } else if (input instanceof DataInputStream) {
-            return ((DataInputStream) input).read();
-        } else {
-            throw new UnsupportedOperationException("Unknown Hollow Blob Input type");
-        }
+        return input.readUnsignedByte();
```
Sunjeet, line 136: Thanks very much for this PR @kilink . The difference between `read()` and `readUnsignedByte` with regards to eof handling might have been the reason the code was originally written this way but upon looking closer, good call on the refactor, if `read()` returns -1 the callers seem to be translating it to exceptions anyway.

Sunjeet (approved):

PR #760, opened 2025-09-18
Impro/type resolver

yujiaaa2019:


hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/HollowObjectTypeMapper.java
```diff
@@ -226,7 +258,81 @@ protected Object parseHollowRecord(HollowRecord record) {
...
+                    int mostSigBitsPosInPojoSchema = schema.getPosition(MappedFieldType.UUID_MOST_SIG_BITS.getSpecialFieldName());
+                    if(mostSigBitsPosInPojoSchema != -1) {
+                        long value = hollowObject.getLong(MappedFieldType.UUID_MOST_SIG_BITS.getSpecialFieldName());
+                        if(value != Long.MIN_VALUE) {
+                            mostSigBits = value;
+                        }
+                    }
+                    int leastSigBitsPosInPojoSchema = schema.getPosition(MappedFieldType.UUID_LEAST_SIG_BITS.getSpecialFieldName());
+                    if(leastSigBitsPosInPojoSchema != -1) {
+                        long value = hollowObject.getLong(MappedFieldType.UUID_LEAST_SIG_BITS.getSpecialFieldName());
```
Sunjeet: not specifically for this PR, but a note on performance: since we already have a field position here I wonder if we can avoid the string match on field name again cc @eduardoramirez

hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/HollowObjectTypeMapper.java
```diff
@@ -226,7 +258,81 @@ protected Object parseHollowRecord(HollowRecord record) {
...
+                        if(value != Long.MIN_VALUE) {
+                            leastSigBits = value;
+                        }
+                    }
+                    if (mostSigBits != null && leastSigBits != null) {
```
Sunjeet: Since there are 2 long components, having Long.MIN_VALUE as a sentinel effectively means certain UUIDs where one of the long blocks is 2^64 are usable. Effecitvely, 2^64 + 2^64 UUIDs will fall into this edge case, unlikely to ever hit this edge case but wonder if it can be handled better

hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/HollowObjectTypeMapper.java
```diff
@@ -1117,7 +1326,14 @@ private static enum MappedFieldType {
         INLINED_STRING(FieldType.STRING),
         REFERENCE(FieldType.REFERENCE),
         ENUM_NAME(FieldType.STRING, "_name"),
-        DATE_TIME(FieldType.LONG, "value");
+        DATE_TIME(FieldType.LONG, "value"),
+        UUID_MOST_SIG_BITS(FieldType.LONG, "mostSigBits"),
```
Sunjeet: we should probably namespace the long components of UUID into their own custom type name, otherwise it will undo the compression benefits for all other long values in the dataset.. since encoding a long normally only #bits required to encode the largest long value, so if UUID is also using the long type then #bits required to encode any long will be 64.

hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/HollowObjectTypeMapper.java
```diff
@@ -80,7 +85,34 @@ public HollowObjectTypeMapper(HollowObjectMapper parentMapper, Class<?> clazz, S
...
-        } else {
+        } else if(clazz == Instant.class) {
+            try {
+                mappedFields.add(new MappedField(MappedFieldType.INSTANT_NANOS));
+                mappedFields.add(new MappedField(MappedFieldType.INSTANT_SECONDS));
+            }
+            catch(Exception e) {
+                throw new RuntimeException(e);
+            }
+        } else if(clazz == UUID.class) {
```
Sunjeet: do we plan to also implement these in the generated `HollowAPI`? Atm clients would need something like
```
com.netflix.foo.generated.UUID uuid = someHollowObject.getMyUuidField();
// System.out.println(uuid); // mostSigBits: -8232749396958297039 leastSigBits: -7121165017364359242
java.util.UUID javaUUID = new java.util.UUID(uuid.getMostSigBits(), uuid.getLeastSigBits());
System.out.println(javaUUID); // 8dbf660a-dd7d-4431-9d2c-8a25710f17b6
```

hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/HollowObjectTypeMapper.java
```diff
@@ -269,6 +375,85 @@ protected Object parseFlatRecord(FlatRecordTraversalNode node) {
...
+                        long secondsValue = objectNode.getFieldValueLong(MappedFieldType.INSTANT_SECONDS.getSpecialFieldName());
+                        if(secondsValue != Long.MIN_VALUE) {
+                            seconds = secondsValue;
+                        }
+                    }
+
+                    int nanoPosInPojoSchema = schema.getPosition(MappedFieldType.INSTANT_SECONDS.getSpecialFieldName());
+                    if(nanoPosInPojoSchema != -1) {
+                        int nanoValue = objectNode.getFieldValueInt(MappedFieldType.INSTANT_SECONDS.getSpecialFieldName());
+                        if(nanoValue != Long.MIN_VALUE) {
```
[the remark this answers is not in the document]
Sunjeet: Oh yeah this seems like a bug

hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectTypeReadState.java
```diff
@@ -204,6 +204,7 @@ public boolean isNull(int ordinal, int fieldIndex) {
             case FLOAT:
                 return (int)fixedLengthValue == HollowObjectWriteRecord.NULL_FLOAT_BITS;
             case DOUBLE:
+            case UUID_LONG:
```
yujiaaa2019: I added this field for avoiding making all LONGs to take 64 bits. But it will be backward incompatible for namespaces using UUID already. 
```
  1. Old producer (LONG) → New consumer (UUID_LONG): The consumer will see a field type mismatch and throw IncompatibleSchemaException
  2. New producer (UUID_LONG) → Old consumer (LONG): The old consumer will see a field type mismatch and throw IncompatibleSchemaException
```

PR #768, opened 2025-11-14
Throw NPE when a field is null during unique index construction

spjegan:
This PR handles null field values during UniqueKey index creation, by throwing `NullPointerException`. It also handles the null fields in the nested field paths in both PrimaryKey and UniqueKey indexes.


kvoli (approved):
LGTM

hollow/src/main/java/com/netflix/hollow/core/index/HollowUniqueKeyIndex.java
```diff
@@ -537,7 +538,7 @@ private synchronized void reindex() {
 
         int ordinal = ordinals.nextSetBit(0);
         while (ordinal != ORDINAL_NONE) {
-            int hashCode = generateRecordHash(ordinal);
+            int hashCode = generateRecordHash(ordinal, typeState.getSchema());
```
azeng-netflix: for all occurrences of  `typeState.getSchema()`, maybe store them in a variable to avoid potential extra lookup in the loops?

spjegan: Thanks. This is removed now as part of the fix to provide the correct type/field name in the exception.

Sunjeet (approved):

PR #769, opened 2025-11-14
[Improve]: Hollow API Generation Handling Missing Types;

azeng-netflix:
In some use cases, HollowProducer would be producing header blob for a dataset, which does not contain data for one or more types yet. Currently, 1. the code can run into `NullPointerException` when generating the API. 2. the user can later encounter `Cannot Find Symbol` compilation errors when using the generated API. 

Noted these undefined types as `undefinedType`, when API generators, like `HollowAPIGenerator`, generate files using a list of `HollowSchema` derived from the header blob, it generates classes containing methods (i.e. getters) referring to the `undefinedType`.

To work around this, by referring to the existing work in RAWHollow, we simply ignore the `undefinedType` during code generation. Also ignore any types taking the `undefinedType` as its type arguments.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/codegen/testdata/HollowTestDataAPIClassGenerator.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/codegen/testdata/HollowTestDataAPIClassGenerator.java
@@ -22,6 +22,8 @@ import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
 
+import static com.netflix.hollow.api.codegen.HollowCodeGenerationUtils.includeType;
+
 class HollowTestDataAPIClassGenerator {
     
     private final HollowDataset dataset;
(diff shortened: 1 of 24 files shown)


hollow/src/main/java/com/netflix/hollow/api/codegen/testdata/HollowTestDataAPIClassGenerator.java
```diff
@@ -47,6 +49,9 @@ public String generate() {
         schemas.sort(Comparator.comparing(HollowSchema::getName));
 
         for(HollowSchema schema : schemas) {
+            if (!includeType(schema, dataset)) {
```
azeng-netflix, line 52: Probably would skip the change for TestDataAPI generator, as it seems the current implementation have bugs... i.e. for a type name like `String`, the generated API would generate a method `String()` which conflicts with the java String... but leave it here for now..

To limit the change's scope, would put the fix in a separate PR.

Sunjeet (approved):

PR #779, opened 2025-12-18
initial commit of the proto adaptor

abersnaze:


hollow-protoadapter/src/main/java/com/netflix/hollow/protoadapter/HollowMessageMapper.java
```diff
@@ -0,0 +1,836 @@
...
+
+                    case ENUM:
+                        // Enums stored as strings
+                        schema.addField(fieldName, HollowObjectSchema.FieldType.STRING);
+                        break;
+
+                    case MESSAGE:
+                        // Nested messages are references
+                        String referencedType = field.getMessageType().getName();
+                        schema.addField(fieldName, HollowObjectSchema.FieldType.REFERENCE, referencedType);
```
Sunjeet: cycles?

PR #780, opened 2025-12-18
[Doc]: Remove docs on the deprecated prefix index;

azeng-netflix:
1. Remove the documentations on the deprecated prefix index to discourage the use of it.
2. Expose the `diving-deeper.md#keeping-an-index-up-to-date` part at the beginning of `indexing-querying.md` page.

### DIFF
--- a/docs/indexing-querying.md
+++ b/docs/indexing-querying.md
@@ -1,5 +1,7 @@
 # Indexing/Querying
 Hollow supports indexing of your dataset for quick retrieval. This guide demonstrates on usages of different indexes available and how they could be used to query for faster retrieval.
+However, indexes by default do not recalculate their state when the consumer refreshes to a new data state. As such, results for queries may become stale or corrupt if the consumer is refreshed.
+Refer to [Keeping an index up-to-date](diving-deeper.md#keeping-an-index-up-to-date) for more details.
 
 ## A Data Model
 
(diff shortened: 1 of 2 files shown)


docs/indexing-querying.md
```diff
@@ -46,7 +48,7 @@ Once we have loaded a dataset into a `HollowConsumer`, we can use the `Movie` in
 
 HollowConsumer consumer = ...;
 
-MoviePrimaryKeyIndex idx = new MoviePrimaryKeyIndex(consumer, true); // param isListenToDataRefresh set to true
+MoviePrimaryKeyIndex idx = new MoviePrimaryKeyIndex(consumer, true); // param isListenToDataRefresh set to true to enable listening to data refresh
```
Sunjeet, line 51: does this handle both, delta and snapshot refreshes? would be good to clarify here

Sunjeet: separately, do we also document the usage of 
consumer.addRefreshListener(yourIndex)? we should emphasize on that usage

azeng-netflix: Yes, the documentation you added here explained [this](https://hollow.how/diving-deeper/#indexing-data-for-retrieval), as well as emphasizing the limitation of `listenForDeltaUpdates`.
Would add code examples for this.

PR #784, opened 2026-01-09
Support for copying metric listeners from existing Producers

spjegan:


hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -746,6 +753,31 @@ public static class Builder<B extends HollowProducer.Builder<B>> {
...
+
+        private void overrideProducerMetricsListenerIfNeeded() {
+            if (customProducerMetricsListener == null) {
+                return;
+            }
+
+            // If the builder already has configured a producer metrics listener, do not override it.
+            boolean hasProducerMetricsListener = eventListeners.stream()
+                    .anyMatch(l -> l instanceof AbstractProducerMetricsListener);
+            if (!hasProducerMetricsListener) {
```
Sunjeet, line 776: just to confirm- if there already is a producer metrics listener, then we dont want to override it with customProducerMetricsListener? Is that the behavior we want?

spjegan: Yes, that was my thinking... if there is already a metrics listener we should not override it. Allowing that would cause the same issue with metrics, right? Let me know if you think otherwise.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -746,6 +753,31 @@ public static class Builder<B extends HollowProducer.Builder<B>> {
         ProducerOptionalBlobPartConfig optionalPartConfig = null;
         HollowConsumer.UpdatePlanBlobVerifier updatePlanBlobVerifier = HollowConsumer.UpdatePlanBlobVerifier.DEFAULT_INSTANCE;
 
+        protected HollowProducerEventListener customProducerMetricsListener = null;
+
+        public B withListenersFromExisting(HollowProducer producer) {
```
Sunjeet: nit: rename to indicate this is specific to the metrics listener, and ship it! Thanks

Sunjeet (approved):

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
spjegan: Done. Renamed this to `withMetricsListenerFromExisting`. Singular because we only copy one `MetricsListener` from the source producer as of now and it is possible there may be more than one (if added by clients using `withListener`).

PR #787, opened 2026-01-21
[Improv]: enable max ordinal check after write state engine prepare write;

azeng-netflix:
Summary:
  This change adds configurable soft limit validation to ByteArrayOrdinalMap and exposes ordinal map
  statistics to PublishListeners for monitoring.

  Changes:

  1. Soft Limit Enforcement
    - Adds soft limit threshold at 268,435,456 (2^28) - half of the hard limit
    - Throws exception when soft limit exceeded (configurable behavior)
    - Override mechanism via HollowProducer.withIgnoreSoftLimits(Supplier<Boolean>)
    - One-time logging per cycle to prevent log spam
  2. Stats Exposure
    - New ByteArrayOrdinalMapStats class captures maxOrdinal, byteDataLength, and loadFactor
    - New PublishStageStats wraps stats for all type write states
    - Extended PublishListener.onPublishComplete() to receive stats (backward compatible)
    - Stats collected after successful publish and passed through listener chain
    3. Perf Improvement
    - Remove redundant scan of `pointersAndOrdinals` to find the max ordinal in `HollowWriteTypeState`. Instead, find and return the max ordinal when doing `prepareForWrite`

  Backward Compatibility:
  - Default constructor maintains existing behavior (soft limits disabled by default)
  - Deprecated onPublishComplete method preserved with default implementation
  - No breaking changes to existing API

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
@@ -45,6 +45,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Executor;
+import java.util.function.Supplier;
 
 /**
  * A HollowProducer is the top-level class used by producers of Hollow data to populate, publish, and announce data states.
@@ -752,6 +753,7 @@ public class HollowProducer extends AbstractHollowProducer {
         boolean doIntegrityCheck = true;
         ProducerOptionalBlobPartConfig optionalPartConfig = null;
         HollowConsumer.UpdatePlanBlobVerifier updatePlanBlobVerifier = HollowConsumer.UpdatePlanBlobVerifier.DEFAULT_INSTANCE;
+        Supplier<Boolean> ignoreSoftLimits = null;
 
         protected HollowProducerEventListener customProducerMetricsListener = null;
 
@@ -970,6 +972,18 @@ public class HollowProducer extends AbstractHollowProducer {
 
         public B withUpdatePlanVerifier(HollowConsumer.UpdatePlanBlobVerifier updatePlanBlobVerifier) {
             this.updatePlanBlobVerifier = updatePlanBlobVerifier;
+            return (B) this;
+        }
+
+        /**
+         * set the ignoreSoftLimits. When set to true, ignore soft limits like
+         * {@link com.netflix.hollow.core.memory.ByteArrayOrdinalMap} ordinal limit, instead of failing.
+         * @param ignoreSoftLimits the supplier for whether to ignore
+         * {@link com.netflix.hollow.core.memory.ByteArrayOrdinalMap} ordinal soft limit breaches
+         * @return this builder
+         */
+        public B withIgnoreSoftLimits(Supplier<Boolean> ignoreSoftLimits) {
+            this.ignoreSoftLimits = ignoreSoftLimits;
             return (B) this;
         }
 
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/listener/PublishListener.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/listener/PublishListener.java
@@ -17,7 +17,9 @@
 package com.netflix.hollow.api.producer.listener;
 
 import com.netflix.hollow.api.producer.HollowProducer;
+import com.netflix.hollow.api.producer.PublishStageStats;
 import com.netflix.hollow.api.producer.Status;
+
 import java.time.Duration;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
@@ -71,7 +73,6 @@ public interface PublishListener extends HollowProducerEventListener {
      * @param elapsed time taken to publish the blob
      */
     void onBlobPublish(Status status, HollowProducer.Blob blob, Duration elapsed);
-
     /**
      * Called if a blob is to be published asynchronously.
      * This method is called for a {@link HollowProducer.Blob.Type#SNAPSHOT snapshot} blob when the
@@ -88,13 +89,34 @@ public interface PublishListener extends HollowProducerEventListener {
     }
 
     /**
-     * Called after the publish stage finishes normally or abnormally. A {@code SUCCESS} status indicates that
-     * the {@code HollowBlob}s produced this cycle has been published to the blob store.
+     * Deprecated in favor of
+     * {@code onPublishComplete(Status status, long version, Duration elapsed, PublishStageStats stats)}.
      *
      * @param status CycleStatus of the publish stage. {@link Status#getType()} will return {@code SUCCESS}
      * when the publish was successful; @{code FAIL} otherwise.
      * @param version version that was published.
      * @param elapsed duration of the publish stage in {@code unit} units
      */
+    @Deprecated
     void onPublishComplete(Status status, long version, Duration elapsed);
+
+    /**
+     * Called when the publish stage finishes normally or abnormally. A {@code SUCCESS} status indicates that
+     * the {@code HollowBlob}s produced this cycle has been published to the blob store.
+     * <p>
+     * This method provides access to publish stage statistics including ordinal map metrics for each type.
+     * The stats parameter will be {@code null} if the publish stage encountered errors before statistics
+     * could be collected (e.g., during blob staging or publishing failures).
+     *
+     * @param status CycleStatus of the publish stage. {@link Status#getType()} will return {@code SUCCESS}
+     * when the publish was successful; {@code FAIL} otherwise.
+     * @param version version that was published.
+     * @param elapsed duration of the publish stage
+     * @param stats stats collected upon publish stage completion, including
+     * {@link com.netflix.hollow.core.memory.ByteArrayOrdinalMapStats} for each type.
+     * Will be {@code null} if the stage encountered errors before stats could be collected.
+     */
+    default void onPublishComplete(Status status, long version, Duration elapsed, PublishStageStats stats) {
+        onPublishComplete(status, version, elapsed);
+    }
 }
--- a/hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
@@ -22,6 +22,8 @@ import com.netflix.hollow.core.memory.pool.WastefulRecycler;
 import java.util.Arrays;
 import java.util.BitSet;
 import java.util.concurrent.atomic.AtomicLongArray;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * This data structure maps byte sequences to ordinals.  This is a hash table.
@@ -37,6 +39,8 @@ import java.util.concurrent.atomic.AtomicLongArray;
  */
 public class ByteArrayOrdinalMap {
 
+    private static final Logger LOG = Logger.getLogger(ByteArrayOrdinalMap.class.getName());
+
     private static final long EMPTY_BUCKET_VALUE = -1L;
 
     private static final int BITS_PER_ORDINAL = 29;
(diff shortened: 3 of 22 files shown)


hollow/src/main/java/com/netflix/hollow/core/write/HollowTypeWriteState.java
```diff
@@ -72,8 +74,12 @@ public abstract class HollowTypeWriteState {
 
 
     public HollowTypeWriteState(HollowSchema schema, int numShards) {
+        this(schema, numShards, null);
+    }
+
+    public HollowTypeWriteState(HollowSchema schema, int numShards, Supplier<Boolean> ignoreOrdinalThresholdBreach) {
         this.schema = schema;
-        this.ordinalMap = new ByteArrayOrdinalMap();
+        this.ordinalMap = new ByteArrayOrdinalMap(256, ignoreOrdinalThresholdBreach);
```
azeng-netflix: 1. Don't leak 256, the ByteArrayOrdinalMap internal hard coded value, to TypeWriteState.
2. Need to check if `ignoreOrdinalThresholdBreach` is always properly passed into the ordinal map.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -973,6 +975,19 @@ public B withUpdatePlanVerifier(HollowConsumer.UpdatePlanBlobVerifier updatePlan
...
 
+        /**
+         * Configures the producer to use a supplier that determines whether to ignore ordinal threshold breaches.
+         * When the supplier returns true, a warning will be logged instead of throwing an exception
+         * when the SINGLE_PARTITION_ORDINAL_LIMIT is exceeded.
+         *
+         * @param ignoreOrdinalThresholdBreach the supplier that returns true to ignore threshold breaches
+         * @return this builder
+         */
+        public B withIgnoreOrdinalThresholdBreachConfig(Supplier<Boolean> ignoreOrdinalThresholdBreach) {
```
Sunjeet: Can make this more generic to all soft limits- `ignoreSoftLimits`

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
```diff
@@ -37,13 +39,20 @@
...
+
     private static final long EMPTY_BUCKET_VALUE = -1L;
 
     private static final int BITS_PER_ORDINAL = 29;
     private static final int BITS_PER_POINTER = Long.SIZE - BITS_PER_ORDINAL;
     private static final long POINTER_MASK = (1L << BITS_PER_POINTER) - 1;
     private static final long ORDINAL_MASK = (1L << BITS_PER_ORDINAL) - 1;
     private static final long MAX_BYTE_DATA_LENGTH = 1L << BITS_PER_POINTER;
+    // a safe limit for the ordinal. Even though the ordinal value range is [0, (1<<29) - 1], in production,
+    // the user should be warned whenever the ordinal value exceeds (1<<28) - 1.
```
Sunjeet: nit: drop "in production" and replace "the user should be warned" with something like- fail the producer cycle unless soft limits are overridden

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
```diff
@@ -154,6 +203,17 @@ private synchronized int assignOrdinal(ByteDataArray serializedRepresentation, i
                     ordinal, ORDINAL_MASK));
         }
 
+        if (ordinal >= SOFT_ORDINAL_LIMIT) {
+            if (logOrdinalThresholBreach && ignoreOrdinalThresholdBreach) {
+                logOrdinalThresholBreach = false;
```
Sunjeet: nit: typo in `logOrdinalThresholBreach` - but I guess it will be renamed to logSoftLimitBreach anyway

hollow/src/main/java/com/netflix/hollow/core/write/HollowTypeWriteState.java
```diff
@@ -277,6 +287,7 @@ public void resizeOrdinalMap(int size) {
      */
     public void prepareForNextCycle() {
         ordinalMap.compact(currentCyclePopulated, numShards, stateEngine.isFocusHoleFillInFewestShards());
+        ordinalMap.setIgnoreOrdinalThresholdBreach(ignoreOrdinalThresholdBreachSupplier == null ? false : ignoreOrdinalThresholdBreachSupplier.get());
```
Sunjeet: here and anywhere else where there isn't a way to override the soft limit failure, lets default to not failing on soft limits

hollow/src/main/java/com/netflix/hollow/api/producer/listener/PublishListener.java
```diff
@@ -72,6 +74,9 @@ default void onBlobStage(Status status, HollowProducer.Blob blob, Duration elaps
      */
     void onBlobPublish(Status status, HollowProducer.Blob blob, Duration elapsed);
 
+    default void onPublishHeaderBlob(HollowWriteStateEngine writeStateEngine) {
```
Sunjeet: cleanup?

azeng-netflix: This is used for listener to do customized report

Sunjeet: coupling with onPublishHeaderBlob doesn't feel great, can we explore reusing the onPopulateComplete listener method?

azeng-netflix: I have moved it to `onPublishComplete`.
The reason of not doing it in `onPopulateComplete`, is that the WriteStateEngine has not been prepared for write yet.

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
azeng-netflix: ✅

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
azeng-netflix: ✅

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
azeng-netflix: ✅

hollow/src/main/java/com/netflix/hollow/core/write/HollowTypeWriteState.java
azeng-netflix: ✅

azeng-netflix:
@azeng-netflix Before merge, check again to see if missing any places where the limit supplier are not properly propagated/copied over.

hollow/src/test/java/com/netflix/hollow/api/producer/HollowProducerTest.java
```diff
@@ -124,6 +134,23 @@ private HollowProducer createProducer(File tmpFolder,
         return producer;
     }
 
+     // check if ByteArrayOrdinalMap.ignoreSoftLimits has been updated to the expected value.
+    private void checkBaomIgnoreOrdinalSoftLimit(boolean expected, HollowProducer producer) {
```
azeng-netflix, line 137: NOTE: Modified this test file to check if ignore ordinal limit can be updated between cycles/restore

hollow/src/main/java/com/netflix/hollow/core/write/HollowTypeWriteState.java
```diff
@@ -460,6 +479,16 @@ public void gatherShardingStats(int maxOrdinal, boolean canReshard) {
         }
     }
 
+    /**
+     * Returns statistics for this instance {@code ordinalMap}.
+     * Should be invoked after {@code prepareForWrite}
+     *
+     * @return a {@link ByteArrayOrdinalMapStats} containing map statistics
+     */
+    public ByteArrayOrdinalMapStats getOrdinalMapStatsAfterPreparation() {
```
Sunjeet: nit: could simplify name to `getStats` , the clarification you have in javadoc feels sufficient

hollow/src/main/java/com/netflix/hollow/core/write/HollowWriteStateEngine.java
```diff
@@ -74,6 +75,8 @@ public class HollowWriteStateEngine implements HollowStateEngine {
     private boolean focusHoleFillInFewestShards = false;
     //// adjust number of shards per type during the course of the delta chain to realize consumer-side delta applications at constant space overhead
     private boolean allowTypeResharding = false;
+    //// supplier to determine whether to ignore ordinal threshold breaches (log warning instead of throwing exception)
+    private Supplier<Boolean> ignoreOrdinalSoftLimitsSupplier;
```
Sunjeet: nit: name this more generic `ignoreSoftLimitsSupplier`, becuase we would reuse it for any other limits. Could also drop supplier from the name

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
```diff
@@ -154,6 +205,22 @@ private synchronized int assignOrdinal(ByteDataArray serializedRepresentation, i
...
+        if (ordinal >= SOFT_ORDINAL_LIMIT) {
+            if (ignoreSoftLimits) {
+                if (logSoftLimitsBreach) {
+                    logSoftLimitsBreach = false;
+                    LOG.log(Level.INFO, String.format("Ordinal %d exceeds the soft ordinal limit of %d",
+                            ordinal, SOFT_ORDINAL_LIMIT));
+                }
+                // continue silently after first log
+            } else {
+                String message = String.format("Ordinal %d exceeds the soft ordinal limit of %d.",
```
Sunjeet: nit: adding a hint in the log msg about support for ignoring soft limits could be useful

hollow/src/main/java/com/netflix/hollow/core/write/HollowListTypeWriteState.java
```diff
@@ -48,11 +49,19 @@ public class HollowListTypeWriteState extends HollowTypeWriteState {
     public HollowListTypeWriteState(HollowListSchema schema) {
         this(schema, -1);
     }
+
+    public HollowListTypeWriteState(HollowListSchema schema, Supplier<Boolean> ignoreOrdinalSoftLimitsSupplier) {
```
Sunjeet: I dont think we really need this new constructor, could just reuse the other one you added?

hollow/src/main/java/com/netflix/hollow/core/write/HollowListTypeWriteState.java
```diff
@@ -48,11 +49,19 @@ public class HollowListTypeWriteState extends HollowTypeWriteState {
...
+        this(schema, -1, ignoreOrdinalSoftLimitsSupplier);
+    }
     
     public HollowListTypeWriteState(HollowListSchema schema, int numShards) {
         super(schema, numShards);
```
Sunjeet: nit: changing to `this(schema, numShards, null)` would be a bit more readable

hollow/src/main/java/com/netflix/hollow/api/producer/listener/PublishListener.java
```diff
@@ -97,4 +98,18 @@ default void onBlobPublishAsync(CompletableFuture<HollowProducer.Blob> blob) {
...
+     * the {@code HollowBlob}s produced this cycle has been published to the blob store.
+     * This overload provides access to the write state engine.
+     *
+     * @param writeStateEngine the write state engine used during this cycle
+     * @param status CycleStatus of the publish stage. {@link Status#getType()} will return {@code SUCCESS}
+     * when the publish was successful; @{code FAIL} otherwise.
+     * @param version version that was published.
+     * @param elapsed duration of the publish stage in {@code unit} units
+     */
+    default void onPublishComplete(HollowWriteStateEngine writeStateEngine, Status status, long version, Duration elapsed) {
```
Sunjeet: An alternative would be to add a Stats within the State object, where different stages like publish, etc. could implement their stats like PublishStage could implement stats like duration which could be generic to all stages, or specific ones like ordinap map stats for publish stage. If you think it fits well, otherwise the current approach seems fine too, but if we're doing this then lets also deprecate the older 3 param onPublishComplete, and make this default impl call that one. Could also add to the javadoc- The default implementation delegates  to the deprecated 3 param onPublishComplete for backwards compatibility.

azeng-netflix: Made this commit to address the [comment](https://github.com/Netflix/hollow/pull/787/changes/ef35557198f6fe28c24d2b0859e8346f91ae148e), but during the implementation, I think a general `Stats` would require the Listener side to do type casting like `stats instanceof PublishStageStats`. would like your opinion on this.

hollow/src/main/java/com/netflix/hollow/core/write/HollowListTypeWriteState.java
azeng-netflix: ✅

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
azeng-netflix: ✅

hollow/src/main/java/com/netflix/hollow/core/write/HollowWriteStateEngine.java
azeng-netflix: Changed to `ignoreSoftLimits`

hollow/src/main/java/com/netflix/hollow/core/write/HollowTypeWriteState.java
azeng-netflix: Changed to `getOrdinalMapStats` in case in the future other stats also need to be exposed from `HollowTypeWriteState`

hollow/src/main/java/com/netflix/hollow/api/producer/listener/PublishListener.java
azeng-netflix: add and use `PublishListener.onPublishComplete(Status status, long version, Duration elapsed, PublishStageStats stats)`

Sunjeet (approved):

PR #789, opened 2026-01-26
[Improv]: max ordinal check, rely on Listener;

azeng-netflix:


azeng-netflix:
Close as prefer the approach in #787

PR #790, opened 2026-01-28
Fixes OOM during error message construction

spjegan:
The duplicate detection finds pair-wise duplicates, which means if there are `1000` duplicate keys, it will find `499,500` duplicate key pairs (`(n * n-1) / 2`). The validator then tries to construct a message with all the duplicate pairs as single error message. This results in the below OOM error message.
```
java.lang.OutOfMemoryError: Requested array size exceeds VM limit
  at java.util.StringJoiner.checkAddLength(StringJoiner.java:203)
  at java.util.StringJoiner.add(StringJoiner.java:194)
  at java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169)
  at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
  at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1708)
  at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
  at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
  at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
  at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
  at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)
  at com.netflix.hollow.api.producer.validation.DuplicateDataDetectionValidator.duplicateKeysToString(DuplicateDataDetectionValidator.java:192)
  at com.netflix.hollow.api.producer.validation.DuplicateDataDetectionValidator.onValidate(DuplicateDataDetectionValidator.java:155)
  at com.netflix.cinder.producer.migration.AbstractPrimaryKeyBackgroundValidator.validateType(AbstractPrimaryKeyBackgroundValidator.java:126)
  at com.netflix.cinder.producer.migration.AbstractPrimaryKeyBackgroundValidator.onValidate(AbstractPrimaryKeyBackgroundValidator.java:85)
```

This PR fixes this by limiting the number of duplicate keys included in the error message to a max of 100. It also adds a similar limit on the error message for the `NullPrimaryKeyValidator` as well.

One potential improvement in the future could be to only compute the MAX number of duplicates (100 for instance), instead of computing everything and then displaying only the 100.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
@@ -46,6 +46,7 @@ import java.util.stream.Collectors;
  * record's type will never fail.
  */
 public class DuplicateDataDetectionValidator implements ValidatorListener {
+    private static final int MAX_DISPLAYED_DUPLICATE_KEYS = 100;
     private static final String DUPLICATE_KEYS_FOUND_ERRRO_MSG_FORMAT =
             "Duplicate keys found for type %s. Primarykey in schema is %s. "
                     + "Duplicate IDs are: %s";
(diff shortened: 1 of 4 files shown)


kvoli (approved):
LGTM

azeng-netflix:
Think we'd need to have an override `hollowPrimaryKeyIndex.getDuplicateKeys` that can returns early to bound the compute and memory usage.

spjegan:
> Think we'd need to have an override `hollowPrimaryKeyIndex.getDuplicateKeys` that can returns early to bound the compute and memory usage.

The duplicate detection can definitely be improved. One option is to convert this to a stream so we only compute what is needed (up to a max limit of say 100).

Sunjeet (approved):

PR #793, opened 2026-02-04
Optimize the memory used for finding duplicate keys

AustinWheel:


hollow/src/main/java/com/netflix/hollow/core/index/HollowPrimaryKeyIndex.java
```diff
@@ -382,26 +382,37 @@ public boolean containsDuplicates() {
     /**
      * @return any keys which are mapped to two or more records.
      */
-    public synchronized Collection<Object[]> getDuplicateKeys() {
+    public synchronized Collection<DuplicateKeyInfo> getDuplicateKeys() {
```
spjegan: This method is used internally. Modifying the signature/behavior could cause issues to those usages. I suggest to introduce a new over-loaded version of this method that accepts a max limit. We can then cap the duplicate detection up to that limit and stop the computation early.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -189,18 +190,23 @@ public int hashCode() {
...
-    private String duplicateKeysToString(Collection<Object[]> duplicateKeys) {
-        long totalCount = duplicateKeys.size();
+    private String duplicateKeysToString(Collection<DuplicateKeyInfo> duplicateKeys) {
+        long totalUniqueKeys = duplicateKeys.size();
+        long totalAffectedRecords = duplicateKeys.stream()
+                .mapToLong(DuplicateKeyInfo::getCount)
+                .sum();
+
         String duplicateKeysString = duplicateKeys.stream()
                 .limit(MAX_DISPLAYED_DUPLICATE_KEYS)
```
spjegan: This limit is not needed now as we are already fetching only the `MAX_DISPLAYED_DUPLICATE_KEYS` above.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -189,18 +190,23 @@ public int hashCode() {
...
-                .map(Arrays::toString)
+                .map(DuplicateKeyInfo::toString)
                 .collect(Collectors.joining(", "));
 
-        if (totalCount > MAX_DISPLAYED_DUPLICATE_KEYS) {
-            return String.format("%s ... (showing %d of %d duplicates)",
-                    duplicateKeysString, MAX_DISPLAYED_DUPLICATE_KEYS, totalCount);
+        if (totalUniqueKeys > MAX_DISPLAYED_DUPLICATE_KEYS) {
+            return String.format("%s ... (%d unique duplicate keys affecting %d total records, showing %d)",
```
spjegan: nit: can we please rephrase this? `unique` and `duplicate` together sounds a bit confusing. May be something like `(%d distinct keys that each have duplicate records affecting %d records; showing %d keys)`?

spjegan (approved):
LGTM

PR #795, opened 2026-02-10
Use protobuf-gradle-plugin and pin the build to use protoc version 3.+

spjegan:


abersnaze:
I would like to avoid the gradle plugin because it is intended for paved path building artifacts and linking to source sets for usage in the rest of the code.

abersnaze:
https://github.com/Netflix/hollow/pull/797

PR #798, opened 2026-02-11
[Improv]: Enable using 4 ByteArrayOrdinalMap per HollowTypeWriteState;

azeng-netflix:
Summary: Partitioned Ordinal Map Feature (4x Ordinal Limit)

  This change introduces the ability to partition a HollowTypeWriteState's ByteArrayOrdinalMap into 4 separate maps instead of 1, effectively quadrupling the ordinal limit (record cardinality limit) per type.

  Key changes:

  1. New builder option (HollowProducer / AbstractHollowProducer):
  - Added withPartitionedOrdinalMap(boolean) to the producer builder
  - Propagated through AbstractHollowProducer constructor → HollowWriteStateEngine → each HollowTypeWriteState

  2. Core: HollowTypeWriteState (biggest change):
  - Instead of a single ordinalMap (ByteArrayOrdinalMap), now supports an array of maps (ordinalMaps[])
  - Ordinals are routed to a specific map partition based on hash bits (2 bits → 4 partitions)
  - New helper methods getPointerForData(ordinal) and getByteDataForOrdinal(ordinal) abstract over single vs. multi-map access
  - All four type write states (Object, List, Set, Map) updated to use these helpers instead of directly accessing ordinalMap
  - Deduplication across all maps -- When adding a record, all maps are searched before assigning to prevent duplicates across partitions

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
@@ -102,6 +102,7 @@ abstract class AbstractHollowProducer {
     private final boolean allowTypeResharding;
     private final boolean forceCoverageOfTypeResharding;   // exercise re-sharding often (for testing)
     private final Supplier<Boolean> ignoreSoftLimits;
+    private final boolean partitionedOrdinalMap;
 
     @Deprecated
     public AbstractHollowProducer(
@@ -110,7 +111,7 @@ abstract class AbstractHollowProducer {
         this(new HollowFilesystemBlobStager(), publisher, announcer,
                 Collections.emptyList(),
                 new VersionMinterWithCounter(), null, 0,
-                DEFAULT_TARGET_MAX_TYPE_SHARD_SIZE, false, false, false, null,
+                DEFAULT_TARGET_MAX_TYPE_SHARD_SIZE, false, false, false, false, null,
                 new DummyBlobStorageCleaner(), new BasicSingleProducerEnforcer(),
                 null, true, HollowConsumer.UpdatePlanBlobVerifier.DEFAULT_INSTANCE, null);
     }
@@ -123,7 +124,7 @@ abstract class AbstractHollowProducer {
                 b.eventListeners,
                 b.versionMinter, b.snapshotPublishExecutor,
                 b.numStatesBetweenSnapshots, b.targetMaxTypeShardSize, b.focusHoleFillInFewestShards,
-                b.allowTypeResharding, b.forceCoverageOfTypeResharding,
+                b.allowTypeResharding, b.forceCoverageOfTypeResharding, b.partitionedOrdinalMap,
                 b.metricsCollector, b.blobStorageCleaner, b.singleProducerEnforcer,
                 b.hashCodeFinder, b.doIntegrityCheck, b.updatePlanBlobVerifier, b.ignoreSoftLimits);
     }
@@ -145,6 +146,7 @@ abstract class AbstractHollowProducer {
             boolean focusHoleFillInFewestShards,
             boolean allowTypeResharding,
             boolean forceCoverageOfTypeResharding,
+            boolean partitionedOrdinalMap,
             HollowMetricsCollector<HollowProducerMetrics> metricsCollector,
             HollowProducer.BlobStorageCleaner blobStorageCleaner,
             SingleProducerEnforcer singleProducerEnforcer,
@@ -166,6 +168,7 @@ abstract class AbstractHollowProducer {
         this.forceCoverageOfTypeResharding = forceCoverageOfTypeResharding;
         this.focusHoleFillInFewestShards = focusHoleFillInFewestShards;
         this.ignoreSoftLimits = ignoreSoftLimits;
+        this.partitionedOrdinalMap = partitionedOrdinalMap;
 
         HollowWriteStateEngine writeEngine = hashCodeFinder == null
                 ? new HollowWriteStateEngine()
@@ -174,6 +177,7 @@ abstract class AbstractHollowProducer {
         writeEngine.allowTypeResharding(allowTypeResharding);
         writeEngine.setFocusHoleFillInFewestShards(focusHoleFillInFewestShards);
         writeEngine.setIgnoreOrdinalLimits(ignoreSoftLimits);
+        writeEngine.setPartitionedOrdinalMap(partitionedOrdinalMap);
 
         this.objectMapper = new HollowObjectMapper(writeEngine);
         if (hashCodeFinder != null) {
@@ -322,6 +326,7 @@ abstract class AbstractHollowProducer {
                 client.triggerRefreshTo(versionInfoDesired);
                 readState = ReadStateHelper.newReadState(client.getCurrentVersionId(), client.getStateEngine());
                 readStates = ReadStateHelper.restored(readState);
+                verifyPartitionOrdinalMapCompatability();
 
                 // Need to restore data to new ObjectMapper since can't restore to non empty Write State Engine
                 Collection<HollowSchema> schemas = objectMapper.getStateEngine().getSchemas();
@@ -332,6 +337,7 @@ abstract class AbstractHollowProducer {
                 writeEngine.allowTypeResharding(allowTypeResharding);
                 writeEngine.setFocusHoleFillInFewestShards(focusHoleFillInFewestShards);
                 writeEngine.setIgnoreOrdinalLimits(ignoreSoftLimits);
+                writeEngine.setPartitionedOrdinalMap(partitionedOrdinalMap);
                 HollowWriteStateCreator.populateStateEngineWithTypeWriteStates(writeEngine, schemas);
                 HollowObjectMapper newObjectMapper = new HollowObjectMapper(writeEngine);
                 if (hashCodeFinder != null) {
@@ -586,6 +592,7 @@ abstract class AbstractHollowProducer {
         }
         long deltaChainVersionCounter = prevDeltaChainVersionCounter + 1;
         writeEngine.addHeaderTag(HEADER_TAG_DELTA_CHAIN_VERSION_COUNTER, String.valueOf(deltaChainVersionCounter));
+        writeEngine.addHeaderTag(HollowStateEngine.HEADER_TAG_PARTITIONED_ORDINAL_MAP, Boolean.toString(partitionedOrdinalMap));
     }
 
     void populate(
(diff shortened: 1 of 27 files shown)


hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
```diff
@@ -480,6 +505,7 @@ public void compact(ThreadSafeBitSet usedOrdinals, int numShards, boolean focusH
         byteData.setPosition(currentCopyPointer);
 
         if(focusHoleFillInFewestShards && numShards > 1)
+            // TODO: fix to be compatible with multi-ByteArrayOrdinalMap
```
azeng-netflix: Currently, making the multi-ByteArrayOrdinalMap feature not allowed to be enabled with focusHoleFillInFewestShards at the same time.
Can improve on this later, as I don't see much use of focusHoleFillInFewestShards checking sourcegraph.

hollow/src/main/java/com/netflix/hollow/core/write/HollowSetTypeWriteState.java
```diff
@@ -41,29 +41,25 @@ public class HollowSetTypeWriteState extends HollowTypeWriteState {
...
+        this(schema, -1, false, null);
     }
 
-    public HollowSetTypeWriteState(HollowSetSchema schema, int numShards) {
-        this(schema, numShards, null);
-    }
-
-    public HollowSetTypeWriteState(HollowSetSchema schema, int numShards, Supplier<Boolean> ignoreSoftLimits) {
-        super(schema, numShards, ignoreSoftLimits);
+    public HollowSetTypeWriteState(HollowSetSchema schema, int numShards, boolean usePartitionedOrdinalMap, Supplier<Boolean> ignoreSoftLimits) {
```
azeng-netflix, line 66: Wonder if we still need to keep the 2 param ctor `public HollowSetTypeWriteState(HollowSetSchema schema, int numShards)`, as they were marked as `public`

azeng-netflix: Or maybe just use `@Deprecated` to be safe for such cases..

hollow/src/main/java/com/netflix/hollow/core/write/HollowWriteStateEngine.java
```diff
@@ -78,6 +81,8 @@ public class HollowWriteStateEngine implements HollowStateEngine {
     private boolean allowTypeResharding = false;
     //// supplier to determine whether to ignore ordinal threshold breaches (log instead of throwing exception)
     private Supplier<Boolean> ignoreSoftLimits;
+    //// partition each type into 4 {@link com.netflix.hollow.core.memory.ByteArrayOrdinalMap} instead of 1
+    private boolean partitionedOrdinalMap = false;
```
azeng-netflix, line 82: Probably not a good namings, but leave it for now.
Some candidates, `multiOrdinalMap`, `increasedOrdinalLimit`..

azeng-netflix:
@azeng-netflix Need to check if `partitionedOrdinalMap` is properly set for `HollowWriteStateEngine`, (i.e. some move constructor like ops).

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -586,6 +592,7 @@ private void updateHeaderTags(HollowWriteStateEngine writeEngine, long toVersion
         }
         long deltaChainVersionCounter = prevDeltaChainVersionCounter + 1;
         writeEngine.addHeaderTag(HEADER_TAG_DELTA_CHAIN_VERSION_COUNTER, String.valueOf(deltaChainVersionCounter));
+        writeEngine.addHeaderTag(HollowStateEngine.HEADER_TAG_PARTITIONED_ORDINAL_MAP, Boolean.toString(partitionedOrdinalMap));
```
azeng-netflix, line 594: Noticed that some producers, (like the producer of VMS), seemed to be updating headers a bit different, need to verify the usages overthere.

hollow/src/main/java/com/netflix/hollow/core/write/HollowMapTypeWriteState.java
```diff
@@ -42,29 +42,30 @@ public class HollowMapTypeWriteState extends HollowTypeWriteState {
...
 
+    @Deprecated
     public HollowMapTypeWriteState(HollowMapSchema schema, int numShards) {
-        this(schema, numShards, null);
+        this(schema, numShards, false, null);
     }
 
-    public HollowMapTypeWriteState(HollowMapSchema schema, int numShards, Supplier<Boolean> ignoreSoftLimits) {
-        super(schema, numShards, ignoreSoftLimits);
+    public HollowMapTypeWriteState(HollowMapSchema schema, int numShards, boolean usePartitionedOrdinalMap, Supplier<Boolean> ignoreSoftLimits) {
```
azeng-netflix, line 67: NOTE: Changed this public constructor as it was added in recently (in Feb), and not used by others. So it's safe to modify.
Same for the cotr changes in other HollowTypeWriteState subclassed.

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
azeng-netflix: NOTE: VMSTransformer would need to enable this via `HollowWriteStateEngine`.

azeng-netflix (commented):
Remove the [artificially added incompatibility](https://github.com/Netflix/hollow/pull/798/changes#diff-f81e4e94ac63f49d77b5198dbe9728709c0f01af7fa2e2ad500c6362b47650b2R987) that prohibits a producer with 4x enabled to restored to a dataset without the 4x enabled (or vice versa).

Previously I made the wrong assumption that such change would cause a lot of records getting assigned to a new ordinal, while in fact, this won't happen because of the special handling done in the restore step, making sure the same records would get assigned to the same ordinals.

hollow/src/main/java/com/netflix/hollow/core/write/HollowTypeWriteState.java
```diff
@@ -117,95 +138,148 @@ public int add(HollowWriteRecord rec) {
...
+        // Search ALL maps for existing record (dedup across maps)
+        int existingGlobal = searchAllMaps(ordinalMaps, scratch, hash);
+        if (existingGlobal != -1) {
+            scratch.reset();
+            return existingGlobal;
+        }
+
+        // Not found — assign in hash-routed map
+        int mapIndex = hash & ordinalMapIndexMask;
+        // TODO: getOrAssignOrdinal performs a redundant check to see if rec exists in the specified
```
Sunjeet: cleanup?

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
```diff
@@ -480,18 +504,19 @@ public void compact(ThreadSafeBitSet usedOrdinals, int numShards, boolean focusH
...
         if(focusHoleFillInFewestShards && numShards > 1)
-            freeOrdinalTracker.sort(numShards);
+            freeOrdinalTracker.sort(numShards, mapIndexBits, mapIdx);
         else
             freeOrdinalTracker.sort();
 
         // Reset the array then fill with compacted values
         // Volatile store not required, could use plain store
         // See VarHandles for JDK >= 9
+        // TODO: why not just swap pao with populatedReverseKeys?
```
Sunjeet: cleanup?

hollow/src/main/java/com/netflix/hollow/core/write/HollowTypeWriteState.java
```diff
@@ -117,95 +138,148 @@ public int add(HollowWriteRecord rec) {
         return ordinal;
     }
 
+    /**
+     * Search all maps in the given array for a record matching the scratch bytes.
+     * @return the global ordinal if found, or -1 if not found
+     */
+    private int searchAllMaps(ByteArrayOrdinalMap[] maps, ByteDataArray scratch, int hash) {
```
Sunjeet, line 145: optimization: can check the hash-routed map first, for any record added since schema change (hash-true routed) it will hit

hollow/src/main/java/com/netflix/hollow/core/write/HollowTypeWriteState.java
```diff
@@ -117,95 +138,148 @@ public int add(HollowWriteRecord rec) {
...
+        int hash = HashCodes.hashCode(scratch);
+
+        // Search ALL maps for existing record (dedup across maps)
+        int existingGlobal = searchAllMaps(ordinalMaps, scratch, hash);
+        if (existingGlobal != -1) {
+            scratch.reset();
+            return existingGlobal;
+        }
+
+        // Not found — assign in hash-routed map
```
Sunjeet, line 180: nit: maybe worth clarifying in a commit here that the search for existence checks all maps, but it still valuable to have hash-routing (vs. say roundrobin or adding to the map with the lowest load etc) here to solve for- concurrently adding the same record from 2 threads- if we didn't hash to the same map then we'd need additional synchronization to solve for that

hollow/src/main/java/com/netflix/hollow/core/write/HollowTypeWriteState.java
azeng-netflix: Made a small optimization to avoid an extra search for the record in the `ordinalMaps[mapIndex]`.
However, by doing that I changed `ByteArrayOrdinalMap.assignOrdinal` from `private` to `public`, would like your input on whether it is worth it.

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
azeng-netflix: ✅

hollow/src/main/java/com/netflix/hollow/core/write/HollowTypeWriteState.java
azeng-netflix: I think even when adding the same records concurrently, the `ByteArrayOrdinalMap.assignOrdinal` already to perform the double-check with proper locking and should handle it?

One advantage of using hash routing makes the optimization in searchAllMaps that you mentioned in the pervious comment possible.

Sunjeet (approved):

PR #801, opened 2026-03-10
[fix] holes cost estimation;

azeng-netflix:
## Summary
 - Fix incorrect hole cost estimation in `getApproximateHoleCostInBytes` for all type shards (list, map, set, object)

 ## Details
 When `numShards > 1`, ordinals are distributed across shards via `ordinal % numShards`, and each shard stores records at shard-local ordinal `ordinal / numShards` (equivalently `ordinal >>> shardOrdinalShift`). The `maxOrdinal` field in `dataElements` is this shard-local max.

For one internal namespace, the difference between the actual value and the estimated one is 8X.

 The bug: the while-loop condition `holeOrdinal <= dataElements.maxOrdinal` compared a global ordinal against a shard-local max, causing the loop to terminate too early and undercount holes. For example, with 8 shards and maxOrdinal=2 (shard-local), the loop would stop at global ordinal 2 instead of scanning up to global ordinal ~23.

 The fix: `(holeOrdinal >>> shardOrdinalShift) <= dataElements.maxOrdinal` correctly converts to shard-local space.

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,0 +1,8 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(./gradlew :hollow-protoadapter:test:*)",
+      "Bash(./gradlew:*)"
+    ]
+  }
+}
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,8 +1,0 @@
-{
-  "permissions": {
-    "allow": [
-      "Bash(./gradlew :hollow-protoadapter:test:*)",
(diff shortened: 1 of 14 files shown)


Sunjeet:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Sunjeet (approved):

PR #802, opened 2026-03-10
Speed up duplicate data validator

AustinWheel:
Adds an opt-in incremental (delta-aware) mode to DuplicateDataDetectionValidator that validates only new ordinals each cycle against a retained primary-key index, trading steady-state memory for cycle-time savings on large datasets. Opt-in via a new constructor that takes a Supplier<Boolean>, which doubles as a runtime kill switch so operators can disable it via dynamic config without redeploying. Handles producer restore correctly by dropping the lagged index on onProducerRestoreStart; default behavior is unchanged for existing call sites.

Here are results from the benchmark when we use just the delta to check for duplicates:
```
  Records   Delta %   deltaBounded   snapshotFullScan (baseline)   Speedup                                                                                                                   
       1M      0.1%      0.079 ms                       30.0 ms     ~380x                                                              
       1M        1%      0.801 ms                       30.2 ms      ~38x                                                                                                                    
       1M       10%      9.747 ms                       30.0 ms       ~3x                                                                                                                    
       5M      0.1%      0.593 ms                      249.1 ms     ~420x                                                                                                                    
       5M        1%      8.905 ms                      294.6 ms      ~33x                                                              
       5M       10%    130.499 ms                      276.5 ms       ~2x
```

And this tracks as the small the delta % the less ordinals there are to check, and those gains are multiplied by the size of the dataset.

Full output:
```
Benchmark                                                  (deltaFraction)  (numRecords)  Mode  Cnt    Score    Error  Units                                                                 
DuplicateDataDetectionValidatorBenchmark.deltaBounded                0.001       1000000  avgt   20    0.079 ±  0.001  ms/op                                                                 
DuplicateDataDetectionValidatorBenchmark.deltaBounded                0.001       5000000  avgt   20    0.593 ±  0.018  ms/op                                                                 
DuplicateDataDetectionValidatorBenchmark.deltaBounded                 0.01       1000000  avgt   20    0.801 ±  0.041  ms/op                                                                 
DuplicateDataDetectionValidatorBenchmark.deltaBounded                 0.01       5000000  avgt   20    8.905 ±  0.362  ms/op                                                                 
DuplicateDataDetectionValidatorBenchmark.deltaBounded                  0.1       1000000  avgt   20    9.747 ±  0.412  ms/op                                                                 
DuplicateDataDetectionValidatorBenchmark.deltaBounded                  0.1       5000000  avgt   20  130.499 ±  4.523  ms/op                                                                 
DuplicateDataDetectionValidatorBenchmark.snapshotBounded             0.001       1000000  avgt   20   32.039 ±  0.459  ms/op                                                                 
DuplicateDataDetectionValidatorBenchmark.snapshotBounded             0.001       5000000  avgt   20  322.196 ± 13.048  ms/op                                                                 
DuplicateDataDetectionValidatorBenchmark.snapshotBounded              0.01       1000000  avgt   20   32.298 ±  0.339  ms/op                                                                 
DuplicateDataDetectionValidatorBenchmark.snapshotBounded              0.01       5000000  avgt   20  308.809 ± 12.248  ms/op                                                                 
DuplicateDataDetectionValidatorBenchmark.snapshotBounded               0.1       1000000  avgt   20   32.572 ±  0.738  ms/op                                                                 
DuplicateDataDetectionValidatorBenchmark.snapshotBounded               0.1       5000000  avgt   20  290.514 ± 11.442  ms/op                                                                 
DuplicateDataDetectionValidatorBenchmark.snapshotFullScan            0.001       1000000  avgt   20   29.882 ±  0.185  ms/op                                                                 
DuplicateDataDetectionValidatorBenchmark.snapshotFullScan            0.001       5000000  avgt   20  249.115 ± 20.437  ms/op                                                                 
DuplicateDataDetectionValidatorBenchmark.snapshotFullScan             0.01       1000000  avgt   20   30.235 ±  0.639  ms/op                                                                 
DuplicateDataDetectionValidatorBenchmark.snapshotFullScan             0.01       5000000  avgt   20  294.560 ± 11.769  ms/op                                                                 
DuplicateDataDetectionValidatorBenchmark.snapshotFullScan              0.1       1000000  avgt   20   30.029 ±  0.303  ms/op                                                                 
DuplicateDataDetectionValidatorBenchmark.snapshotFullScan              0.1       5000000  avgt   20  276.480 ± 15.275  ms/op
```


AustinWheel:
There is one implication to call out about doing duplicate detection with deltas which is that delta-detection won't find pre-existing duplicates. The snapshot path (when `previousOrdinals` is empty) does detect pre-existing duplicates, but if you added `DuplicateDataDetectionValidator` after a few cycles have run and duplicates exist, then those duplicates wouldn't necessarily be detected, that is until a snapshot cycle or a producer restart.

hollow-perf/build.gradle
```diff
@@ -6,6 +6,7 @@ apply plugin: 'java'
 
 dependencies {
   implementation project(':hollow')
+  implementation project(':hollow-test')
```
Sunjeet: Is this used?

Sunjeet:
would be good to support both modes for duplicate data detector esp. as the incremental validator rolls out, should allow fallback to the old impl. + the old impl would remain useful in few cases where even snapshots need to validate for duplicates

Sunjeet:
Curious to try another approach- could the duplicate detection happen in the validator impl (the only place its needed) vs. in the pirmary key index? The key thing to figure out with that impl would be how to handle updating the index so that when the validator is invoked the index corresponds to the prior read state engine. But if implemented that way it would be a cleaner separation of concerns. Here's why:                                                                                                                                                                                                                                                                                                             

```                                                                                                                                                                                                                                                                                                                                                                             
  The index's job is to answer "given a key, what ordinal matches?" — it already does this well with getMatch(Object... keys). The delta-aware optimization is really a validation concern, not an indexing concern.                                                                                                                                                           
   
  The validator could do this:                                                                                                                                                                                                                                                                                                                                                 
  1. Get the PopulatedOrdinalListener (or equivalent) to find new ordinals                                                                                                                                                                                                                                                                                                   
  2. For each new ordinal, derive its key using HollowPrimaryKeyValueDeriver
  3. Call the index's existing getMatch() — if the returned ordinal differs from the new ordinal, it's a duplicate
  4. Cap at maxDuplicateKeys and stop early

  This has several advantages:

  - Index stays simple — getDuplicateKeys() (full scan) is the only duplicate detection method on the index, easy to reason about correctness
  - Delta optimization lives where the policy decision is — the validator decides whether to check all records or only new ones, rather than burying that decision inside the index
  - No new state coupling — the index doesn't need to reach into PopulatedOrdinalListener for something that isn't core to indexing
  - Testability — validator logic is tested at the validator level, index tests stay focused on lookup correctness
```

.claude/settings.local.json
```diff
@@ -2,7 +2,10 @@
   "permissions": {
     "allow": [
       "Bash(./gradlew :hollow-protoadapter:test:*)",
-      "Bash(./gradlew:*)"
+      "Bash(./gradlew:*)",
+      "Bash(gh pr:*)",
+      "Bash(GH_HOST=github.com gh api repos/Netflix/hollow/issues/802/comments)",
```
Sunjeet: nit: cleanup

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -151,23 +165,104 @@ public ValidationResult onValidate(ReadState readState) {
...
+
+            Collection<DuplicateKeyInfo> duplicateKeys = previousCycleIndex.getDuplicateKeys(MAX_DISPLAYED_DUPLICATE_KEYS);
+            if (!duplicateKeys.isEmpty()) {
+                String message = String.format(DUPLICATE_KEYS_FOUND_ERRRO_MSG_FORMAT, dataTypeName, fieldPaths,
+                        duplicateKeysToString(duplicateKeys));
+                result = vrb.failed(message);
+            } else {
+                result = vrb.passed();
+            }
+        } else {
```
Sunjeet, line 255: if a producer restores from an older version (e.g. pinned back to fork a delta chain) i think this index might not correspond to the read state. Is that right? Maybe we need additional validation of when it is ok to reuse previousCycleIndex?

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -151,23 +165,104 @@ public ValidationResult onValidate(ReadState readState) {
...
+     * @param laggedIndex an index whose hash table reflects the previous cycle's state
+     * @param newOrdinals ordinals added in the current delta
+     * @param populatedOrdinals all currently populated ordinals
+     * @return duplicate keys found among new ordinals
+     */
+    List<Object[]> findDuplicateKeysInDelta(
+            HollowPrimaryKeyIndex laggedIndex, BitSet newOrdinals, BitSet populatedOrdinals) {
+        // Maps each key to its first ordinal, or -1 if already reported as duplicate
+        Map<List<Object>, Integer> seen = new HashMap<>();
+        List<Object[]> duplicateKeys = new ArrayList<>();
```
Sunjeet, line 306: Due to the additional memory requirement this index may not be suitable in all cases. Maybe we expose a method on the producer builder to opt-in to the incremental validation feature (where it is ok to trade off memory for producer cycle time)

Sunjeet: and in the new mode, allow falling back to the old (slower) validator as a kill switch

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -941,6 +942,31 @@ public B withTypeResharding(boolean allowTypeResharding, boolean forceCoverage)
...
+         * (e.g. an Archaius / Spring property) so operators can disable incremental detection without a
+         * redeploy. Pass {@code () -> false} to opt in without a runtime toggle.
+         * <p>
+         * Default: off — every DDD validator does a full scan per cycle, matching the pre-incremental
+         * behavior.
+         *
+         * @param disabled supplier returning {@code true} to force full-scan; {@code false} to use
+         *                 incremental. Must be non-null and safe to invoke from the producer cycle thread.
+         */
+        public B withIncrementalDuplicateDataDetection(Supplier<Boolean> disabled) {
```
Sunjeet: Since this is specific to duplicate data detection, instead introducing this on the hollow producer builder, it would be better to accept a Supplier<Boolean> on the DDD validator custructor instead

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -151,23 +176,147 @@ public ValidationResult onValidate(ReadState readState) {
...
+    /**
+     * On restore the producer's read state may move to an arbitrary version (e.g. pinned back to
+     * fork a delta chain), so the lagged index built from the prior cycle no longer corresponds
+     * to the read state ordinal space. Drop it; the next cycle will rebuild via the snapshot path.
+     */
+    @Override
+    public void onProducerRestoreStart(long restoreVersion) {
+        if (previousCycleIndex != null) {
+            LOG.info(String.format(
+                    "Duplicate detection for type '%s': dropping lagged index due to producer restore to version %d.",
```
Sunjeet: nit: add "incremental duplication detection" in log msg as an additional clue to why this might be necessary

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -151,23 +176,147 @@ public ValidationResult onValidate(ReadState readState) {
...
+                }
+            } else {
+                seen.put(keyList, ordinal);
+            }
+            ordinal = newOrdinals.nextSetBit(ordinal + 1);
         }
 
-        return vrb.passed();
+        // Check new-vs-existing: matched ordinal must still be populated (not removed in this delta)
+        for (Map.Entry<List<Object>, Integer> entry : seen.entrySet()) {
```
azeng-netflix, line 326: ~~Would this give false positive signal?~~
~~Thinking about the case for primary key `PK`, it's in `v0`. Then in `v1`, record `PK` is **modified**, which would result in ordinal removal and addition. In this case, `PK` would be found both in `seen` and `laggedIndex` which was built from `v0` data -- hence gets marked as dup. Would this happen?~~
NVM read it wrong

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
azeng-netflix: Also, the naming is a bit confusing.
```
        HollowProducer producer = HollowProducer.withPublisher(blobStore)
                .withBlobStager(new HollowInMemoryBlobStager())
                .withIncrementalDuplicateDataDetection(() -> false)
                .withListener(new DuplicateDataDetectionValidator(TypeWithPrimaryKey.class))
```
For the above code, my thought was that it had incremental duplicate data detection turned **OFF**, while in fact, it's **ON**.

azeng-netflix (approved):

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -204,6 +389,15 @@ private String duplicateKeysToString(Collection<DuplicateKeyInfo> duplicateKeys)
                 duplicateKeysString, totalUniqueKeys, totalAffectedRecords, MAX_DISPLAYED_DUPLICATE_KEYS);
     }
 
+    private String deltaDuplicateKeysToString(List<Object[]> duplicateKeys) {
```
azeng-netflix, line 398: nit: can reuse `duplicateKeysToString`?

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -151,23 +219,140 @@ public ValidationResult onValidate(ReadState readState) {
...
+            LOG.info(String.format(
+                    "Incremental duplicate data detection for type '%s': dropping lagged index due to producer restore to version %d.",
+                    dataTypeName, restoreVersion));
+            previousCycleIndex = null;
         }
-        return hollowPrimaryKeyIndex.getDuplicateKeys(MAX_DISPLAYED_DUPLICATE_KEYS);
+    }
+
+    @Override
+    public void onProducerRestoreComplete(Status status, long versionDesired, long versionReached, Duration elapsed) {
```
azeng-netflix, line 357: nit: is this required?

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -151,23 +219,140 @@ public ValidationResult onValidate(ReadState readState) {
...
+                result = vrb.failed(message);
+            } else {
+                result = vrb.passed();
+            }
+        } else {
+            HollowTypeReadState typeState = stateEngine.getTypeState(dataTypeName);
+            PopulatedOrdinalListener listener = typeState.getListener(PopulatedOrdinalListener.class);
+            BitSet populatedOrdinals = listener.getPopulatedOrdinals();
+
+            BitSet newOrdinals = new BitSet();
```
azeng-netflix: nit: can initialize BitSet with sth like max(populatedOrdinals.size(), listener.getPreviousOrdinals().size())

azeng-netflix (approved):

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
AustinWheel: Yeah it's required bc we're implementing the RestoreListener

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
AustinWheel: yeah this is a good callout

PR #803, opened 2026-03-11
Namespace support for collection types

spjegan:
Adds `@HollowCollectionTypeName` and `@HollowMapTypeName` annotations to assign dedicated type states to List/Set element types and Map key/value types. This delivers the same compression benefit as the existing @HollowTypeName on scalar fields — a dedicated pool with fewer ordinals and narrower bit-width per reference — without requiring wrapper POJOs.

Validation at mapper initialization time catches annotation misuse and type name conflicts.

Users of the wrapper-POJO workaround can migrate to the new annotations with no consumer-side schema changes.


adrian-skybaker:
Is there any interesting in merging this? The wrapper POJOs we use cause overhead and degrade introspection based tooling (or at least require workarounds).

hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/HollowObjectTypeMapper.java
```diff
@@ -441,6 +459,8 @@ private MappedField(MappedFieldType specialField) {
             this.typeNameAnnotation = null;
             this.hashKeyAnnotation = null;
             this.numShardsAnnotation = null;
+            this.collectionTypeNameAnnotation = null;
```
Sunjeet: do we need collectionTypeNameAnnotation and mapTypeNameAnnotation in object type mapper?

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -179,6 +183,9 @@ private AbstractHollowProducer(
         if (hashCodeFinder != null) {
             objectMapper.doNotUseDefaultHashKeys();
         }
+        if (collectionTypeNamingEnabled) {
```
Sunjeet: The feature flag enabled on the builder feels redundant given the feature is already opt-in by requiring user to apply it to the data model

spjegan: Agree, this is redundant for new fields. But adding these annotations to an existing field is a schema change and that requires coordinated producer and consumer changes. We could probably document this behavior, but if someone does it without realizing the implications it could be a surprise. The feature flag enables two step migration (add annotations in one PR and enable it in second) that helps with safe deploy of the changes. Happy to drop the feature flag if we are ok with that potential risk in migration use cases.

hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/HollowObjectTypeMapper.java
spjegan: Refactored this to local variables.

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
Sunjeet: +1 to documenting the behavior both on the interface and in user docs. IMO the additional feature flag feels overkill, esp. after proving feature stability we'd want to retire it and that would require additional cleanup

spjegan: Sure, I have already added the documentation. Will update the PR to remove the feature flag.

spjegan: This is done.

Sunjeet (approved):

PR #804, opened 2026-03-11
Revisit building primary key index incrementally with using deltas

AustinWheel:


hollow/src/test/java/com/netflix/hollow/core/index/PrimaryKeyIndexDeltaUpdateReproTest.java
```diff
@@ -0,0 +1,230 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+/**
+ * Reproduces the delta update corruption bug in HollowPrimaryKeyIndex.
+ * <p>
+ * The delta update feature was disabled in 2019 (commit 20ec2864c) due to
+ * "matches unexpectedly returning no result," attributed to a possible C2 JIT issue.
+ * <p>
+ * The actual root cause: {@link HollowPrimaryKeyIndex} registers as a
```
Sunjeet: I dont think that was the root cause. The delta only support is a known usage limtation, documented here https://hollow.how/diving-deeper/?h=listenfordel#keeping-an-index-up-to-date

AustinWheel: yeah I hadn't done enough testing yet, so far I haven't been able to find a root cause for it yet 😪

PR #805, opened 2026-03-18
Skip no-op delta rebuilds for root-local hash indexes

tnabulsi-netflix:
## Summary
- skip `HollowHashIndex` delta reindexing when the index is root-select and all match fields are direct non-reference fields on the root type
- keep the existing rebuild-on-every-delta behavior for traversing or reference-based hash indexes
- add regression tests covering skipped rebuilds, rebuilds on root changes, and unchanged behavior for non-root-local indexes

## Testing
- `./gradlew :hollow:test --tests com.netflix.hollow.core.index.HollowHashIndexTest --rerun-tasks -Pcom.netflix.gradle-predictive-test-selection.enabled=false`

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
@@ -29,6 +29,7 @@ import com.netflix.hollow.core.read.dataaccess.HollowObjectTypeDataAccess;
 import com.netflix.hollow.core.read.engine.HollowReadStateEngine;
 import com.netflix.hollow.core.read.engine.HollowTypeStateListener;
 import com.netflix.hollow.core.read.engine.object.HollowObjectTypeReadState;
+import com.netflix.hollow.core.schema.HollowObjectSchema.FieldType;
 import java.util.Arrays;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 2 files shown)


hollow/src/main/java/com/netflix/hollow/core/index/HollowHashIndex.java
```diff
@@ -117,6 +122,39 @@ private void reindexHashIndex() {
         this.hashStateVolatile = new HollowHashIndexState(builder);
     }
 
+    private boolean canSkipNoOpDeltaUpdate() {
```
azeng-netflix, line 125: This part is not needed, as ordinal change is a direct signal on whether or not a hash needs to rebuilt.
One reason is that, for top level type A, if its member type B has a field change, the change would cause the type A record to have an ordinal change as well. 
This statement holds true, also because:
1. for a type, an ordinal will not be reassigned within a single cycle.
2. Hollow Data set version update via delta would apply every delta on the version chain, and would not skip a version.
3. a sub type's ordinal change would propagate all the way up to the top level type, all the record along the hierarchy would get assigned a new ordinal.

But also wonder if there are other edge cases..

azeng-netflix: @azeng-netflix need to check if double snapshot scenario is also handled and add unit test that would check this invariant in the future.

PR #807, opened 2026-03-28
Improve initialization speed for HollowPrimiaryKeyIndex by allowing parallel hash computation

earthconfed:
###   **Summary**                                                                                                                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
###   Parallel hash computation during reindex                                                                                                                                                                                                                                                                                                                                                                             
   
Parallel hash computation: During reindex(), record hashes can be computed in parallel using Arrays.parallelSetAll for large datasets (configurable via system properties). Hash insertion into the table remains serial to avoid race conditions.                                                                        
                                                            
###   Configurable parallel hash threshold                                                                                                                                                                                                                                                                                                                                                                                 
                                                            
  The threshold that gates parallel hashing defaults to 4096 and is tunable at JVM startup:                                                                                                                                                                                                                                                                                                                            
  -Dcom.netflix.hollow.core.index.HollowPrimaryKeyIndex.parallelHashThreshold=<N>
  This follows the same pattern as the existing allowDeltaUpdate flag in the same class.                                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                                                                       
###   Test coverage                                                                                                                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                                                       
  - testParallelHashMatchesSerialHashForSameLookups — verifies the parallel path (>4096 records) and serial path (<4096 records) both return correct ordinals and that missing keys return -1                                                                                                                                                                                                                          
  - testParallelHashThresholdBoundary — exercises exactly 4095 records (serial) and 4096 records (parallel) to validate correct behaviour at the boundary

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/index/HollowPrimaryKeyIndex.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/index/HollowPrimaryKeyIndex.java
@@ -59,6 +59,7 @@ import java.util.logging.Logger;
 public class HollowPrimaryKeyIndex implements HollowTypeStateListener, TestableUniqueKeyIndex {
     private static final Logger LOG = Logger.getLogger(HollowPrimaryKeyIndex.class.getName());
 
+
     private final HollowObjectTypeReadState typeState;
     private final int[][] fieldPathIndexes;
     private final FieldType[] fieldTypes;
@@ -460,6 +461,9 @@ public class HollowPrimaryKeyIndex implements HollowTypeStateListener, TestableU
     private static final boolean ALLOW_DELTA_UPDATE =
             Boolean.getBoolean("com.netflix.hollow.core.index.HollowPrimaryKeyIndex.allowDeltaUpdate");
 
+    private static final int PARALLEL_HASH_THRESHOLD =
+            Integer.getInteger("com.netflix.hollow.core.index.HollowPrimaryKeyIndex.parallelHashThreshold", 4096);
+
     @Override
     public synchronized void endUpdate() {
         if (hashTableVolatile == null) {
@@ -488,7 +492,6 @@ public class HollowPrimaryKeyIndex implements HollowTypeStateListener, TestableU
                 well fix that.  Attempts to reproduce this locally has so far failed.
                 Given the importance of indexing a full reindex is performed on such a failure.  This, however,
                 will make it more difficult to detect such issues.
-
                 This approach does not protect against the case where the index is corrupt and not yet
                 detected, until a further update.  In such cases it may be possible for clients, in the interim
                 of a forced reindex, to operate on a corrupt index: queries may incorrectly return no match.  As such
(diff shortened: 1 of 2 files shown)


hollow/src/main/java/com/netflix/hollow/core/index/HollowPrimaryKeyIndex.java
```diff
@@ -460,6 +461,9 @@ public void removedOrdinal(int ordinal) { }
     private static final boolean ALLOW_DELTA_UPDATE =
             Boolean.getBoolean("com.netflix.hollow.core.index.HollowPrimaryKeyIndex.allowDeltaUpdate");
 
+    private static final int PARALLEL_HASH_THRESHOLD =
+            Integer.getInteger("com.netflix.hollow.core.index.HollowPrimaryKeyIndex.parallelHashThreshold", 4096);
```
Sunjeet: can we avoid properties and instead make this a param on the index constructor? I dont think we support any properties in hollow- but being able to specify behavior on builders/constructors is great

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrimaryKeyIndex.java
```diff
@@ -528,24 +531,45 @@ private synchronized void reindex() {
...
 
         FixedLengthElementArray hashedArray = new FixedLengthElementArray(memoryRecycler, (long)hashTableSize * (long)bitsPerElement);
 
         int hashMask = hashTableSize - 1;
 
-        int ordinal = ordinals.nextSetBit(0);
-        while(ordinal != ORDINAL_NONE) {
-            int hashCode = recordHash(ordinal);
-            int bucket = hashCode & hashMask;
+        if (shouldComputeHashesInParallel(cardinality)) {
```
Sunjeet, line 558: if the goal is to make index computation faster, ICYMI there is support for updating primary key indexes incrementally with delta updates. Its not exercised in prod, but something to try

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrimaryKeyIndex.java
earthconfed: I've added a param in the constructor.

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrimaryKeyIndex.java
earthconfed: I will look into that as well, but it shouldn't stop this improve from going in.

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrimaryKeyIndex.java
```diff
@@ -651,6 +691,10 @@ private boolean bucketInRange(int fromBucket, int toBucket, int testBucket) {
         }
     }
 
+    private boolean shouldComputeHashesInParallel(int cardinality) {
+        return cardinality >= parallelHashThreshold;
```
Sunjeet, line 695: is it intentional to have this new feature enabled on all existing consumers? I dont think we should do that

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrimaryKeyIndex.java
```diff
@@ -528,24 +547,45 @@ private synchronized void reindex() {
...
-            int hashCode = recordHash(ordinal);
-            int bucket = hashCode & hashMask;
+        if (shouldComputeHashesInParallel(cardinality)) {
+            int[] ordinalArray = new int[cardinality];
+            int pos = 0;
+            for (int o = ordinals.nextSetBit(0); o >= 0; o = ordinals.nextSetBit(o + 1)) {
+                ordinalArray[pos++] = o;
+            }
+            int[] hashCodes = new int[cardinality];
+            Arrays.parallelSetAll(hashCodes, i -> recordHash(ordinalArray[i]));
```
Sunjeet, line 565: Arrays.parallelSetAll will use a JVM wide ForkJoinPool.commonPool() - we should use a named, owned thread pool instead

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrimaryKeyIndex.java
Sunjeet: sgtm.. parallel computation for snapshot and incremental computation for deltas

PR #808, opened 2026-04-01
fix: flat record node boxed boolean accessor properly returns nulls

eduardoramirez:


hollow/src/main/java/com/netflix/hollow/core/write/objectmapper/flatrecords/traversal/FlatRecordTraversalObjectNode.java
```diff
@@ -113,7 +113,7 @@ public boolean getFieldValueBoolean(String field) {
   }
 
   public Boolean getFieldValueBooleanBoxed(String field) {
-    return getFieldValueBoolean(field);
```
eduardoramirez, line 116: this method collapses nulls to false

```java
public boolean getFieldValueBoolean(String field) {
  Boolean b = reader.readFieldBoolean(ordinal, field);
  return Boolean.TRUE.equals(b);
}
```

PR #810, opened 2026-04-11
Resolve missing type dependencies from snapshot during producer restore

abersnaze:
## Summary

Fixes a failure during `HollowProducer.restore()` when the snapshot contains types not registered via `initializeDataModel()`. The restore would throw `IllegalStateException: Missing schema(s) in dependency ordered list`.

This happens when using `HollowMessageMapper` (proto adapter) which lazily creates schemas for types like `Struct`, `Value`, and `ListValue` during `runCycle`. On restart, the producer only has the base types registered, but the snapshot contains the lazy types too.

### Fix

During restore, resolve missing type dependencies by adopting types from the snapshot that are **transitively referenced** by the producer's registered schemas but not in the producer's model. Types intentionally removed by the producer are not re-added — only dependency-resolved types are adopted.

### Why not adopt all snapshot types?

Existing behavior allows producers to intentionally remove types between versions (e.g., `initializeDataModel(TypeA, TypeE)` dropping `TypeB`). Blindly merging all snapshot types would break this. The transitive-dependency approach preserves intentional removals while resolving missing dependencies.

## Test plan

- [x] `testRestoreWithAdditionalTypesInSnapshot` — dynamic types adopted from snapshot
- [x] `testRestoreWithExtraFieldsInSnapshotSchema` — producer schema wins on field differences
- [x] `testRestoreWithNewFieldInProducerSchema` — new fields work on restore
- [x] `testRestoreWithDynamicListAndNestedTypes` — transitive adoption of list + nested types
- [x] Existing `canRemoveAndModifyNewTypesFromRestoredState` — intentional type removal still works
- [x] Full hollow test suite passes

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
@@ -41,10 +41,8 @@ import com.netflix.hollow.core.read.HollowBlobInput;
 import com.netflix.hollow.core.read.engine.HollowBlobHeaderReader;
 import com.netflix.hollow.core.read.engine.HollowBlobReader;
 import com.netflix.hollow.core.read.engine.HollowReadStateEngine;
-import com.netflix.hollow.core.schema.HollowCollectionSchema;
-import com.netflix.hollow.core.schema.HollowMapSchema;
-import com.netflix.hollow.core.schema.HollowObjectSchema;
 import com.netflix.hollow.core.schema.HollowSchema;
+import com.netflix.hollow.core.schema.HollowSchemaUtil;
 import com.netflix.hollow.core.schema.HollowSchemaHash;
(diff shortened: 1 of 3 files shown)


Sunjeet (approved):

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -353,6 +366,60 @@ private HollowProducer.ReadState restore(
...
+                        result.put(refType, snapshotByName.get(refType));
+                        changed = true;
+                    }
+                }
+            }
+        }
+        return result.values();
+    }
+
+    private static List<String> getReferencedTypeNames(HollowSchema schema) {
```
Sunjeet: nit: this could live in `HollowSchemaUtil`

PR #811, opened 2026-04-14
Fix Struct/Value fields not written to Hollow, breaking delta detection

abersnaze:
## Summary

Two bugs caused `google.protobuf.Struct` field changes to be invisible to Hollow's delta detection. The producer would report "no delta in output state" on every cycle even when Struct data changed, breaking the Cinder delta chain.

### Bug 1: Value schema field name mismatch

The hand-coded Value schema in `ensureValueSchemaExists()` used camelCase names (`stringValue`, `numberValue`, etc.), but proto's `FieldDescriptor.getName()` returns snake_case (`string_value`, `number_value`). The field lookup in `addObject()` found no match, so all Value fields were silently skipped. Every Value serialized as empty bytes, producing identical ordinals.

**Fix:** Use snake_case names to match the convention used by `createSchemas()` for all other proto types. Also fix `null_value` type from BOOLEAN to STRING (NullValue is an enum).

### Bug 2: parseCollection() didn't handle HollowMapSchema

`parseCollection()` only handled `HollowCollectionSchema` (lists/sets). The Struct's `fields` map (`MapOfStringToValue`) is a `HollowMapSchema` which fell through, writing an empty map record every time.

**Fix:** Add `HollowMapSchema` handling that processes proto map entries. Also use `computeIfAbsent` for lazy schema loading since Struct/Value schemas are created dynamically.

## Reproduction

Found while building a Cinder producer that publishes DGW namespace proto messages containing `google.protobuf.Struct` config fields. Namespace config changes (e.g., adding a nonce value) were never detected by Hollow, so no deltas were published.

## Test plan

- [x] `testDeltaDetectionWithStructFieldChange` — Struct value change detected across cycles
- [x] `testDeltaDetectionAfterRestoreWithStructFieldChange` — Struct change detected after restore
- [x] Updated `testGoogleStructAndValueAsReferences` assertions for snake_case field names
- [x] All existing protoadapter tests pass

### DIFF
--- a/hollow-protoadapter/src/main/java/com/netflix/hollow/protoadapter/HollowProtoAdapter.java
+++ b/hollow-protoadapter/src/main/java/com/netflix/hollow/protoadapter/HollowProtoAdapter.java
@@ -17,6 +17,7 @@
 package com.netflix.hollow.protoadapter;
 
 import com.google.protobuf.ByteString;
+import com.google.protobuf.Descriptors;
 import com.google.protobuf.Descriptors.FieldDescriptor;
 import com.google.protobuf.Message;
 import com.netflix.hollow.core.schema.HollowCollectionSchema;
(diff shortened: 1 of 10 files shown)


hollow-protoadapter/src/main/java/com/netflix/hollow/protoadapter/HollowProtoAdapter.java
```diff
@@ -441,11 +441,52 @@ private int wrapPrimitiveValue(FieldDescriptor.JavaType javaType, Object value,
...
+            HollowMapWriteRecord mapRec = (HollowMapWriteRecord) collectionRec;
+            String keyType = mapSchema.getKeyType();
+            String valueType = mapSchema.getValueType();
+
+            for (Object value : values) {
+                if (value instanceof Message) {
+                    Message entry = (Message) value;
+                    com.google.protobuf.Descriptors.Descriptor entryDesc = entry.getDescriptorForType();
+                    com.google.protobuf.Descriptors.FieldDescriptor keyField = entryDesc.findFieldByName("key");
+                    com.google.protobuf.Descriptors.FieldDescriptor valueField = entryDesc.findFieldByName("value");
```
spjegan: nit: can these be moved to imports?

hollow-protoadapter/src/test/java/com/netflix/hollow/protoadapter/HollowProtoAdapterTest.java
```diff
@@ -1039,5 +1039,182 @@ public void testMapWithValueTypeValues() throws Exception {
...
+    @Test
+    public void testDeltaDetectionWithStructFieldChange() throws Exception {
+        Class<?> personClass = protoClassLoader.loadClass("com.netflix.hollow.test.proto.PersonProtos$Person");
+        Class<?> structClass = protoClassLoader.loadClass("com.google.protobuf.Struct");
+        Class<?> valueClass = protoClassLoader.loadClass("com.google.protobuf.Value");
+
+        Method personNewBuilder = personClass.getMethod("newBuilder");
+        Method structNewBuilder = structClass.getMethod("newBuilder");
+        Method valueNewBuilder = valueClass.getMethod("newBuilder");
+        Method structBuilderPutFields = structClass.getDeclaredClasses()[0].getMethod("putFields", String.class, valueClass);
```
spjegan, line 1063: Does it make sense to include other types as well, along with `String`?

spjegan (approved):
LGTM

PR #812, opened 2026-04-16
Validations on number of blob headers and empty snapshot in integrity check

spjegan:
Added couple of validations
- Throw if the number of headers exceed 32,767. More headers than this limit results in data corruption.
- Integrity check throws if the read snapshot is empty. This is to fail-fast and avoid publishing empty snapshots when there is data corruption.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
@@ -870,6 +870,18 @@ abstract class AbstractHollowProducer {
             HollowReadStateEngine pending = readStates.pending().getStateEngine();
             readSnapshot(artifacts.snapshot, pending);
 
+            long writtenOrdinals = getWriteEngine().getOrderedTypeStates().stream()
+                    .mapToLong(ts -> ts.getPopulatedBitSet().cardinality())
+                    .sum();
+            long pendingOrdinals = pending.getTypeStates().stream()
+                    .mapToLong(ts -> ts.getPopulatedOrdinals().cardinality())
+                    .sum();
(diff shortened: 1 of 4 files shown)


hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -870,6 +870,18 @@ private ReadStateHelper checkIntegrity(
...
+                    .mapToLong(ts -> ts.getPopulatedBitSet().cardinality())
+                    .sum();
+            long pendingOrdinals = pending.getTypeStates().stream()
+                    .mapToLong(ts -> ts.getPopulatedOrdinals().cardinality())
+                    .sum();
+            if (writtenOrdinals > 0 && pendingOrdinals == 0) {
+                throw new IllegalStateException(
+                        "Integrity check failed: snapshot produced zero records (wrote " + writtenOrdinals + " records)."
+                                + " This indicates possible blob corruption.");
+            }
```
eduardoramirez, line 883: should this be under the `if (readStates.hasCurrent())` block?

eduardoramirez (approved):

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
spjegan: Moving it under `if (readStates.hasCurrent())` would prevent this check throwing on corruption issues in first cycle. I prefer to leave it outside, let me know if you think otherwise.

eduardoramirez: your reasoning makes sense to me. no ask from me

PR #814, opened 2026-04-21
Sort blob information header entries alphabetically in diff UI

bihaoxwork:
### Summary
The **Blob Information** table on the Hollow Diff UI renders header tags in a non-deterministic order. `DiffPage.getHeaderEntries()` merges the FROM- and TO-side header tag keys into a `HashSet<String>` before building the display list, so the row order depends on string hash distribution and varies between runs and JDK versions.

### Fix
Replace the `HashSet<String>` in `DiffPage.getHeaderEntries()` with a `TreeSet<String>`, which iterates keys in natural alphabetical order. The Velocity template `diff-footer.vm` already renders rows in the list order it receives, so the fix is localized to the Java side. The numeric `#` column is assigned during iteration and is renumbered to match the new sorted order. No changes to the data model (`HollowHeaderEntry`) or any public API — this is purely a rendering order change.

### Test plan
- Added `HollowDiffUITest.blobInformationTableIsSortedAlphabetically` (mirrors the style of `HollowHistoryUITest`): boots a `HollowDiffUIServer` backed by `FakeHollowDiffGenerator` and asserts the four seeded header keys (`fromTag`, `tag1`, `tag2`, `toTag`) appear in alphabetical order in the rendered HTML.
- Verified locally against real Cinder-managed namespace (`WorkdayProject.v4`, ` ad_creatives`, `WorkdaySupplier.v7`) via `CinderDevUtils.diff(...)` — the Blob Information table now renders in alphabetical order, with `hollow.*` entries preceding `hollowservice.*` entries.
- Existing `hollow-diff-ui` test suite passes.

### DIFF
--- a/hollow-diff-ui/src/main/java/com/netflix/hollow/diff/ui/pages/DiffPage.java
+++ b/hollow-diff-ui/src/main/java/com/netflix/hollow/diff/ui/pages/DiffPage.java
@@ -25,10 +25,10 @@ import com.netflix.hollow.tools.diff.HollowTypeDiff;
 import com.netflix.hollow.ui.HollowUISession;
 import java.io.Writer;
 import java.util.ArrayList;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeSet;
 import javax.servlet.http.Cookie;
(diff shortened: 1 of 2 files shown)


azeng-netflix (approved):
Looks Good

PR #821, opened 2026-04-29
Allow heterogeneous Value types for the same Struct field name

abersnaze:
## Summary

\`google.protobuf.Struct\` is explicitly designed to allow the same field name to hold different \`Value\` oneof cases across different Struct instances — that is the whole point of Struct as a flexible JSON-like container. The previous code rejected this with \`IllegalStateException(\"Conflicting types for Struct field...\")\`, which broke any namespace dataset where the same \`persistence_config\` key legitimately holds different scalar types across rows (e.g. \`data_ttl_secs\` as a number in one namespace and a string in another).

This was found in production: dgw-control's KV cinder producer was failing every cycle with:
\`\`\`
java.lang.IllegalStateException: Conflicting types for Struct field 'data_ttl_secs' in NamespaceInternal.config: previously number_value, now string_value
  at com.netflix.hollow.protoadapter.HollowMessageMapper.validateStructFields(HollowMessageMapper.java:475)
\`\`\`

The Hollow \`Value\` object schema already declares one nullable column per oneof case (\`number_value\`, \`string_value\`, \`bool_value\`, \`struct_value\`, \`list_value\`, \`null_value\`), so any case round-trips through the same Hollow type. \`validateStructFields\` was preventing legitimate writes that the read side already handles correctly.

## Changes

- Remove \`validateStructFields\` and the \`structFieldTypes\` tracking map.
- Replace the existing \`testStructFieldTypeConflictDetection\` (which asserted the bad behaviour) with \`testStructFieldHeterogeneousTypesRoundTrip\`, which writes three Person records whose \`Struct.metadata.age\` is a number, a string, and a bool respectively, and asserts every record reads back with only the expected oneof column populated and full \`readHollowRecord\` round-trip preserves the exact Value case.

## Test plan

- [x] \`./gradlew :hollow-protoadapter:test\` — all tests pass (including the new round-trip)

### DIFF
--- a/hollow-protoadapter/src/main/java/com/netflix/hollow/protoadapter/HollowMessageMapper.java
+++ b/hollow-protoadapter/src/main/java/com/netflix/hollow/protoadapter/HollowMessageMapper.java
@@ -74,11 +74,6 @@ public class HollowMessageMapper {
     private final Map<String, HollowProtoAdapter> adapters = new ConcurrentHashMap<String, HollowProtoAdapter>();
     private boolean ignoreListOrdering = false;
 
-    // Track field types for Struct fields to validate consistency across instances
-    // Maps "ParentType.fieldPath.structFieldName" -> Value type (e.g., "Person.metadata.age" -> "numberValue")
-    // Thread-safe: uses ConcurrentHashMap with atomic putIfAbsent for conflict detection
-    private final ConcurrentHashMap<String, String> structFieldTypes = new ConcurrentHashMap<String, String>();
-
     // Lock objects for schema creation synchronization
     private final Object valueSchemaLock = new Object();
     private final Object structSchemaLock = new Object();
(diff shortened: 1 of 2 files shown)


abersnaze:
Closing while I verify the fix correctness more thoroughly

abersnaze:
Verified the fix conclusively before reopening:

1. **Test reliably reproduces the production failure** - I temporarily reverted just the production-code change (kept the new test) and ran the test against the old `validateStructFields` code. It failed with the same `IllegalStateException` we hit in production.

2. **Fix makes the test pass** - Re-applying the fix makes the new round-trip test pass and all 19 existing tests still pass (`./gradlew :hollow-protoadapter:test` → 19 tests, 0 failures, 0 errors).

3. **Fix is correct by design** - The `Value` Hollow schema (created in `ensureValueSchemaExists`) declares all 6 oneof case columns. Each record sets exactly one. `readHollowValue` reads back by checking `!obj.isNull(...)` for each case. Heterogeneous values across records have always been supported by the schema; `validateStructFields` was rejecting writes the read side already handled correctly.

The new test reproduces the production scenario verbatim: 3 Person records with the same `Struct` field name (`age`) holding 3 different `Value` oneof cases (number, string, bool). All three round-trip with the exact oneof case preserved.

PR #828, opened 2026-05-29
[Feat] Add JSON endpoint to HollowDiffUI for serving precomputed diff data.

xguo-stack:
### Description 
Provide function to generate diff report in JSON format. 

### Changes Made 

1. New `com.netflix.hollow.tools.diff.report package` to store the JSON object before serialization
1.1 `HollowDiffReport.java`: Json Object Class 
1.2 `HollowDiffReportBuilder.java`: helper class to transform diff into the Json Object
1.3 `HollowDiffReportMetadata.java`: Metadata regarding the diff
1.4 `HollowDiffReportOptions.java`: Report config to control how many rows of detail to include in the JSON response. 

2. `main/java/com/netflix/hollow/diff/ui`: Diff→serialized JSON logic 
2.1 Update `HollowDiffUI` constructor  to take in metadata and options parameters. 
2.2 Update `HollowDiffUIRouter` : `addDiff` function would take in the above parameters 
2.3 Update `DiffUIServer`:  `addDiff` function would expect the above parameters 
2.4 Update `DiffUIWebServer`: `addDiff` function would take in the above parameters for its router
2.5 Update `HollowDiffUIServer`: overload a new `addDiff` function to  take in the above parameters.

### DIFF
--- a/hollow-diff-ui/src/main/java/com/netflix/hollow/diff/ui/DiffUIServer.java
+++ b/hollow-diff-ui/src/main/java/com/netflix/hollow/diff/ui/DiffUIServer.java
@@ -18,8 +18,17 @@
 package com.netflix.hollow.diff.ui;
 
 import com.netflix.hollow.tools.diff.HollowDiff;
+import com.netflix.hollow.tools.diff.report.HollowDiffReportMetadata;
+import com.netflix.hollow.tools.diff.report.HollowDiffReportOptions;
 import com.netflix.hollow.ui.UIServer;
 
 public interface DiffUIServer extends UIServer {
-       HollowDiffUI addDiff(String diffPath, HollowDiff diff, String fromBlobName, String toBlobName);
+
+    HollowDiffUI addDiff(
+            String diffPath,
+            HollowDiff diff,
+            String fromBlobName,
+            String toBlobName,
+            HollowDiffReportMetadata metadata,
+            HollowDiffReportOptions options);
 }
--- a/hollow-diff-ui/src/main/java/com/netflix/hollow/diff/ui/DiffUIWebServer.java
+++ b/hollow-diff-ui/src/main/java/com/netflix/hollow/diff/ui/DiffUIWebServer.java
@@ -1,6 +1,8 @@
 package com.netflix.hollow.diff.ui;
 
 import com.netflix.hollow.tools.diff.HollowDiff;
+import com.netflix.hollow.tools.diff.report.HollowDiffReportMetadata;
+import com.netflix.hollow.tools.diff.report.HollowDiffReportOptions;
 import com.netflix.hollow.ui.HollowUIWebServer;
 import com.netflix.hollow.ui.HttpHandlerWithServletSupport;
 
@@ -12,7 +14,14 @@ class DiffUIWebServer extends HollowUIWebServer implements DiffUIServer {
         this.router = router;
     }
 
-    public HollowDiffUI addDiff(String diffPath, HollowDiff diff, String fromBlobName, String toBlobName) {
-        return this.router.addDiff(diffPath, diff, fromBlobName, toBlobName);
+    @Override
+    public HollowDiffUI addDiff(
+            String diffPath,
+            HollowDiff diff,
+            String fromBlobName,
+            String toBlobName,
+            HollowDiffReportMetadata metadata,
+            HollowDiffReportOptions options) {
+        return router.addDiff(diffPath, diff, fromBlobName, toBlobName, metadata, options);
     }
 }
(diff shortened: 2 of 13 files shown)


hollow-diff-ui/src/main/java/com/netflix/hollow/diff/ui/DiffUIServer.java
```diff
@@ -18,8 +18,17 @@
...
 
 import com.netflix.hollow.tools.diff.HollowDiff;
+import com.netflix.hollow.tools.diff.report.HollowDiffReportMetadata;
+import com.netflix.hollow.tools.diff.report.HollowDiffReportOptions;
 import com.netflix.hollow.ui.UIServer;
 
 public interface DiffUIServer extends UIServer {
-       HollowDiffUI addDiff(String diffPath, HollowDiff diff, String fromBlobName, String toBlobName);
+
+    HollowDiffUI addDiff(
```
azeng-netflix, line 27: Should be able to overload the method instead of changing the existing interface, to avoid breaking backward compatibility.

hollow-diff-ui/src/main/java/com/netflix/hollow/diff/ui/DiffUIWebServer.java
```diff
@@ -12,7 +14,14 @@ public DiffUIWebServer(HollowDiffUIRouter router, int port) {
         this.router = router;
     }
 
-    public HollowDiffUI addDiff(String diffPath, HollowDiff diff, String fromBlobName, String toBlobName) {
-        return this.router.addDiff(diffPath, diff, fromBlobName, toBlobName);
+    @Override
+    public HollowDiffUI addDiff(
```
azeng-netflix, line 18: same for backward compatibility.

hollow-diff-ui/src/test/java/com/netflix/hollow/diffview/HollowDiffUIServerTest.java
```diff
@@ -1,5 +1,6 @@
 package com.netflix.hollow.diffview;
 
+import com.netflix.hollow.diff.ui.HollowDiffUI;
```
azeng-netflix, line 3: unused import.

hollow-diff-ui/src/main/java/com/netflix/hollow/diff/ui/HollowDiffUI.java
```diff
@@ -85,6 +98,10 @@ public boolean serveRequest(String pageName, HttpServletRequest req, HttpServlet
         } else if("collapsediffrow".equals(pageName)) {
             diffViewOutputGenerator.collapseRow(req, resp);
             return true;
+        } else if("json".equals(pageName)) {
```
azeng-netflix, line 101: need an appropriate naming (like `diffreport`)

hollow/src/main/java/com/netflix/hollow/tools/diff/report/HollowDiffReportMetadata.java
```diff
@@ -0,0 +1,45 @@
+package com.netflix.hollow.tools.diff.report;
+
+/**
+ * Caller-supplied labels for the two sides of a {@link com.netflix.hollow.tools.diff.HollowDiff}.
+ */
+public final class HollowDiffReportMetadata {
```
azeng-netflix, line 6: `namespace` is a rather internal concept which we'd like to avoid leaking to the OSS project.
Maybe just make it more generic which behaves liek a  `Map<String, Pair<String>>`,  where the map contains key value pair --`{metadata_key: <from_metadata_value, to_metadata_value>}`

Also, I think `version` would always exist in the header blob (this is done by Hollow Producer), so maybe no need to add it here.

hollow/src/main/java/com/netflix/hollow/tools/diff/report/HollowDiffReportOptions.java
```diff
@@ -0,0 +1,40 @@
+package com.netflix.hollow.tools.diff.report;
+
+/**
+ * Limits applied when exporting a {@link com.netflix.hollow.tools.diff.HollowDiff} to JSON.
+ */
+public final class HollowDiffReportOptions {
+
+    public static final int DEFAULT_MAX_UNMATCHED_SAMPLE = 200;
```
azeng-netflix, line 8: For this PR I think we can remove unmatched samples to keep the result simple.

hollow/src/main/java/com/netflix/hollow/tools/diff/report/HollowDiffReport.java
```diff
@@ -0,0 +1,257 @@
...
+public final class HollowDiffReport {
+
+    private String fromNamespace;
+    private String toNamespace;
+    private Long fromVersion;
+    private Long toVersion;
+    private List<BlobHeader> blobHeaders;
+    private List<TypeDiff> typeDiffs;
+
+    public String getFromNamespace() {
```
azeng-netflix, line 17: Related to this [comment](https://github.com/Netflix/hollow/pull/828/changes#r3353870511)

PR #832, opened 2026-06-23
Fix DDD Validator to update lagged index during a failed cycle

AustinWheel:
Before the DDD validator would update the lagged index based on it succeeding or failing for the current cycle. This makes the assumption that if this DDD validator succeeds, then the cycle succeeded. In reality when you have multiple validators another can fail and cause the cycle to fail. In that case this index could have been incorrectly updated as if the cycle would succeed.

To address this the PR introduces 

- CycleListener and handles index updates based on the cycle succeeding, not just this validator.
- A check `dropLaggedIndexIfEngineChanged` in case the read state engine changes

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
@@ -200,7 +200,6 @@ public class DuplicateDataDetectionValidator implements ValidatorListener, Resto
         return NAME + "_" + dataTypeName;
     }
 
-    // Drop the retained lagged index (and its engine binding) so the next cycle rebuilds a baseline.
     private void dropLaggedIndex() {
         previousCycleIndex = null;
         indexStateEngine = null;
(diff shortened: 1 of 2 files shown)


hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -358,6 +369,43 @@ public void onProducerRestoreComplete(Status status, long versionDesired, long v
...
+                                    + "to version %d; dropping it so the next cycle rebuilds from a snapshot baseline.",
+                            dataTypeName, version), e);
+                    dropLaggedIndex();
+                }
+            }
+        } else if (cycleAction == CycleAction.BUILT_BASELINE) {
+            // A baseline built this cycle was rolled back; drop it so the next cycle rebuilds.
+            dropLaggedIndex();
+        }
+        cycleAction = CycleAction.NONE;
```
azeng-netflix: nit: would prefer to remove it, as the producer's contract is that `onCycleStart` is invoked at first.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -358,6 +369,43 @@ public void onProducerRestoreComplete(Status status, long versionDesired, long v
...
+                            "Incremental duplicate data detection for type '%s': failed to advance lagged index "
+                                    + "to version %d; dropping it so the next cycle rebuilds from a snapshot baseline.",
+                            dataTypeName, version), e);
+                    dropLaggedIndex();
+                }
+            }
+        } else if (cycleAction == CycleAction.BUILT_BASELINE) {
+            // A baseline built this cycle was rolled back; drop it so the next cycle rebuilds.
+            dropLaggedIndex();
+        }
```
azeng-netflix, line 395: A note for other reviewers, no change reuqired -- no need to drop the index even if `(status.getType() =!= Status.StatusType.SUCCESS && cycleAction != CycleAction.BUILT_BASELINE)`.
Cause, in that case, producer itself would rollback the delta to the previous version `v`, and the `previousCycleIndex` would still be properly indexed for the version `v`.

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -192,23 +200,28 @@ public String getName() {
...
+        previousCycleIndex = null;
+        indexStateEngine = null;
+    }
+
     @Override
     public ValidationResult onValidate(ReadState readState) {
         ValidationResult.ValidationResultBuilder vrb = ValidationResult.from(this);
         vrb.detail(DATA_TYPE_NAME, dataTypeName);
 
+        HollowReadStateEngine stateEngine = readState.getStateEngine();
```
azeng-netflix, line 213: Also as a note, no changes required, without restore, the `HollowReadStateEngine` should be the same across cycles.

azeng-netflix (approved):
Looks good. Initially failed to realize the additional complexity that gets added when a validator retains states across cycles.
Think that's something ~~worth adding to our context package as well~~ (removed as it might be too generic to be added, but this would be a good reference PR for future validator change).

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
AustinWheel: Ah my bad, I thought about adding a comment to explain that but decided against it, but it prob would've been more clear

PR #833, opened 2026-06-24
Chunk generated API constructor to avoid 64KB method limit for large data models

JiazhengSun:
## Problem

The generated `HollowAPI` implementation inlines per-type initialization (type API construction, `addTypeAPI`, factory lookup, and provider setup) directly into a single constructor in `HollowAPIClassJavaGenerator`. Past a few hundred types this constructor's bytecode exceeds the JVM 64KB per-method limit and the generated source fails to compile with **"code too large"** (observed around ~460 types). Today the only workaround is to shard a data model into smaller sub-APIs, which doesn't scale as a model grows.

## Change

When a data model has more than `CONSTRUCTOR_CHUNKING_THRESHOLD` (250) types, the per-type initialization is split across `init_N(...)` helper methods (and `detachCaches()` is likewise chunked), each handling a bounded batch of types. The per-type fields are non-final in this mode, since they are assigned from helper methods.

**At or below the threshold, the generated source is byte-for-byte identical to before** — the existing code path is unchanged; the per-type/sampler/detach snippets were extracted into helpers that the original path calls verbatim. So only models that previously failed to compile change shape.

## Tests

`HollowLargeDataModelAPIGeneratorTest`:
- `largeDataModelCompilesViaChunkedConstructor` — generates **and compiles** a 600-type model (previously failed with "code too large").
- `smallDataModelKeepsInlinedConstructor` — asserts a small model keeps the inlined constructor and `final` fields (no-regression guard).

The full `com.netflix.hollow.api.codegen.*` suite passes, confirming sub-threshold output is unchanged. A FindBugs filter entry excludes the experimental `SKIPPED_CLASS_TOO_BIG` info report that fires when analyzing the intentionally-large generated test class.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/codegen/HollowAPIClassJavaGenerator.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/codegen/HollowAPIClassJavaGenerator.java
@@ -46,6 +46,7 @@ import com.netflix.hollow.core.read.dataaccess.missing.HollowSetMissingDataAcces
 import com.netflix.hollow.core.schema.HollowSchema;
 import com.netflix.hollow.core.schema.HollowSchemaSorter;
 import com.netflix.hollow.core.util.AllHollowRecordCollection;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


JiazhengSun:
Prototyped with Claude for problem illustration purpose. Needs verification from Hollow contributor.

Sunjeet:
re: Past a few hundred types...

Is this something that we expect to incur in practice (a few hundred types in the data model)- is it something that needs to be supported, or should the data model be reconsidered?

PR #834, opened 2026-06-25
Adds allowDeltaUpdate supplier to HollowPrimaryKeyIndex

AustinWheel:
Adds a constructor such that a supplier can be provided for `allowDeltaUpdate`. This way a `HollowPrimaryKeyIndex` can be toggled to use delta updates at runtime. The change remains backwards compatible for any uses settings `ALLOW_DELTA_UPDATE`

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
@@ -251,8 +251,7 @@ public class DuplicateDataDetectionValidator implements ValidatorListener, Resto
 
     private ValidationResult validateSnapshotBaseline(HollowReadStateEngine stateEngine, PrimaryKey primaryKey,
                                                       String fieldPaths, ValidationResult.ValidationResultBuilder vrb) {
-        previousCycleIndex = new HollowPrimaryKeyIndex(stateEngine, primaryKey,
-                () -> Boolean.TRUE.equals(incrementalEnabled.get()));
+        previousCycleIndex = new HollowPrimaryKeyIndex(stateEngine, primaryKey, incrementalEnabled);
         indexStateEngine = stateEngine;
         cycleAction = CycleAction.BUILT_BASELINE;
         LOG.log(Level.FINE, String.format("Duplicate detection for type '%s': incremental mode, snapshot baseline.", dataTypeName));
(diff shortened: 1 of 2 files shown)


hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
```diff
@@ -251,7 +251,8 @@ private ValidationResult validateFullScan(HollowReadStateEngine stateEngine, Pri
 
     private ValidationResult validateSnapshotBaseline(HollowReadStateEngine stateEngine, PrimaryKey primaryKey,
                                                       String fieldPaths, ValidationResult.ValidationResultBuilder vrb) {
-        previousCycleIndex = new HollowPrimaryKeyIndex(stateEngine, primaryKey);
+        previousCycleIndex = new HollowPrimaryKeyIndex(stateEngine, primaryKey,
+                () -> Boolean.TRUE.equals(incrementalEnabled.get()));
```
workeatsleep: should this just be "incrementalEnabled" instead of "() -> Boolean.TRUE.equals(incrementalEnabled.get())"?

hollow/src/main/java/com/netflix/hollow/core/index/HollowPrimaryKeyIndex.java
```diff
@@ -459,6 +472,7 @@ public void removedOrdinal(int ordinal) { }
 
     private static final boolean ALLOW_DELTA_UPDATE =
             Boolean.getBoolean("com.netflix.hollow.core.index.HollowPrimaryKeyIndex.allowDeltaUpdate");
+    private static final BooleanSupplier DEFAULT_ALLOW_DELTA_UPDATE = () -> ALLOW_DELTA_UPDATE;
```
workeatsleep: nit: rename to SYSTEM_ALLOW_DELTA_UPDATE, the name DEFAULT sounds like a fixed value

workeatsleep (approved):

hollow/src/main/java/com/netflix/hollow/api/producer/validation/DuplicateDataDetectionValidator.java
AustinWheel: That would be cleaner but I'm using a `BooleanSupplier` in the HollowPrimaryKeyIndex (`incrementalEnabled` is a `Supplier<Boolean>`). I feel like we should have gone with this before in the [DuplicateDataDetectionValidator PR](https://github.com/Netflix/hollow/pull/802#discussion_r3219341000). At this point we could have HollowPrimaryKeyIndex use `Supplier<Boolean>` for consistency

PR #835, opened 2026-07-03
Use single-read getElementValue for readLong when field fits in <=56 bits

michaelbraun:
## Summary

`HollowObjectTypeReadStateShard.readLong` always used the two-word `getLargeElementValue`. For long fields that encode in `<=56` bits it can use the faster single-read `getElementValue` instead — the same fast/slow split `readValue()` and the delta/copy paths already use. `readLong` is a hot read primitive, and the wide path also has a data-dependent branch that mispredicts at medium widths; the single-read path is branchless.

`<=56` is safe for arbitrarily-aligned object fields (7-bit max offset + 57 = 64; 56 keeps the existing margin). `>56`-bit fields are untouched.

## Correctness

`HollowObjectTypeReadStateReadLongTest` verifies `readLong` returns identical values on both sides of the 56-bit boundary (`0/±1`, powers of two, `Long.MIN/MAX_VALUE`), in both memory modes — on-heap (`FixedLengthElementArray`) and shared-memory (`EncodedLongBuffer`), with an `instanceof` check so each is actually exercised. Existing read-engine and object-mapper tests pass.

## Benchmarks (JDK 25)

Baseline = this branch with only `readLong` reverted to always-`getLargeElementValue`. `maxBits=40` hits the fast path; `maxBits=62` is the unchanged-path control.

**x86-64** (AMD EPYC 9R14, 16 MiB L3) — end-to-end `readLong`, µs/op ±99.9% CI:

| records | maxBits | baseline | change | delta |
|---|---|---|---|---|
| 1M (in-cache) | 40 | 16,312 ± 217 | 9,305 ± 76 | **−43%** |
| 1M | 62 (control) | 15,932 ± 659 | 15,405 ± 531 | unchanged |
| 2.9M (cache-miss) | 40 | 32,541 ± 2,180 | 21,853 ± 2,075 | **−33%** |
| 2.9M | 62 (control) | 45,521 ± 1,997 | 45,574 ± 1,710 | unchanged |

Fast-path CIs don't overlap in either regime; controls are flat. Primitive `getElementValue` vs `getLargeElementValue`: 1.50–1.72× at 32/48/56 bits. (Shared 4-vCPU VM, so the 2.9M points carry ~7–10% error.)

**arm64** (Apple M2, directional): ~39% end-to-end for ≤56-bit fields; 1.9–2.2× primitive; control unchanged.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectTypeReadStateShard.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectTypeReadStateShard.java
@@ -79,11 +79,7 @@ class HollowObjectTypeReadStateShard implements HollowTypeReadStateShard {
     }
 
     public long readLong(int ordinal, int fieldIndex) {
-        long bitOffset = fieldOffset(ordinal, fieldIndex);
-        int numBitsForField = dataElements.bitsPerField[fieldIndex];
-        return numBitsForField <= 56 ?
-                dataElements.fixedLengthData.getElementValue(bitOffset, numBitsForField)
-                : dataElements.fixedLengthData.getLargeElementValue(bitOffset, numBitsForField);
+        return readValue(ordinal, fieldIndex);
     }
 
     public long readBoolean(int ordinal, int fieldIndex) {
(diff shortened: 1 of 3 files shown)


hollow/src/main/java/com/netflix/hollow/core/read/engine/object/HollowObjectTypeReadStateShard.java
```diff
@@ -81,7 +81,9 @@ public long readDouble(int ordinal, int fieldIndex) {
     public long readLong(int ordinal, int fieldIndex) {
```
Sunjeet, line 81: nit: readLong and readValue are now identical, can be deduped

Sunjeet (approved):

PR #841, opened 2026-07-15
Null check in TypeFilter to exclude types without schema

spjegan:
`TypeFilter` throws NPE when a referenced type does not have a schema. This may happen when the referenced type does not have data. This PR is an attempt to gracefully handle that case instead of failing-early.

### DIFF
--- a/hollow/src/test/java/com/netflix/hollow/core/read/engine/HollowBlobReaderMissingReferencedTypeTest.java
+++ b/hollow/src/test/java/com/netflix/hollow/core/read/engine/HollowBlobReaderMissingReferencedTypeTest.java
@@ -1,0 +1,98 @@
+/*
+ *  Copyright 2016-2019 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+package com.netflix.hollow.core.read.engine;
+
+import static com.netflix.hollow.core.read.filter.TypeFilter.newTypeFilter;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
+
+import com.netflix.hollow.api.objects.generic.GenericHollowObject;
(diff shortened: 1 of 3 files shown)


hollow/src/test/java/com/netflix/hollow/core/read/filter/TypeFilterTest.java
```diff
@@ -288,6 +290,152 @@ private static <T> List<T> l(T...items) {
...
+    }
+
+    @Test
+    public void resolveToleratesMissingCollectionElementType() {
+        // Omega -> ListOfLE (element LE); LE is absent. LERef is reachable only through LE.
+        List<HollowSchema> schemas = schemasWithout(l(Omega.class), "LE");
+
+        TypeFilter subject = newTypeFilter()
+                .excludeAll()
+                .includeRecursive("Omega")
```
azeng-netflix, line 375: Seems that even map's value `V` is missing, `subject.includes("Omega", "map")` would still be `true`. Wonder if that would cause issue when actually loading the data.

spjegan: It does not cause an issue since the value type has no records, it has no ordinals so it cannot be referenced. I have added an unit test `[readsMapWithRecursiveFilterWhenValueTypeSchemaIsAbsent](https://github.com/Netflix/hollow/pull/841/changes#diff-ef82bfcdfca28a8e3cbc97efad41b7d57fe47d5282d47b4c3f229e9257e77fc5R93)` for this.

hollow/src/test/java/com/netflix/hollow/core/read/engine/HollowBlobReaderMissingReferencedTypeTest.java
```diff
@@ -0,0 +1,237 @@
...
+     *       state ({@code getTypeState == null}); it never creates a new type state. So a long-running
+     *       consumer that applies only deltas will NOT pick up the type until it re-snapshots.</li>
+     *   <li><b>Fresh snapshot</b>: once the type has records, its schema is back in the header, the
+     *       recursive filter re-resolves to include it, and it loads correctly.</li>
+     * </ul>
+     * This is pre-existing Hollow behavior for any type absent at snapshot time (filtered out or
+     * omitted), not a consequence of the null-schema fix.
+     */
+    @Test
+    public void referencedTypeThatGainsRecordsLoadsOnFreshSnapshotButIsDiscardedOnDelta() throws IOException {
```
spjegan, line 153: @azeng-netflix this test is to cover the case where the producer started producing values for the previously-null-type. The consumer discards the delta since its type-state does not have that type. Loading it from snapshot starts loading it again.

azeng-netflix: Would be good if we can test it with a RAWHollow namespace (or our test namespace)

azeng-netflix (approved):
Thanks for the fix!
This should be a safe change to be merged in, as it does not change code paths for users who are not having the null issue.

PR #842, opened 2026-07-16
Zero-allocation snapshot iterators for list/set/map read paths

michaelbraun:
## Summary

Collection iteration (`HollowList/Set/MapTypeReadState.ordinalIterator(...)`) previously validated every element/bucket independently: each element re-read the volatile `shardsVolatile`, re-read the record's `start`/`end` bounds, and executed `Unsafe.loadFence()` **twice** (once to validate the bounds weren't torn, once to validate the value read) — **2N+1** fences to iterate a record of N elements/buckets.

This replaces the iterator returned by the concrete read states with a **validated-snapshot iterator** that captures `(shard, bounds)` once and validates **once per surfaced element** — **N+1** fences — while allocating nothing per element.

Motivation: a production CPU profile showed `Unsafe.loadFence()` (`lfence` on x86-64) as the single hottest frame in the Hollow read path, driven by per-element validation during collection iteration.

## Approach

- `snapshot()` runs the existing bounds-validation loop once (read `shardsVolatile` → shard → `start`/`end`, `loadFence`-validate, retry until stable), capturing a non-torn `(shard, start/end, size)`.
- Each `next()` reads one element/bucket from the captured shard (in-range by construction, so no throw), then does a single `loadFence`-ordered shard-identity re-validation. If a delta/reshard replaced the shard, the value is discarded, the snapshot is re-established, and the index is retried — so no torn value is ever returned.

## Correctness

Rests on the existing read-engine contract, unchanged:
- Bounds validated non-torn **before** any element arithmetic (a torn bound would otherwise produce a wild offset → AIOOBE on-heap / out-of-mapping in shared-memory).
- Reads from a captured (possibly recycled) shard are in-bounds and memory-safe because retired shards' backing arrays are recycled but **never truncated/nulled**.
- No value returned until a passing trailing validation covering its read.
- Per-element validation semantics match the original iterators (equal-or-stricter consistency; the set/map bucket-range bound is actually tightened vs. the old code, which could over-read a neighbor's buckets on a mid-iteration shrink).

Reviewed by two independent correctness passes (including an adversarial audit of the retry/OOB paths, empty-bucket sentinels, and the `numBuckets == hashTableSize(size)` invariant) — no bugs found. Residual notes (sustained-delta liveness; `invalidate()`-during-read teardown NPE) are pre-existing library-wide properties, neither introduced nor worsened here.

## Scope

- Only the concrete read states switch. The interface-level iterators (`HollowListOrdinalIterator`, `HollowSetOrdinalIterator`, `HollowMapEntryOrdinalIteratorImpl`) are retained for proxy/missing/historical data access.
- `potentialMatchOrdinalIterator` (single hash-chain lookup) is intentionally left unchanged.
- `HollowSetCopier`/`HollowMapCopier` updated to the new iterator types for `getCurrentBucket()` hash-position preservation.

## Testing

- Per-type correctness tests incl. **concurrent iterate-while-reshard** stress tests with torn-read oracles (list `v[j]==j`; set `{0..count-1}`; map `value==key+1`).
- Full suite: **1151 tests, 0 failures**.

## Benchmarks (JMH, `hollow-perf`, JDK 25)

Iteration time, 2000 records, snapshot vs. old per-element/per-bucket:

| size | list | set | map |
|---|---|---|---|
| 1   | 1.46x | 1.08x | 1.11x |
| 16  | 2.15x | 1.92x | 1.71x |
| 256 | 2.38x | 2.47x | 2.05x |

Zero added allocation at every size (the earlier size-1 regression from a materializing variant is gone). Gains grow with collection size. Measured on arm64; on x86-64 the `lfence`-bound savings should be at least as large.

The `hollow-perf/build.gradle` change adds guarded `-PjmhInclude` / `-PjmhJvm` / `-PjmhProf` hooks (inert unless the property is passed) for running/filtering benchmarks; drop if undesired.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/read/engine/list/HollowListTypeReadState.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/read/engine/list/HollowListTypeReadState.java
@@ -36,7 +36,6 @@ import com.netflix.hollow.core.read.engine.PopulatedOrdinalListener;
 import com.netflix.hollow.core.read.engine.ShardsHolder;
 import com.netflix.hollow.core.read.engine.SnapshotPopulatedOrdinalsReader;
 import com.netflix.hollow.core.read.filter.HollowFilterConfig;
-import com.netflix.hollow.core.read.iterator.HollowListOrdinalIterator;
 import com.netflix.hollow.core.read.iterator.HollowOrdinalIterator;
 import com.netflix.hollow.core.schema.HollowListSchema;
 import com.netflix.hollow.core.schema.HollowSchema;
@@ -227,10 +226,10 @@ public class HollowListTypeReadState extends HollowCollectionTypeReadState imple
     @Override
     public HollowOrdinalIterator ordinalIterator(int ordinal) {
         sampler.recordIterator();
-        return new HollowListOrdinalIterator(ordinal, this);
+        return new HollowListSnapshotOrdinalIterator(ordinal, this);
     }
 
-    private boolean readWasUnsafe(HollowListTypeShardsHolder shardsHolder, int ordinal, HollowListTypeReadStateShard shard) {
+    boolean readWasUnsafe(HollowListTypeShardsHolder shardsHolder, int ordinal, HollowListTypeReadStateShard shard) {
         HollowUnsafeHandle.getUnsafe().loadFence();
         HollowListTypeShardsHolder currShardsHolder = shardsVolatile;
         return shardsHolder != currShardsHolder
--- a/hollow/src/main/java/com/netflix/hollow/core/read/engine/set/HollowSetSnapshotOrdinalIterator.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/read/engine/set/HollowSetSnapshotOrdinalIterator.java
@@ -1,0 +1,115 @@
+/*
+ *  Copyright 2016-2019 Netflix, Inc.
+ *
+ *     Licensed under the Apache License, Version 2.0 (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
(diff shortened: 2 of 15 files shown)


hollow/src/main/java/com/netflix/hollow/core/read/engine/list/HollowListTypeReadState.java
```diff
@@ -227,10 +226,10 @@ public int size(int ordinal) {
     @Override
     public HollowOrdinalIterator ordinalIterator(int ordinal) {
         sampler.recordIterator();
-        return new HollowListOrdinalIterator(ordinal, this);
+        return new HollowListSnapshotOrdinalIterator(ordinal, this);
```
Sunjeet, line 229: seems safe, i don't see any code outside this repo casting to the old type

hollow/src/main/java/com/netflix/hollow/core/read/engine/set/HollowSetSnapshotOrdinalIterator.java
```diff
@@ -0,0 +1,115 @@
...
+ * <p>
+ * Correctness mirrors the list snapshot iterator: {@code startBucket}/{@code endBucket} are validated non-torn
+ * before use, so every relative bucket index in {@code [0, numBuckets)} maps to an in-bounds read of the shard's
+ * bucket region (garbage-but-safe if the shard's backing arrays were recycled after a concurrent delta, never
+ * out-of-bounds). Each {@link #next()} scans forward to the next non-empty bucket reading from the snapshot,
+ * then re-validates shard identity via the {@code loadFence} in
+ * {@link HollowSetTypeReadState#readWasUnsafe}; if the shard changed, the scan is discarded and re-run against a
+ * freshly established snapshot before any value is returned.
+ */
+public class HollowSetSnapshotOrdinalIterator implements HollowOrdinalIterator {
```
Sunjeet, line 38: do Map and Set need to be public, or could they be package private like List's snapshot ordinal iterator

Sunjeet:
Other than the couple minor comments, the change LGTM, but it's one that we'd want to gradually roll out opt-in behind a feature flag (ideally passed on the consumer bulider)

Sunjeet:
Incorrect assumptions in the PR description, but these don't affect the code change- code still looks good

```which could over-read a neighbor's buckets on a mid-iteration shrink```
is correctly handled by content-addressing. An ordinal can not be reused in-place, if the collection changed then it gets a new ordinal

``` torn bound would otherwise produce a wild offset → AIOOBE on-heap / out-of-mapping in shared-memory```
cannot occur since no delta transitions in shared-memory mode

DanielThomas:
I'm going to pick up this work. I'll integrate this with the PR stack I'm already working on.

PR #844, opened 2026-07-28
Added support for caching previous image

spjegan:
When object longevity is enabled, Hollow does not cache the removed records in the prior version. This PR adds support for caching it, matching what `HollowPerfAPICache` does already. It is enabled via an opt-in flag in the generated API class. The change is backward-compatible and there is no change to the existing path.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/codegen/HollowAPIClassJavaGenerator.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/codegen/HollowAPIClassJavaGenerator.java
@@ -153,6 +153,10 @@ public class HollowAPIClassJavaGenerator extends HollowConsumerJavaFileGenerator
         builder.append("    }\n\n");
 
         builder.append("    public ").append(className).append("(HollowDataAccess dataAccess, Set<String> cachedTypes, Map<String, HollowFactory<?>> factoryOverrides, ").append(className).append(" previousCycleAPI) {\n");
+        builder.append("        this(dataAccess, cachedTypes, factoryOverrides, previousCycleAPI, false);\n");
+        builder.append("    }\n\n");
+
+        builder.append("    public ").append(className).append("(HollowDataAccess dataAccess, Set<String> cachedTypes, Map<String, HollowFactory<?>> factoryOverrides, ").append(className).append(" previousCycleAPI, boolean retainRemovedOrdinalsInCache) {\n");
         builder.append("        super(dataAccess);\n");
         builder.append("        HollowTypeDataAccess typeDataAccess;\n");
         builder.append("        HollowFactory factory;\n\n");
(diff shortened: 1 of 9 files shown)


hollow/src/main/java/com/netflix/hollow/api/objects/provider/HollowObjectCacheProvider.java
```diff
@@ -59,12 +65,18 @@ public HollowObjectCacheProvider(HollowTypeDataAccess typeDataAccess, HollowType
                 while(ordinal >= arr.size())
                     arr.add(null);
 
-                if(previous != null && previousOrdinals.get(ordinal) && populatedOrdinals.get(ordinal)) {
+                boolean previouslyPopulated = previous != null && previousOrdinals.get(ordinal);
+                boolean currentlyPopulated = populatedOrdinals.get(ordinal);
+                if(previouslyPopulated && (currentlyPopulated || retainRemovedOrdinals)) {
```
azeng-netflix, line 68: NOTE: 
`previousOrdinals != null` when apply delta
If ordinal x is removed at v0, it would be preserved in v1, but in v2 it would still get removed.

hollow/src/main/java/com/netflix/hollow/api/codegen/HollowAPIClassJavaGenerator.java
```diff
@@ -153,6 +153,10 @@ public String generate() {
         builder.append("    }\n\n");
 
         builder.append("    public ").append(className).append("(HollowDataAccess dataAccess, Set<String> cachedTypes, Map<String, HollowFactory<?>> factoryOverrides, ").append(className).append(" previousCycleAPI) {\n");
+        builder.append("        this(dataAccess, cachedTypes, factoryOverrides, previousCycleAPI, false);\n");
```
azeng-netflix, line 156: NOTE: the new retainAwareConstructor.

hollow/src/main/java/com/netflix/hollow/api/objects/provider/HollowObjectCacheProvider.java
azeng-netflix: PerfAPI does [similar things](https://github.com/Netflix/hollow/blob/c8884fb6a44205abc184a99bdcf64c96d43c9b1f/hollow/src/main/java/com/netflix/hollow/api/perfapi/HollowPerfAPICache.java#L51)

azeng-netflix (approved):
Looks good and should also be safe as it's behind a feature flag.
However, for our internal use case, do user use the generator directly themselves?

hollow/src/main/java/com/netflix/hollow/api/codegen/HollowAPIClassJavaGenerator.java
spjegan: yes, this is the existing constructor that delegates (with false) to the new retainAware constructor.

hollow/src/main/java/com/netflix/hollow/api/objects/provider/HollowObjectCacheProvider.java
spjegan: yes, this is similar to PerfApi's behavior.

PR #845, opened 2026-07-30
Skip validation for a cycle

xguo-stack:
### What changes:

Introduce an optional `Supplier<Boolean>` on `HollowProducer.Builder`, threaded through `AbstractHollowProducer` the same way as `ignoreSoftLimits`. It is evaluated at the start of each cycle's validate stage; when it returns `true`, the validate stage is **skipped entirely** for that one cycle: the producer returns before firing validation-start, so no `ValidatorListener` runs (regardless of whether it was added via the builder or `HollowProducer.addListener()`), and no validation event is fired to `HollowProducerListener`s / `ValidationStatusListener`s.

Because nothing is fired, consumers of those events see **no validate stage for that cycle** — rather than an empty, passing stage (which would otherwise be indistinguishable from "zero validators configured"). Re-evaluated every cycle, so the next cycle validates normally unless the supplier again returns `true`. Defaults to never skipping, so behavior is unchanged unless a supplier is configured.

### Why changes:

Useful for one-shot operator overrides where the decision of *when* to skip validation lives outside the producer.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
@@ -106,6 +106,9 @@ abstract class AbstractHollowProducer {
     private final boolean forceCoverageOfTypeResharding;   // exercise re-sharding often (for testing)
     private final Supplier<Boolean> ignoreSoftLimits;
     private final boolean partitionedOrdinalMap;
+    // Evaluated at the start of each cycle's validation stage; when it returns true the entire
+    // validation stage is skipped for that cycle (all ValidatorListeners are bypassed).
+    private final Supplier<Boolean> skipValidation;
 
     @Deprecated
     public AbstractHollowProducer(
@@ -116,7 +119,7 @@ abstract class AbstractHollowProducer {
                 new VersionMinterWithCounter(), null, 0,
                 DEFAULT_TARGET_MAX_TYPE_SHARD_SIZE, false, false, false, false, null,
                 new DummyBlobStorageCleaner(), new BasicSingleProducerEnforcer(),
-                null, true, HollowConsumer.UpdatePlanBlobVerifier.DEFAULT_INSTANCE, null);
+                null, true, HollowConsumer.UpdatePlanBlobVerifier.DEFAULT_INSTANCE, null, null);
     }
 
     // The only constructor should be that which accepts a builder
@@ -129,7 +132,8 @@ abstract class AbstractHollowProducer {
                 b.numStatesBetweenSnapshots, b.targetMaxTypeShardSize, b.focusHoleFillInFewestShards,
                 b.allowTypeResharding, b.forceCoverageOfTypeResharding, b.partitionedOrdinalMap,
                 b.metricsCollector, b.blobStorageCleaner, b.singleProducerEnforcer,
-                b.hashCodeFinder, b.doIntegrityCheck, b.updatePlanBlobVerifier, b.ignoreSoftLimits);
+                b.hashCodeFinder, b.doIntegrityCheck, b.updatePlanBlobVerifier, b.ignoreSoftLimits,
+                b.skipValidation);
     }
 
     private final HollowProducerListener producerMetricsListener;
@@ -156,7 +160,8 @@ abstract class AbstractHollowProducer {
             HollowObjectHashCodeFinder hashCodeFinder,
             boolean doIntegrityCheck,
             HollowConsumer.UpdatePlanBlobVerifier updatePlanBlobVerifier,
-            Supplier<Boolean> ignoreSoftLimits) {
+            Supplier<Boolean> ignoreSoftLimits,
+            Supplier<Boolean> skipValidation) {
         this.publisher = publisher;
         this.announcer = announcer;
         this.versionMinter = versionMinter;
@@ -172,6 +177,7 @@ abstract class AbstractHollowProducer {
         this.focusHoleFillInFewestShards = focusHoleFillInFewestShards;
         this.ignoreSoftLimits = ignoreSoftLimits;
         this.partitionedOrdinalMap = partitionedOrdinalMap;
+        this.skipValidation = skipValidation;
 
         HollowWriteStateEngine writeEngine = hashCodeFinder == null
                 ? new HollowWriteStateEngine()
@@ -968,6 +974,14 @@ abstract class AbstractHollowProducer {
 
         ValidationStatus status = null;
         try {
+            if (skipValidation != null && Boolean.TRUE.equals(skipValidation.get())) {
+                log.info("Skipping validation stage for this cycle; skipValidation supplier returned true. "
+                        + "All validators are bypassed for this cycle only.");
+                status = new ValidationStatus(Collections.emptyList());
+                psb.success();
+                return;
+            }
+
             // Stream over the concatenation of the old and new validators
             List<ValidationResult> results =
                     listeners.getListeners(ValidatorListener.class)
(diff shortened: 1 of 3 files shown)


hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -968,6 +974,14 @@ private void validate(ProducerListeners listeners, HollowProducer.ReadState read
 
         ValidationStatus status = null;
         try {
+            if (skipValidation != null && Boolean.TRUE.equals(skipValidation.get())) {
+                log.info("Skipping validation stage for this cycle; skipValidation supplier returned true. "
+                        + "All validators are bypassed for this cycle only.");
```
workeatsleep: remove the "only" in  "All validators are bypassed for this cycle only", 
based on the change, we cannot tell it is bypassed for this cycle only, it is bypassed or not purely based on what the supplier returns.

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -116,7 +119,7 @@ public AbstractHollowProducer(
                 new VersionMinterWithCounter(), null, 0,
                 DEFAULT_TARGET_MAX_TYPE_SHARD_SIZE, false, false, false, false, null,
                 new DummyBlobStorageCleaner(), new BasicSingleProducerEnforcer(),
-                null, true, HollowConsumer.UpdatePlanBlobVerifier.DEFAULT_INSTANCE, null);
+                null, true, HollowConsumer.UpdatePlanBlobVerifier.DEFAULT_INSTANCE, null, null);
```
workeatsleep: instead of "null" use "() -> false" as the last contructor param

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
workeatsleep, line 984: let's see what the beacon and cycle log looks like with this option VS completely ignore the validate step to see which one makes more sense to user

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -999,6 +1001,26 @@ public B withIgnoreSoftLimits(Supplier<Boolean> ignoreSoftLimits) {
             return (B) this;
         }
 
+        /**
+         * Register a supplier that decides, per cycle, whether to skip the producer's validation stage.
+         * <p>
+         * The supplier is evaluated once at the start of each cycle's validation stage. When it returns
```
workeatsleep: make the comment more concise

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
```diff
@@ -755,6 +755,8 @@ public static class Builder<B extends HollowProducer.Builder<B>> {
         ProducerOptionalBlobPartConfig optionalPartConfig = null;
         HollowConsumer.UpdatePlanBlobVerifier updatePlanBlobVerifier = HollowConsumer.UpdatePlanBlobVerifier.DEFAULT_INSTANCE;
         Supplier<Boolean> ignoreSoftLimits = null;
+        // Defaults to a supplier that never skips; validation always runs unless overridden.
```
AustinWheel: nit: is this comment needed? skipValidation feels readable without it

hollow/src/main/java/com/netflix/hollow/api/producer/HollowProducer.java
AustinWheel: can we also change the wording, I feel "skip" and "bypass" are less precise than saying something like "no `ValidatorListener` runs and all are passed"

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -106,6 +106,9 @@ abstract class AbstractHollowProducer {
     private final boolean forceCoverageOfTypeResharding;   // exercise re-sharding often (for testing)
     private final Supplier<Boolean> ignoreSoftLimits;
     private final boolean partitionedOrdinalMap;
+    // Evaluated at the start of each cycle's validation stage; when it returns true the entire
+    // validation stage is skipped for that cycle (all ValidatorListeners are bypassed).
```
AustinWheel: nit: same thing about the wording of skipped/bypassed

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
xguo-stack: ### Option 1 -- make validation a no-op

Image

Image

### Option 2 -- skip the validation step completely. 

Image

xguo-stack (commented):
Added verification result.

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
xguo-stack: We will go with option 2

Sunjeet (approved):

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -106,6 +106,12 @@ abstract class AbstractHollowProducer {
     private final boolean forceCoverageOfTypeResharding;   // exercise re-sharding often (for testing)
     private final Supplier<Boolean> ignoreSoftLimits;
     private final boolean partitionedOrdinalMap;
+    // Evaluated at the start of each cycle, before the validate stage would otherwise begin; when it
```
Sunjeet: nit: evaluated at the start of the validate stage

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -964,6 +973,17 @@ private void applyDelta(HollowProducer.Blob blob, HollowReadStateEngine stateEng
     }
 
     private void validate(ProducerListeners listeners, HollowProducer.ReadState readState) {
+        if (skipValidation != null && Boolean.TRUE.equals(skipValidation.get())) {
+            // Return before firing validation-start at all: no ValidatorListener runs, and no Validate
+            // stage start/complete event reaches CycleListeners/ValidationStatusListeners for this
+            // cycle. Beacon/cycle log tooling will show no Validate stage entry for this cycle, rather
+            // than a synthetic "passed with zero results" entry that reads the same as "zero
+            // validators configured".
+            log.info("skipValidation supplier returned true; the validate stage does not run for this "
```
Sunjeet: nit: also log the version for which validate stage was skipped

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -106,6 +106,12 @@ abstract class AbstractHollowProducer {
     private final boolean forceCoverageOfTypeResharding;   // exercise re-sharding often (for testing)
     private final Supplier<Boolean> ignoreSoftLimits;
     private final boolean partitionedOrdinalMap;
+    // Evaluated at the start of each cycle, before the validate stage would otherwise begin; when it
+    // returns true the validate stage does not run at all for that cycle -- no ValidatorListener runs,
+    // and no Validate stage start/complete event is fired to CycleListeners/ValidationStatusListeners
```
Sunjeet: nit: the comment feels more detailed than necessary, but also would be good to check if this ai-flagged feedback is true
```
 The javadoc names the wrong listener interface                                                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                           
  ProducerListenerSupport.java:259-282 shows fireValidationStart/fireValidationComplete firing to HollowProducerListener and ValidationStatusListener. CycleListener only declares onCycleSkip/onNewDeltaChain/onCycleStart/onCycleComplete — it has never received validation callbacks.                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                                           
  The javadoc and field comment both name CycleListener (vacuous — nothing changes for it) and omit HollowProducerListener (:283, :294), the legacy interface that does get these events and that most existing integrations implement. Someone auditing "will my monitoring break?" is misled in both directions. This overlaps with the wording comments already on the PR, but it's a   
  factual error, not phrasing.    
```

PR #847, opened 2026-08-10
Skip sorting in compact when there are no removals/replacements

spjegan:
Each producer cycle unconditionally sorts the ordinal map for each type as part of `prepareForNextCycle`. This sorting can be skipped when there are no removed/replaced ordinals in a cycle. Flame graph from an internal cycle

Screenshot 2026-08-10 at 12 47 00 PM

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
@@ -21,6 +21,7 @@ import com.netflix.hollow.core.memory.encoding.VarInt;
 import com.netflix.hollow.core.memory.pool.WastefulRecycler;
 import java.util.Arrays;
 import java.util.BitSet;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicLongArray;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -54,6 +55,11 @@ public class ByteArrayOrdinalMap {
     /// ordinal's value space and forcing a broken delta chain.
     /// NOTE: not using final to allow value modification through reflection for test purpose.
     private static int SOFT_ORDINAL_LIMIT = 1 << (BITS_PER_ORDINAL - 1);
+
+    // Opt-in (default off): skip the compact sort+rehash when no ordinals are freed.
+    static boolean COMPACT_SKIP_WHEN_NO_FREES =
+            Boolean.getBoolean("com.netflix.hollow.core.memory.ByteArrayOrdinalMap.enableCompactSkip");
+    static final AtomicLong COMPACT_SKIP_COUNT = new AtomicLong();
 
     /// Thread safety:  We need volatile access semantics to the individual elements in the
     /// pointersAndOrdinals array.
@@ -463,13 +469,33 @@ public class ByteArrayOrdinalMap {
         long[] populatedReverseKeys = new long[size];
 
         int counter = 0;
+        boolean anyFreed = false;
         AtomicLongArray pao = pointersAndOrdinals;
 
         for (int i = 0; i < pao.length(); i++) {
             long key = pao.get(i);
             if (key != EMPTY_BUCKET_VALUE) {
                 populatedReverseKeys[counter++] = key << BITS_PER_ORDINAL | key >>> BITS_PER_POINTER;
+                if (COMPACT_SKIP_WHEN_NO_FREES && !anyFreed) {
+                    int ordinal = (int) (key >>> BITS_PER_POINTER);
+                    if (!usedGlobalOrdinals.get((ordinal << mapIndexBits) | mapIdx)) {
+                        anyFreed = true;
+                    }
+                }
             }
+        }
+
+        // Nothing freed => no holes to reclaim and the hash mapping is unchanged.
+        if (COMPACT_SKIP_WHEN_NO_FREES && !anyFreed) {
+            if (focusHoleFillInFewestShards && numShards > 1) {
+                freeOrdinalTracker.sort(numShards, mapIndexBits, mapIdx);
+            } else {
+                freeOrdinalTracker.sort();
+            }
+            pointersByOrdinal = null;
+            unusedPreviousOrdinals = null;
+            COMPACT_SKIP_COUNT.incrementAndGet();
+            return;
         }
 
         Arrays.sort(populatedReverseKeys);
--- a/hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
+++ b/hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
@@ -56,10 +56,11 @@ public class ByteArrayOrdinalMap {
     /// NOTE: not using final to allow value modification through reflection for test purpose.
     private static int SOFT_ORDINAL_LIMIT = 1 << (BITS_PER_ORDINAL - 1);
 
-    // Opt-in (default off): skip the compact sort+rehash when no ordinals are freed.
-    static boolean COMPACT_SKIP_WHEN_NO_FREES =
-            Boolean.getBoolean("com.netflix.hollow.core.memory.ByteArrayOrdinalMap.enableCompactSkip");
-    static final AtomicLong COMPACT_SKIP_COUNT = new AtomicLong();
+    // Opt-in (default off): when a cycle frees no ordinals there are no holes to reclaim and the hash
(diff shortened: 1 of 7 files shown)


hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
```diff
@@ -55,6 +56,11 @@ public class ByteArrayOrdinalMap {
     /// NOTE: not using final to allow value modification through reflection for test purpose.
     private static int SOFT_ORDINAL_LIMIT = 1 << (BITS_PER_ORDINAL - 1);
 
+    // Opt-in (default off): skip the compact sort+rehash when no ordinals are freed.
+    static boolean COMPACT_SKIP_WHEN_NO_FREES =
+            Boolean.getBoolean("com.netflix.hollow.core.memory.ByteArrayOrdinalMap.enableCompactSkip");
+    static final AtomicLong COMPACT_SKIP_COUNT = new AtomicLong();
```
azeng-netflix, line 62: nit: Expose this field and add logging or metrics at cinder level for observability?

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
```diff
@@ -55,6 +56,11 @@ public class ByteArrayOrdinalMap {
     /// NOTE: not using final to allow value modification through reflection for test purpose.
     private static int SOFT_ORDINAL_LIMIT = 1 << (BITS_PER_ORDINAL - 1);
 
+    // Opt-in (default off): skip the compact sort+rehash when no ordinals are freed.
+    static boolean COMPACT_SKIP_WHEN_NO_FREES =
+            Boolean.getBoolean("com.netflix.hollow.core.memory.ByteArrayOrdinalMap.enableCompactSkip");
```
azeng-netflix, line 61: nit: `enableCompactSkip` sounds it would skip compaction, maybe `opportunisticCompact`?

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
```diff
@@ -463,13 +469,33 @@ public void compact(ThreadSafeBitSet usedGlobalOrdinals, int numShards, boolean
...
+                    if (!usedGlobalOrdinals.get((ordinal << mapIndexBits) | mapIdx)) {
+                        anyFreed = true;
+                    }
+                }
+            }
+        }
+
+        // Nothing freed => no holes to reclaim and the hash mapping is unchanged.
+        if (COMPACT_SKIP_WHEN_NO_FREES && !anyFreed) {
+            if (focusHoleFillInFewestShards && numShards > 1) {
```
azeng-netflix, line 490: nit: possible to avoid the duplicate code?

azeng-netflix (approved):
Looks good.

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
spjegan: Added a FINE log statement. On exposing it via Cinder... wondering if we should enable this by default.

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
spjegan: Done

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
spjegan: Done.

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
```diff
@@ -55,6 +56,12 @@ public class ByteArrayOrdinalMap {
     /// NOTE: not using final to allow value modification through reflection for test purpose.
     private static int SOFT_ORDINAL_LIMIT = 1 << (BITS_PER_ORDINAL - 1);
 
+    // Opt-in (default off): when a cycle frees no ordinals there are no holes to reclaim and the hash
+    // mapping is unchanged, so the sort+rehash portion of compact() can be skipped.
+    static boolean OPPORTUNISTIC_COMPACT =
+            Boolean.getBoolean("com.netflix.hollow.core.memory.ByteArrayOrdinalMap.opportunisticCompact");
```
Sunjeet: nit: avoid use of system properties

Sunjeet: nit: name of property can be shortened, hollow.<something>

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
```diff
@@ -463,13 +470,32 @@ public void compact(ThreadSafeBitSet usedGlobalOrdinals, int numShards, boolean
         long[] populatedReverseKeys = new long[size];
```
Sunjeet: the fast path still allocates and fills long[size], could likely also be optimized away

Sunjeet: also, lets assert in the tests that the fast path was actually taken

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
```diff
@@ -521,6 +544,21 @@ public void compact(ThreadSafeBitSet usedGlobalOrdinals, int numShards, boolean
...
+            freeOrdinalTracker.sort(numShards, mapIndexBits, mapIdx);
+        else
+            freeOrdinalTracker.sort();
+    }
+
+    /**
+     * Number of {@link #compact} calls that took the opportunistic fast path — cycles in which no ordinals
+     * were freed, so the sort+rehash was skipped. Exposed for observability.
+     */
+    public static long getOpportunisticCompactCount() {
```
Sunjeet: note- this will be global to jvm - do we need this public?

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
```diff
@@ -463,13 +470,32 @@ public void compact(ThreadSafeBitSet usedGlobalOrdinals, int numShards, boolean
...
 
         int counter = 0;
+        boolean anyFreed = false;
         AtomicLongArray pao = pointersAndOrdinals;
 
         for (int i = 0; i < pao.length(); i++) {
             long key = pao.get(i);
             if (key != EMPTY_BUCKET_VALUE) {
                 populatedReverseKeys[counter++] = key << BITS_PER_ORDINAL | key >>> BITS_PER_POINTER;
+                if (OPPORTUNISTIC_COMPACT && !anyFreed) {
```
Sunjeet: non-final flag is read twice, and if it flips in between then we may get some odd behavior - would be good to freeze it at the top

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
spjegan: Removed the system property and added a method `withOpportunisticCompact` in Producer.Builder class.

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
spjegan: >the fast path still allocates and fills long[size], could likely also be optimized away
Done.

>also, lets assert in the tests that the fast path was actually taken
Added an assert that tests that the fast path was taken.

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
spjegan: This is not needed with the recent changes. Removed it.

hollow/src/main/java/com/netflix/hollow/core/memory/ByteArrayOrdinalMap.java
spjegan: This is a method parameter now, so it is frozen.

PR #848, opened 2026-08-13
Parallel integ check in producer cycle

spjegan:
Integrity checks in the producer side are currently parallelized across types. This means, if a type has many shards, the integrity check happens sequentially for each shard. This PR changes that by parallelizing it "per type and per shard". This feature is an opt-in via the builder. Opting out retains the old functionality. The benefit we get by parallelizing by shard is applicable only when type-sharding is enabled.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
+++ b/hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
@@ -94,6 +94,7 @@ abstract class AbstractHollowProducer {
     long lastSuccessfulCycle = 0;
     final HollowObjectHashCodeFinder hashCodeFinder;
     final boolean doIntegrityCheck;
+    final boolean parallelPerShardChecksum;
     // Count to track number of cycles run by a primary producer. In the future, this can be useful in determining stickiness of a
     // producer instance.
     int cycleCountSincePrimaryStatus = 0;
@@ -121,7 +122,7 @@ abstract class AbstractHollowProducer {
                 new VersionMinterWithCounter(), null, 0,
                 DEFAULT_TARGET_MAX_TYPE_SHARD_SIZE, false, false, false, false, null,
                 new DummyBlobStorageCleaner(), new BasicSingleProducerEnforcer(),
-                null, true, HollowConsumer.UpdatePlanBlobVerifier.DEFAULT_INSTANCE, null, () -> false);
+                null, true, HollowConsumer.UpdatePlanBlobVerifier.DEFAULT_INSTANCE, null, () -> false, false);
     }
 
     // The only constructor should be that which accepts a builder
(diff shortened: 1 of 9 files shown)


hollow/src/main/java/com/netflix/hollow/tools/checksum/HollowChecksum.java
```diff
@@ -120,6 +150,67 @@ public void run() {
         return totalChecksum;
     }
 
+    private static HollowChecksum forStateEngineParallelPerShard(HollowReadStateEngine stateEngine, HollowReadStateEngine commonSchemasWithState) {
+        SimultaneousExecutor executor = new SimultaneousExecutor(HollowChecksum.class, "checksum-common-schemas");
```
Sunjeet: nit: executor can be an instance variable

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
```diff
@@ -888,10 +888,14 @@ private ReadStateHelper checkIntegrity(
                 HollowReadStateEngine current = readStates.current().getStateEngine();
 
                 log.info("CHECKSUMS");
-                HollowChecksum currentChecksum = HollowChecksum.forStateEngineWithCommonSchemas(current, pending);
+                // Read the per-shard checksum mode once for the whole integrity check: the per-shard and per-type
+                // computations yield different values by design, so all four checksums below must use the same mode
+                // even if the system property were toggled mid-cycle (which would otherwise cause a spurious failure).
+                boolean parallelPerShardChecksum = Boolean.getBoolean(HollowChecksum.PARALLEL_PER_SHARD_PROPERTY);
```
Sunjeet: avoid using system property, favor setting using the producer builder

hollow/src/main/java/com/netflix/hollow/core/write/HollowBlobWriter.java
```diff
@@ -80,12 +88,28 @@ public void writeSnapshot(OutputStream os, ProducerOptionalBlobPartConfig.Option
...
+            if(parallelPerShard) {
+                // Fan out one task per (type, shard). prepareCalculateSnapshot allocates the shared per-shard
+                // structures up-front (on this thread) so the shard tasks only fill their own slot.
+                typeState.prepareCalculateSnapshot();
+                int numSnapshotShards = typeState.numSnapshotShards();
+                for(int shard = 0; shard < numSnapshotShards; shard++) {
+                    final int shardNumber = shard;
+                    executor.execute(new Runnable() {
+                        public void run() {
+                            typeState.calculateSnapshotShard(shardNumber);
```
Sunjeet: note this will require more memory- the entire snapshot's buffers will be resident in memory

hollow/src/main/java/com/netflix/hollow/api/producer/AbstractHollowProducer.java
spjegan: Moved this to a builder method.

hollow/src/main/java/com/netflix/hollow/core/write/HollowBlobWriter.java
spjegan: Yeah, I have removed the changes related to parallelizing the snapshot serialization from this PR. Will revisit that later in a separate PR.

PR #850, opened 2026-08-24
compaction- gradually with maxDeltaBytesPerCycle

Sunjeet:
By default a compaction cycle relocates every misplaced record at once, which can produce a very large delta.  
Supplying an `approxDeltaBytesPerCycle` to the `CompactionConfig` bounds this -- each cycle then compacts only the 
single type with the largest hole footprint, and only as many of its records as fit within the budget, so compaction converges 
over a series of smaller cycles. The budget also counts the records which reference a relocated record, since those 
must be rewritten too. If it is too small to cover even one record, nothing is compacted and a warning is logged.  
See the `HollowCompactor` javadoc for more details.

### DIFF
--- a/hollow/src/main/java/com/netflix/hollow/tools/compact/HollowCompactor.java
+++ b/hollow/src/main/java/com/netflix/hollow/tools/compact/HollowCompactor.java
@@ -120,7 +120,7 @@ public class HollowCompactor {
      * @return {@code true} if compaction is necessary, otherwise {@code false}
      */
     public boolean needsCompaction() {
-        return !compactionPlan().isEmpty();
+        return !calculateCompactionPlan().isEmpty();
     }
     
     /**
@@ -137,7 +137,7 @@ public class HollowCompactor {
      *   
      */
     public void compact() {
-        Map<String, Integer> plan = compactionPlan();
+        Map<String, Integer> plan = calculateCompactionPlan();
 
         Map<String, BitSet> relocatedOrdinals = new HashMap<String, BitSet>();
         PartialOrdinalRemapper remapper = new PartialOrdinalRemapper();
@@ -217,7 +217,7 @@ public class HollowCompactor {
      * The types to compact in this cycle, mapped to the number of records to relocate from each.  A type is omitted
      * when the smallest batch it could produce (record + closure) > approxDeltaBytesPerCycle.
      */
-    private Map<String, Integer> compactionPlan() {
+    private Map<String, Integer> calculateCompactionPlan() {
         if(compactionPlan != null)
             return compactionPlan;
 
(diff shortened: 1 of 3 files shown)


hollow/src/main/java/com/netflix/hollow/tools/compact/HollowCompactor.java
```diff
@@ -185,9 +213,59 @@ public void compact() {
         }
     }
     
+    /**
+     * The types to compact in this cycle, mapped to the number of records to relocate from each.  A type is omitted
+     * when the smallest batch it could produce (record + closure) > approxDeltaBytesPerCycle.
+     */
+    private Map<String, Integer> compactionPlan() {
```
workeatsleep: nit: rename to calculateCompactionPlan so it does not feel like a static object getting created

workeatsleep (approved):

azeng-netflix:
For our target dataset, what's the cost of running it?

hollow/src/main/java/com/netflix/hollow/tools/compact/HollowCompactor.java
```diff
@@ -202,7 +280,78 @@ private Set<String> findCompactionTargets() {
...
+            return 0;
+        }
+
+        return numRelocations;
+    }
+
+    private long approximateDeltaBytes(String compactionTarget, BitSet candidateRelocations) {
+        Map<String, BitSet> churnedOrdinals = new HashMap<String, BitSet>();
+        churnedOrdinals.put(compactionTarget, candidateRelocations);
+        TransitiveSetTraverser.addReferencingOutsideClosure(readEngine, churnedOrdinals);
```
azeng-netflix, line 329: Would line 183 duplicate the computation done here? If so, is it worth to cache the results of addReferencingOutsideClosure and reuse it later?

hollow/src/main/java/com/netflix/hollow/tools/compact/HollowCompactor.java
```diff
@@ -202,7 +280,78 @@ private Set<String> findCompactionTargets() {
...
+        return costliestType == null ? compactionTargets : Collections.singleton(costliestType);
+    }
+
+    /**
+     * Determine how many of the highest ordinals in the given type may be relocated within the delta size budget
+     */
+    private int relocationsWithinDeltaBudget(String compactionTarget, BitSet populatedOrdinals, int numRelocations) {
+        long deltaBytes = approximateDeltaBytes(compactionTarget, highestPopulatedOrdinals(populatedOrdinals, numRelocations));
+
+        while(deltaBytes > approxDeltaBytesPerCycle && numRelocations > 1) {
```
azeng-netflix: should we just run `numRelocations = (int)Math.max(1, Math.min(numRelocations - 1, numRelocations * approxDeltaBytesPerCycle / deltaBytes));` once to get a rough estimate?
I am thinking cases, where it can run multiple times if the record size within a type has large variance (i.e. some records are heavily referenced by other types, where others are hardly referenced).

hollow/src/main/java/com/netflix/hollow/tools/compact/HollowCompactor.java
Sunjeet: incorporated

hollow/src/main/java/com/netflix/hollow/tools/compact/HollowCompactor.java
Sunjeet: Good point, but we'd have to cache BitSets, feels ok to for e.g. have ~3% slower compaction cycles instead of requiring ~3% more ram

hollow/src/main/java/com/netflix/hollow/tools/compact/HollowCompactor.java
Sunjeet: Yeah this feels better- incorporated, thanks!

PR #851, opened 2026-09-03
fix: Latent index size limits

Sunjeet:
Indexes now support upto 1.5B records.  

fix: HollowHashIndex crashes at ~188M records with NegativeArraySizeException.
fix: HollowPrimaryKeyIndex threw IllegalArgumentException at ~751M records.

The fix computes the threshold in long and widens bucket counts one bit, to 2^31 buckets = ~1.5B records for both index types.


spjegan (approved):
LGTM.

I am curious if we have timed index building with 1.5B records and how acceptable it is. It looks like `growIntermediateHashTable` seems to rehash everything every time it grows. This is an existing behavior so can be taken care later, but with 1.5B records it may become a performance bottleneck.

PR #853, opened 2026-09-21
Skip read validation for immutable shards

DanielThomas:
This avoids the trailing read validation when the underlying arrays are known to never be reused. This is a first in a series of pull requests in a stack that use the fact that arrays are not recycled as an invariant for further optimisation.

This builds on the work in #647. It added `GarbageCollectorAwareRecycler` and selects `WastefulRecycler` for concurrent garbage collectors, but does not remove the synchronization cost required by `RecyclingRecycler`.

## Optimization Opportunities

In a 30-second Images CPU profile on JDK 25 with ZGC, Hollow appeared in **26.8% of samples**. More than half of that time was concentrated in two low-level read operations:

- Trailing read-validation fences: **7.4% of total CPU**
- Fixed-length value extraction: **7.0% of total CPU**

The remaining cost was spread across object fields, strings, collections, and primary-key lookups. This PR starts a stack that removes unnecessary validation for immutable shards, then optimizes the underlying fixed-length, string, collection, and byte read paths.


Sunjeet:
CPU the win on ARM should be larger than x86.

A minor plumbing thing that'd be good to fix- 
```
HollowTypeReadState: derives the flag from stateEngine.getMemoryRecycler(). But the recycler that actually allocates and later recycles those segments is the one passed to the public readSnapshot(in,      
  recycler, numShards) / applyDelta(in, schema, recycler, numShards) and stored as HollowTypeDataElements.memoryRecycler (HollowTypeDataElements.java:14). Two independent sources.                            
                                                                                                                                                                                                               
  In-tree they always agree — HollowBlobReader.java:365,378 passes stateEngine.getMemoryRecycler(), and every other caller routes through HollowBlobReader. But nothing enforces it, and both entry points are 
  public API taking an arbitrary recycler. Hand a RecyclingRecycler to an engine constructed with a non-recycling one and you get validation skipped while arrays are reused — precisely the failure mode the  
  package doc names: "both validations pass while the value returned is recycled memory, with no exception and no detection."                                                                                  
                                                                                                                                                                                                               
  Secondary issue in the same spot: the flag is a constructor-time snapshot of a property owned by an object that doesn't exist yet and is rebuilt on every delta. Even with no mismatch, the derivation is    
  chronologically backwards.                                                                                                                                                                                   
                                                                                                                                                                                                               
  Fix, cheapest first:                                                                                                                                                                                         
  // in readSnapshot/applyDelta, after data elements are built                                                                                                                                                 
  if (shardsAreImmutable && dataElements.memoryRecycler.recyclesArrays())                                                                                                                                      
      throw new IllegalStateException("...");                                                                                                                                                                  
  Better: derive the flag from dataElements.memoryRecycler at publish time rather than from the engine at construction time.                                                                                   
```

PR #856, opened 2026-09-22
Decode and determine equality for strings directly from immutable byte segments

DanielThomas:
This is a String decoding change I've wanted to make for years, and a string equality change that also shows up on read paths.

When decoding `String` from `SegmentedByteArray`, every call resolves `segments[index]` again, and with ZGC, that also means another reference load barrier. In an Images profile, those byte lookups while decoding strings accounted for about 1.6% of total process CPU.

The immutability established by #853 lets this path load the segment table once and then process each contiguous `byte[]` range directly. That same segmented traversal is also now used by `isStringFieldEqual(...)`, which is primarily exercised by primary-key matching.

The JDK has since seen a lot of investment in `String` to support compact strings, so a contiguous ASCII range now goes directly to the specialized `String(byte[], ..., ISO_8859_1)` path. On JDK 19 and later, contiguous ranges also use the JDK ASCII decoder to identify and inflate the leading ASCII run through the `countPositives` intrinsic. Earlier releases retain the scalar path; JDK 17 exposed the decoder fast path but did not yet have the prefix-returning intrinsic and regressed on mixed strings.

## Benchmark

`HollowObjectTypeReadStateShardBenchmark.testReadString`, Azul JDK 25.0.4.1, `-XX:+UseZGC`, 500 strings per operation, and a 10% probability of `U+123E` characters:

| Maximum string length | Parent (#853) | This PR | Improvement |
|---:|---:|---:|---:|
| 25 | 0.0274 ms/op | 0.0220 ms/op | 19.7% |
| 150 | 0.0952 ms/op | 0.0610 ms/op | 35.9% |

The additional JDK 19+ ASCII-prefix path, measured on Azul JDK 21.0.12.1 with 500 strings per operation:

| Maximum length | Unicode probability | Scalar prefix | JDK intrinsic | Improvement |
|---:|---:|---:|---:|---:|
| 150 | 1% | 0.045 ms/op | 0.040 ms/op | 11% |
| 150 | 10% | 0.064 ms/op | 0.047 ms/op | 27% |
| 1000 | 1% | 0.323 ms/op | 0.172 ms/op | 47% |
| 1000 | 10% | 0.401 ms/op | 0.359 ms/op | 10% |

`HollowObjectTypeReadStateShardBenchmark.testStringEquality`, Azul JDK 25.0.4.1, 500 matching comparisons per operation, non-recycling storage, and the same Unicode distribution:

| Maximum string length | Per-byte equality | Segmented equality | Improvement |
|---:|---:|---:|---:|
| 25 | 0.033 ms/op | 0.027 ms/op | 18.2% |
| 150 | 0.118 ms/op | 0.081 ms/op | 31.4% |


Sunjeet:
Redundant Arrays.fill on the default (non-immutable) path. (minor regression)                                                                                                                             
                                                                                                                                                                                                               
  The diff removed the else:                                                                                                                                                                                   
                                                                                                                                                                                                               
  if (length > chararr.length)                                                                                                                                                                                 
      chararr = new char[length];                                                                                                                                                                              
                                                                                                                                                                                                               
  if(immutable)                                                                                                                                                                                                
      return ((SegmentedByteArray)data).readVIntString(position, length, chararr);                                                                                                                             
                                                                                                                                                                                                               
  Arrays.fill(chararr, 0, length, '\0');                                                                                                                                                                       
                                                                                                                                                                                                               
  When length > 100, chararr was just allocated and is already zeroed, so the fill is a pure O(length) waste — on the path every default-configured consumer still uses. Restoring the else for the            
  non-immutable branch keeps it free.

hollow/src/main/java/com/netflix/hollow/core/memory/SegmentedByteArray.java
```diff
@@ -77,6 +88,166 @@ public byte get(long index) {
...
+                    output[count++] = (char)value;
+                    value = 0;
+                }
+            }
+        }
+
+        return new String(output, 0, count);
+    }
+
+    private static String readContiguousVIntString(byte[] data, int offset, int length, char[] output) {
```
Sunjeet: Does not seem tested outside of jmh, and our CI runs java 8

PR #858, opened 2026-09-22
Use contiguous fixed-length storage for immutable data

DanielThomas:
`FixedLengthElementArray.getElementValue` is 6.83% of CPU in the Images profile. Most samples attributed inside the method land around `segments[whichSegment]`. 

Every fixed-length read resolves a nested `long[]` before loading the primitive value. That indirection is useful when fixed-size arrays are recycled, but hits a load barrier in concurrent GCs, and the preceding PRs give us a read path where retired arrays cannot be reused.

On JDK 25 with ZGC on an Apple M1 Max, one million random 40-bit reads produced:

| Backing data | 1M records (~5 MB) | 10M records (~50 MB) |
|---|---:|---:|
| Segmented `long[][]` | 1,715.6 us | 6,952.3 us |
| Contiguous `long[]` | 926.6 us | 4,429.3 us |
| Change | **-46.0%** | **-36.3%** |


Sunjeet:
Test coverage is becoming GC dependent, we'll want to parameterize the tests

Sunjeet:
Indexes call new FixedLengthElementArray(...) directly and bypass the factory.. That could be a followup

